PR #14, opened 2017-03-09
Fix formatting in "Variables" section of docs

jdanford:


djc:
Thank you for the fixes! Can't merge from phone due to the bad coverage feedback, will merge them soon though. 

Any other thoughts on things that could be improved in the docs? Any questions that went unanswered?

djc:
Merged, thanks!

jdanford:
I only skimmed them for now – I'll let you know when I take a deeper look!

PR #16, opened 2017-03-10
Add `lower` and `upper` filters

LegNeato:


djc:
Should have thought of these myself, thanks again! 

How are you liking it so far? Any documentation gaps, surprises, pitfalls?

LegNeato:
Thanks for the super fast response! 🍻 

Liking it so far, thanks so much for the lib and the clear docs.  No documentation gaps, surprises, or pitfalls  💯

PR #17, opened 2017-03-10
Fix compilation when deriving from Template and std and askama aren't in scope

Eijebong:
I'm not sure if it's the right way to fix it but it fixes it.

Here are the errors i was getting
```
5 | #[derive(Template)]
  |          ^^^^^^^^ Use of undeclared type or module `std`

5 | #[derive(Template)]
  |          ^^^^^^^^ Use of undeclared type or module `askama`
```

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -447,6 +447,9 @@ pub fn generate(ast: &syn::DeriveInput, path: &str, mut nodes: Vec<Node>) -> Str
     }
 
     let mut gen = Generator::new();
+    gen.writeln("use askama;");
+    gen.writeln("use std;");
+
     if !blocks.is_empty() {
         if base.is_none() {
             gen.define_trait(path, &block_names);


Eijebong:
Ok now it's complaining about traits not being found, I might have missed something and this PR might be wrong/useless, i'll investigate

djc:
Hey, thanks for trying out Askama! Yeah, I'm not sure this would be an improvement. Can you give me a reduced example of how you ran into this, so we can look at what the best solution might be? Right now, I'm thinking improved documentation/examples might be the best way forward for this.

Eijebong:
Here is a minimal example.

https://github.com/Eijebong/bug_askama

djc:
Thanks, that helped! With the change I just made, the `use` statements should no longer be necessary, making it a bit friendlier to use.

Eijebong:
I get that with blocks now:

error[E0405]: cannot find trait `TraitFrombase2ehtml2etera` in this scope

I've updated my repo with an example :)
Do you want me to open a proper issue ?

djc:
I briefly checked on my phone, but are you sure you get the same error message in the bug repo? The bug repo looks like it should be fine.

The trick is that the inheriting struct must have the base struct in scope, and it will find by a generated name, so you should have something like `use base_module::*` in the inheriting struct's module.

If that doesn't help, I'll take a proper look tomorrow.

djc:
Took a look anyway, with these changes it works for me:

```diff
diff --git a/src/a.rs b/src/a.rs
index 7b5eda7..e69de29 100644
--- a/src/a.rs
+++ b/src/a.rs
@@ -1 +0,0 @@
-mod views;
diff --git a/src/main.rs b/src/main.rs
index a22e76e..fb8f534 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,7 +2,7 @@ extern crate askama;
 #[macro_use] extern crate askama_derive;
 use askama::Template;
 mod a;
-
+mod views;
 fn main() {
     let v = views::MyTemplate {_parent: views::MyBaseTemplate{}};
     println!("{}", v.render());
diff --git a/src/views.rs b/src/views.rs
index e94f37d..adb5600 100644
--- a/src/views.rs
+++ b/src/views.rs
@@ -1,3 +1,6 @@
+use std;
+use askama;
+
 #[derive(Template)]
 #[template(path="base.html")]
 pub struct MyBaseTemplate {
```

The first two hunks seemed to be needed because rustc complains that you didn't declare the module correctly. The last one happens, of course, because of this bug (which is only fixed in master).

Eijebong:
HA ! Found it !
I was compiling the real project not the example, that's why you had those errors, sorry...
 I found the real error though, i'm updating the repo now.
The problem is that I have a syntax error in base.html

djc:
Your base template doesn't define any blocks, which is causing it not get recognized as a base template. I added some documentation about this in 2585a4dae0dc88b9814c0c06c9489e2b86a1aa6d.

PR #20, opened 2017-03-19
added filters: int, float, round (#18)

andylibrian:
added built-in filters (#18):
 - int
 - float
 - round


djc:
Can you say more about your use case? I'm not sure `Display` makes sense as the argument type in these cases. I was thinking maybe something more like `Into<isize>` or `num_traits::int::PrimInt`. Any opinions?

djc:
I'll close this for now, since I don't think it makes sense as-is. Feel free to reopen if you have a use case for any of these filters.

PR #21, opened 2017-05-13
Remove duplicate "if/else"

jdanford:


djc:
Thanks for reporting and fixing!

PR #30, opened 2017-08-13
Generate with full path to Result

### LINKED ISSUE #29: wrong number of type arguments
I hit this when upgrading from ffc74a17 to 8bda0d64. I haven't had time to dig in yet, (will try and set aside some time tomorrow if you haven't magically fixed it by then), but I thought I'd give you a quick heads up (i.e. hold off on publishing the updates to crates.io).

```
error[E0244]: wrong number of type arguments: expected 1, found 2
   --> src/templates/mod.rs:139:10
    |
139 | #[derive(Template)]
    |          ^^^^^^^^ expected 1 type argument
```

anowell:
Fixes #29


djc:
Thanks!

Issue #24, opened 2017-08-02
Can't start a template with a { brace


anowell:
Simply starting a template file with a single `{` character (e.g a JSON file) causes the AST and code to be empty:
```
[]                                                                                                                                                                             
impl< 'a > ::askama::Template for ServiceProxyConfig< 'a > {                                                                                                                   
    fn render_to(&self, writer: &mut ::std::fmt::Write) {                                                                                                                      
    }                                                                                                                                                                          
}    
```
Prepend an extra space or newline and the template renders fine.

djc:
Thanks for the report! I fixed it in 82d85bbc20ec4c525d65c4e12e157a934ac2719e and released version 0.3.4 to crates.io, I also added a test case to make sure it doesn't regress.

Let me know if you have any other feedback about the project, please!

anowell:
Thanks! 

Re: feedback - Overall, I'm a huge fan. I'm using it for a port of an internal tool that relies on a handful of Jinja2 templates, and in general, I like it much more than Jinja2 itself. I've had a small handful of cases where askama renders invalid templates as empty or missing content (like this issue) - so I'd like to see basically ANY template error break compilation. I've also opted to flatten my template directory for now because I was tripping over the module path when using template inheritance for templates not in the root of the `templates` directory (might revisit that a bit later). Anyhow, I'll be ramping my usage of askama from my current 4 templates to about 20 over the next couple days, so I might be poking around your issue tracker a bit more ;-)

djc:
Great feedback!

* I'd definitely like Askama to throw an error for invalid templates, rather than rendering empty templates or missing content. I hope the change I did for this issue will help, as it is improves one of the more fragile parts of the parser, but please file bugs also if it doesn't raise an error for templates that you think should be invalid.

* Can you explain more about the template inheritance issues?

* Are there any Jinja features that you missed/had to work around?

* I was a bit surprised about the JSON template. For something like this, I guess I would use serde-json somehow. I was thinking it could make sense to add a feature-gated dependency on `serde-json` that adds a `json` filter, so you can just call `|json` in your templates. Would that be useful for your use cases?

anowell:
> template inheritance issues?

I'll recreate the repro for it in the next day or two and create an issue to discuss it in more detail.

> any Jinja features that you missed/had to work around

So far nothing has jumped out that I really miss (still early though). I can see that I'll need to work around inclusion of partials. In particular, I have a jinja2 template with this:
```
[...]

PR #31, opened 2017-08-13
Adding json and json_pretty filters

anowell:
You mentioned possibly adding a json filter back in [this comment](https://github.com/djc/askama/issues/24#issuecomment-319890853). At the time, I didn't think I'd need it, but it turns out I did need it (to avoid the temptation to do `"{{my_vec.join("\",\"")}}"` for a simple json list). So I went ahead and hacked together an implementation. I'd be happy to tweak the PR in pretty much any way you want. Some notes:

- [ ] I did move `filters.rs` to `filters/mod.rs`. Wasn't sure how you'd prefer it, but I wanted to minimize usage of `cfg` attributes, and thought this cleaner than a nested `mod json { ... mod tests { ... } }` in `filters.rs`.
- [ ] The tests are kinda dumb - really they are just testing `serde_json`'s functionality. Open to ideas.
- [ ] I hate that it can panic (on serialization errors) currently. Now that you've embraced `render()` returning an error, Askama should probably define it's own error type so that you can return `fmt::Error` or `serde::Error`.
- [ ] Is it valuable to have `json` and `json_pretty` filters? I'm only using the pretty variant.
- [ ] Being feature-flagged behind a non-default feature, the docs for these filters won't appear on docs.rs. So there should probably be some other documentation mentioning the `serde-json` feature.
- [ ] I was surprised to see you check in Cargo.lock for a crate lib. Any particular reason?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,6 +10,8 @@ name = "askama"
 version = "0.3.4"
 dependencies = [
  "askama_derive 0.3.4",
+ "serde 1.0.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -20,6 +22,16 @@ dependencies = [
  "quote 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "syn 0.11.11 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
+
+[[package]]
+name = "dtoa"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "itoa"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "libc"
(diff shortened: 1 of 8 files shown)


djc:
This looks nice!

* Agree on moving the filters module(s) to a separate directory.
* It would be nice to have at least one end-to-end test in `testing` that has a `json::Value` in a template struct and just renders it.
* Totally agree that panicking at run-time is not good. I'll do a bit of work to create an appropriate error type. Are you happy with `error-chain` so far? It seems to be the canonical thing to use these days, but I've also found it looking pretty complex and overkill for my simple use cases.
* I don't really like having both filters... this is where optional arguments would be really nice to have an optional `pretty` argument, but without a default value it's not as appealing. I think we should only have a `json` filter now which prettifies by default.
* There appears to be a trick to make it work on `docs.rs`, see [this Cargo.toml](https://github.com/briansmith/ring/blob/master/Cargo.toml).
* I'm not quite sure what the convention is. Looking at a few libraries, I guess none of them have it! I'll remove it.

djc:
69c447f0c7b515648d0a0cb69466c1e71d81a71e removes `Cargo.lock`. 5f3e93da3f1815f24b82b026b95e748267d1b5e8 introduces an Askama `Error` type, using error-chain.

anowell:
Great. I should be able to address all that tonight.

- error-chain: I'm mostly happy with it, particularly for using `chain_err` with string messages in higher-level binaries. For simple libraries where I don't need complex error cause chains, I find myself drawn a bit more toward [quick-error](https://crates.io/crates/quick-error). I guess it depends on whether you prefer `Error` being an enum, or a struct with an `ErrorKind` enum. Either way, I'm happy to remove the panic and add a JSON serialization error.
- I'll add an end-to-end `Value` test, reduce to a single `json` filter (pretty only), and tweak to include json-serde in docs.rs (nice find).
- Also, my understanding of the `Cargo.lock` convention is roughly: commit the `.lock` for applications, but not for libraries.

anowell:
Closed in favor of https://github.com/djc/askama/pull/32

Issue #2, opened 2017-03-06
Re-export custom derives from askama crate


dtolnay:
> Custom derive macros can not be exported together with other items, so you have to depend on a separate crate for it.

They can actually. If you are okay requiring Rust >=1.15 to compile askama, you can include this in askama/src/lib.rs:

```rust
#[allow(unused_imports)]
#[macro_use]
extern crate askama_derive;
pub use askama_derive::*;
```

Then users can write:

```rust
#[macro_use]
extern crate askama;
```

and get access to the derives that way.

djc:
Ah, so a crate can export both, but the crate that contains the proc macro cannot export anything else? As it's currently setup, `askama_derive` depends on `askama`, but I guess circular dependencies are not supported, right?

dtolnay:
I would prefer to drop the dependency of askama_derive on askama by moving the codegen-related code into the derive crate. Then you can re-export the derive from askama and save users the trouble of depending on two crates.

djc:
Yeah, that would be much nicer. It was how I set it up initially, but since I misunderstood the exporting restrictions figured that it made more sense to put all the logic together.

PR #32, opened 2017-08-19
Adding optional json filter (Take #2)

anowell:
I didn't forget about this; just got buried.

The last PR wasn't going to merge cleanly anymore, so I opened a new one rebased from master. I believe this addresses all the discussion items except that the json filter can still panic - which I think will require having all the filters return `Result` - but haven't jumped into that yet.

### DIFF
--- a/testing/Cargo.toml
+++ b/testing/Cargo.toml
@@ -6,7 +6,8 @@ workspace = ".."
 build = "build.rs"
 
 [dependencies]
-askama = { path = "../askama", version = "*" }
+serde_json = "1.0"
+askama = { path = "../askama", version = "*", features = ["serde-json"] }
 
 [build-dependencies]
-askama = { path = "../askama", version = "*" }
+askama = { path = "../askama", version = "*", features = ["serde-json"] }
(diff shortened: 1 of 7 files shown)


testing/Cargo.toml
```diff
@@ -6,7 +6,8 @@ workspace = ".."
 build = "build.rs"
 
 [dependencies]
-askama = { path = "../askama", version = "*" }
+serde_json = "1.0"
```
djc: I think this needs to be `optional = true`?

djc:
ec5a2ebbb375d5ad49daeba72ec6aeb874d4a81e requires filters to return `Result`.

(BTW, feel free to force-push rebased changes to your pull request branches, rather than opening a new PR for them.)

testing/Cargo.toml
anowell: I made it optional in `askama/Cargo.toml` since obviously not all users will want the serde/serde_json dependencies. But that didn't seem as relevant to the end-to-end 'testing' directory. So this change currently requires serde-json and enables the askama serde-json feature in order to run the end-to-end tests. Of course I could wire up the end-to-end tests to also conditionally compile the json tests. Your call. ;-)

anowell:
rebased and force-pushed with `Result` support.

testing/Cargo.toml
djc: Ah, makes sense.

djc:
Merged, thanks for sticking with it!

PR #33, opened 2017-08-24
Allow {% block name %} {% endblock name %}

Eijebong:
This makes converting an app from tera to askama easier.


djc:
Thanks, that seems like a small cost for the convenience.

PR #34, opened 2017-08-24
Add length & count filters

defyrlt:
cc https://github.com/djc/askama/issues/18

### DIFF
--- a/askama/src/filters/mod.rs
+++ b/askama/src/filters/mod.rs
@@ -94,6 +94,16 @@ pub fn trim(s: &fmt::Display) -> Result<String> {
     Ok(s.trim().to_owned())
 }
 
+/// Outputs to length of the argument
+pub fn length<T: AsRef<str>>(s: T) -> Result<usize> {
+    Ok(s.as_ref().len())
+}
+
+/// Alias for the `length()` filter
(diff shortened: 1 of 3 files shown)


djc:
Hey, thanks for the PR! Do you actually use these in your templates? If so, do you use them primarily for string lengths? I'm not sure the conversion through `Display` makes sense for these filters, it seems somewhat wasteful especially if you're going to pass strings to them anyway.

defyrlt:
Yea, personally I use it for strings, can't remember last time I needed filter like that for a non-string. I'll change it to work with `String` instead (or `&str` if possible)

djc:
Maybe try with `AsRef<&str>`, that should cover all the cases?

djc:
And in that case, what about just using `.len()` instead of a filter?

defyrlt:
What if there's a filter that is producing a different string? You can't do something like `(foo|bar).len()`, can you?

djc:
I think that should work, have you tried it?

defyrlt:
Actually, it does work! I'll close this then.

defyrlt:
p.s: a good thing to document!

djc:
Got any suggestions for where/how to document? For me it's fairly obvious that this should work, but obviously I'm way too familiar with the internals.

PR #35, opened 2017-08-24
Enable escaping by default

### LINKED ISSUE #23: Escaping by default for HTML templates
This looks like a really neat project!

Before I would use it in production though, it needs to have escaping be enabled by default for HTML. Otherwise it will inevitably lead to security problems.

See here for what some other template engines do:

* Handlebars: http://sunng87.github.io/handlebars-rust/handlebars/index.html#escaping
* Tera: https://github.com/Keats/tera#autoescaping
* Jinja2 itself: http://jinja.pocoo.org/docs/2.9/templates/#working-with-automatic-escaping

This library doing things at compile-time, maybe there's a different approach you could use, e.g. escaping all `String`s by default, but introducing a new type that wraps a string that will result in unescaped output.

What are your thoughts on this?

defyrlt:
Fixes #23 

Questions:
1. Should `escape` stay being a filter, or that should be moved out and not possible to use?
2. Should `safe` take `&fmt::Display` as an input, or something like `AsRef<&str>` should be enough?
3. Any other cases that should be tested?


djc:
Wow, thanks for working on this. I think I need to sleep on the API for this a little.

defyrlt:
Sounds good.
Looking at travis error - do you have a guess why that is happening? I'm doing exactly what https://github.com/defyrlt/askama/blob/enable-escaping-by-default/askama_derive/src/generator.rs#L232 is doing, unless I'm missing something.

djc:
It probably has something to do with the feature I just pushed in 1c9066f9b40e0f4eb8fd67f6492efc96308eccd4, which you may have missed.

defyrlt:
Indeed, thanks, fixed :)

djc:
I was thinking along these lines:

```diff
diff --git a/askama/src/lib.rs b/askama/src/lib.rs
index 70202f5..943868d 100644
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -258,6 +258,32 @@ pub mod rocket {
     pub use self::rocket::response::{Responder, Response};
 }
 
+use std::fmt::{Display, Formatter};
+
+enum MarkupDisplay<'a, T> where T: 'a + Display {
+    Safe(&'a T),
+    Unsafe(&'a T),
+}
+
+impl<'a, T> From<&'a T> for MarkupDisplay<'a, T> where T: 'a + Display {
+    fn from(t: &'a T) -> MarkupDisplay<'a, T> {
+        MarkupDisplay::Unsafe(t)
+    }
+}
+
+impl<'a, T> Display for MarkupDisplay<'a, T> where T: 'a + Display {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match *self {
+            MarkupDisplay::Unsafe(t) => {
+                write!(f, "{}", filters::escape(&t).map_err(|_| std::fmt::Error)?)
+            },
+            MarkupDisplay::Safe(t) => {
+                t.fmt(f)
+            },
+        }
+    }
+}
+
 // Duplicates askama_derive::path::template_dir()
 fn template_dir() -> PathBuf {
     let mut path = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
```

What do you think? Can we use that to integrate? Hopefully we could use specialization to have `MarkupDisplay` be safe-by-default for some types (integers, for example), and then implement a `safe` filter which turns a `MarkupDisplay` value into an always-`Safe` one.

@mitsuhiko, since you mentioned wanting to work on a Rust-based markupsafe crate before, I'd like to hear your thoughts on what this stuff should look like.

defyrlt:
`safe` filter is something that should be used to indicate that argument is safe and shouldn't be escaped, not to _make_ it safe (as in Jinja). Maybe you mean that `escape` filter should work with `MarkupDisplay` values? If so, then yes, I think we can integrate your example, but specialization isn't stable yet, so it probably doesn't make a lot of sense to implement this right now.

defyrlt:
@djc: any thoughts on this?

djc:
My alternative work in progress is in #43.

defyrlt:
I guess this can be closed :)

PR #36, opened 2017-08-24
Add `unwrap()` to the hello world example in README

defyrlt:
Right now I'm getting this:
```
error[E0277]: the trait bound `std::result::Result<std::string::String, askama::errors::Error>: std::fmt::Display` is not satisfied
```


djc:
Oops, thanks for fixing that!

PR #38, opened 2017-08-25
Don't require for a child template to have every block from its parent

Eijebong:
This allows you to have default blocks in parent templates that can be
ignored in children templates.


djc:
Oh, nice catch! It was definitely intended to work that way, but my tests didn't cover that case. I figured there was probably a simpler way to implement it, and there was; see fc7be3c927d2078a317797c0d5477ad608c4c625. I kept your test however, which still passes, see aeac47cee0e14b9fa38c01082876667f0ec8d874.

Eijebong:
Oh, yeah it's way better this way 👍

PR #39, opened 2017-08-25
Add `join` filter

defyrlt:
Qs:
1. Does that API work, or it should be more/less generic?
2. I think some `AsRef`s are redundant and can be dropped, I'll check it out a little later.


djc:
Oh, tricky one. Maybe something like `&IntoIter<&T> where T: Display` for `input`, and just `&str` for separator?

defyrlt:
I'm targeting this right now
```rust
pub fn join<T, I, M, S>(input: M, separator: S) -> Result<String>
    where T: fmt::Display,
          I: Iterator<Item = T>,
          M: IntoIterator<Item = T, IntoIter = I>,
          S: AsRef<str>
```
to be able to pass in Vec<String>, Vec<&str>, &[&str], &[String] etc, but there are [certain problems](https://play.rust-lang.org/?gist=16f778801aa820cad7991dbc2cbbe9e5&version=stable) - line 27 in the gist is generated by codegen and it gets tricky there. I'll continue investigating that, but it looks like changing the way generator is passing around arguments (to avoid double ref) is not an option.

djc:
I'm not so experienced at type system gymnastics at this level, either, I'm afraid -- it looks generally like the right direction. Maybe ask users.rust-lang.org or #rust for help?

defyrlt:
I did, the answer was basically that this needs to be derefed manually. I might ask again during weekdays. It looks like I can figure out whether filter argument is a reference or not, will give it a shot if there isn't a better way.

defyrlt:
@djc: I pushed a very dirty POC of what I described above in a separate commit.
The basic idea is that it'll dynamically figure out whether filter argument should be referenced by checking what it is already (reference or not). Can you think of cases when this can be harmful?

djc:
I wouldn't be opposed to using more type information if it actually fixed something, but I don't think this gets us much closer to something sound. For one thing, this won't be able to look through type aliases. For another, it only works on context attributes, not on `{% let %}`-defined variables, or method return values. So, if we could get the actual type information we need across the board, I'd be all for it, but I think for now it cannot be more than a very limited hack, which I don't really want.

I do think we should take it to the community and figure out if there's some other way. Conceptually it seems like the compiler should be able to dereference for us always -- but maybe coercion is limited in some way that could be improved in the future. I was trying some quick tests yesterday to see whether there's some way to do a `Deref` impl that does what we need, but couldn't yet get that to work, either.

defyrlt:
Ok, created [this](https://users.rust-lang.org/t/troubles-creating-generic-function-interface-for-iterables-their-references/12595) topic. Let's see if there's a way around it. In the mean time I'll try to see if we can tell whether something is a type alias and if there's a way to work with `{% let %}` defined variables.

defyrlt:
@djc: updated with a suggestion from the forum. custom handlers can probably be separated into something, but it doesn't feel necessary at the moment. what do you think?

djc:
Yeah, this approach seems fine. Can you redo the commits to do away with the fluff? I'd like to have one commit that refactors the generator code to arrive at the structure you ended up with (separating out `_visit_format`, etc) without adding functionality, then add a commit that adds the new filter + tests after that.

defyrlt:
@djc: updated, did you get a chance to look at it?

djc:
I had missed it, sorry! No notification when new commits are pushed... Looks good, though, thanks for rebasing!

PR #42, opened 2017-08-29
Add support for Shio

mehcode:
@djc I wouldn't merge this until Shio hits 0.1. I'll keep it up-to-date until then.

### DIFF
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -20,6 +20,7 @@ default = []
 serde-json = ["askama_shared/serde-json"]
 with-iron = ["iron", "askama_derive/iron"]
 with-rocket = ["rocket", "askama_derive/rocket"]
+with-shio = ["askama_derive/shio"]
 
 [dependencies]
 askama_derive = { path = "../askama_derive", version = "0.3.4" }
--- a/askama_derive/Cargo.toml
+++ b/askama_derive/Cargo.toml
@@ -15,6 +15,7 @@ proc-macro = true
 default = []
 iron = ["askama_shared/iron"]
 rocket = ["askama_shared/rocket"]
+shio = ["askama_shared/shio"]
 
 [dependencies]
 askama_shared = { version = "0.3.4", path = "../askama_shared" }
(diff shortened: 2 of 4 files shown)


djc:
So what exactly is the purpose of opening this, just to get my feedback? It looks fine to me, would be happy to merge this as soon as you're ready for it. Do you have some idea about the timeline?

mehcode:
@djc Should have clarified. Just to get your feedback really. I can close it if you don't want it to remain open for probably a couple weeks.

---

I have a couple of things I was going to edit the PR with and list but I'll list them here:
 - Why [depend and re-export](https://github.com/djc/askama/blob/master/askama/src/lib.rs#L240-L245)? If I as a user of Askama, upgrade Iron/Rocket/Shio then Askama will no longer work for me (as you'll be implementing a `Responder`/`Modifier` trait of an _older_ version). This sounds like a use case for something like node.js's `peerDependencies`.

 - Why [`render().unwrap()`](https://github.com/djc/askama/blob/9f3b590206e3dfe33b7129b1c8ff010f60318cf2/askama_shared/src/generator.rs#L659)? I returned the error in Shio and there must be error propagation in the other frameworks.

djc:
Good questions!

Deriving implementations for structs means that you can add code for an item to that item's module, but there's way to "communicate" between different derivations or anything. That means that each derive must stand on itself, but also that it must not conflict with other derivations in the same module. It also means the derivations inject stuff into a scope owned by the author of the original item. So, I want to inject as few things as possible, and I need to make sure I don't conflict with anything they imported into their scope. I also don't want any duplicate imports. I think the depend and re-export strategy is a pretty good solution to that.

Talking about incompatibilities actually makes me think like this is an even better strategy, because this way Askama gets to clearly declare it's dependency on a particular version of Rocket/Iron/Shio that is known to work, and if the Askama user wants to pull in a different version, there might be some error raised.

The `unwrap()` is clearly suboptimal, I was just trying to get something working so far. I really should put in the work to raise a proper error value for that case.

djc:
Just noticed shio is at 0.2.0 already. Time to revive this?

djc:
@mehcode one more ping?

mehcode:
@djc Sorry for leaving you hanging here. I'm going to be working on Shio again shortly. I'll send you an updated PR once I do some clean up work.

PR #43, opened 2017-09-02
Escape by default

djc:


askama_shared/src/lib.rs
```diff
@@ -19,6 +19,47 @@ pub use generator::generate;
 mod generator;
 mod parser;
 
+use std::fmt::{self, Display, Formatter};
+
+pub enum MarkupDisplay<'a, T> where T: 'a + Display {
+    Safe(&'a T),
```
defyrlt: I think you don't need to store a reference to `T`, you can just make it `T`. In that case (`T: Display`) it can store both `T` and `&T` (as well as `&mut T`), because `Display` [is implemented for references to `T`](https://play.rust-lang.org/?gist=09f16050ae06f899b588bb59db6a8ab8&version=stable).

askama_shared/src/generator.rs
```diff
@@ -353,9 +353,11 @@ impl<'a> Generator<'a> {
 
     fn write_expr(&mut self, ws: &WS, s: &Expr) {
         self.handle_ws(ws);
-        self.write("writer.write_fmt(format_args!(\"{}\", ");
+        self.write("let askama_expr = &");
```
defyrlt: does this need to be a reference?

askama_shared/src/generator.rs
```diff
@@ -353,9 +353,11 @@ impl<'a> Generator<'a> {
...
         self.visit_expr(s);
-        self.writeln("))?;");
+        self.writeln(";");
+        self.writeln("writer.write_fmt(format_args!(\"{}\", \
+                      &::askama::MarkupDisplay::from(askama_expr)))?;");
```
defyrlt: same question here - does this need to be referenced? I think `format_args!` will figure it out for you anyway.

askama_shared/src/lib.rs
```diff
@@ -19,6 +19,47 @@ pub use generator::generate;
...
+
+impl<'a, T> From<&'a T> for MarkupDisplay<'a, T> where T: 'a + Display {
+    fn from(t: &'a T) -> MarkupDisplay<'a, T> {
+        MarkupDisplay::Unsafe(t)
+    }
+}
+
+impl<'a, T> From<usize> for MarkupDisplay<'a, T> where T: 'a + Display {
+    fn from(t: usize) -> MarkupDisplay<'a, T> {
+        MarkupDisplay::Unsafe(t)
```
defyrlt: should this be `Safe` instead?

djc: It should, but this won't work anyway -- because we don't have specialization on stable.

askama_shared/src/generator.rs
djc: For example, `visit_expr()` might result in `self.x` where `x` is defined as a `String`. We just want to take a reference to that member and pass it to the formatting function.

askama_shared/src/generator.rs
djc: I've typically found that the formatting functions only want references; they don't take ownership. This makes sense to me, so it would be good to make that clear in the code.

djc:
@defyrlt, see my new changes -- this actually passes the existing tests. Care to review?

djc:
Remaining work: escaping by default should probably be optional somehow. One way to make it nicer would be to select on template file extensions, so that "foo.html" escapes, but "foo.txt" does not.

defyrlt:
Looks good to me :+1:

I think escaping setting based on file extension can be misleading. People can just store their full templates or inner parts or maybe email parts in something different from `.html` and then be very surprised by the fact that auto escaping is disabled for such files. I think a good place for it would be in `#[template]`, as a parameter.

djc:
That could be annoying to do for all the template in a project, though. Anyway, for now it's probably a fine choice.

defyrlt:
Agreed. Maybe there's a way to have create-level or module-level setting, while keeping template-specific setting. But I feel like it should be explicit.

Issue #52, opened 2017-09-24
Whitespace suppression not honored for macro blocks


larros:
Whitespace before and after macro blocks are not suppressed even though the suppression is used. The parser supports the syntax but the macro nodes are excluded from the code generation. The solution to this is to include macro nodes in code generation but only for the purpose of whitespace supression.

Example: 
``` rust
#[derive(Template)]
#[template(source = "{% macro foo(bar) %}{% endmacro -%}\nafter", ext="txt")]
struct SuppressTemplate {
}

#[test]
fn test_suppress() {
    let t = SuppressTemplate{};
    assert_eq!(t.render().unwrap(), "after");
}
```
Fails with:
´´´
'assertion failed: `(left == right)`
  left: `"\nafter"`,
 right: `"after"`
´´´

djc:
Sounds good to me, do you want to create a PR for this? One thing to think about is how to error out on templates that put macro definitions on anywhere but the top-level -- that needs to be preserved (ideally, tested).

larros:
I can take a look at this after the import is completed. The import nodes have the the same problem as macros.

djc:
Thanks, that would be great!

larros:
I've started looking at this now and I think I know what the end result should be. I just want some confirmation.

These are my current thoughts
- The outer white space rules on the macro node should apply to the macro declaration site
- The inner white space rules on the macro node should apply to the call site
- The white space rules on the call node should apply to the call site

djc:
That sounds exactly right!

PR #55, opened 2017-10-01
Fix whitespace handling for macros and imports

larros:
Resolves https://github.com/djc/askama/issues/52.

I'm not completely sure about the flush/prepare usage but I had to change write_call for the test cases to pass.


askama_shared/src/generator.rs
```diff
@@ -353,11 +353,21 @@ impl<'a> Generator<'a> {
...
+                    }
+                    self.write_macro(m);
+                },
+                Node::Import(ref ws, _) => {
+                    if let AstLevel::Nested = level {
+                        panic!("import blocks only allowed at the top level");
+                    }
+                    self.write_import(ws);
+                },
                 Node::Extends(_) => {
```
djc: Should we also deal with the whitespace from the `Extends` nodes, to get it really right? (It doesn't currently parse a `WS`, but maybe it should to be totally consistent.)

askama_shared/src/generator.rs
```diff
@@ -436,7 +446,7 @@ impl<'a> Generator<'a> {
 
     fn write_call(&mut self, state: &'a State, ws: &WS, name: &str, args: &[Expr]) {
         let def = state.macros.get(name).expect(&format!("macro '{}' not found", name));
-        self.handle_ws(ws);
+        self.flush_ws(ws);
```
djc: This didn't make sense to me at first, but I think the need for this change is explained by the preparing/flushing `ws1` and `ws2` from the macro definition.

askama_shared/src/generator.rs
```diff
@@ -353,11 +353,21 @@ impl<'a> Generator<'a> {
                     self.handle_include(state, ws, path);
                 },
                 Node::Call(ref ws, name, ref args) => self.write_call(state, ws, name, args),
-                Node::Macro(_, _) |
-                Node::Import(_, _) |
+                Node::Macro(_, ref m) => {
+                    if let AstLevel::Nested = level {
+                        panic!("macro blocks only allowed at the top level");
+                    }
+                    self.write_macro(m);
```
djc: Can this just be `self.handle_ws(&m.ws)` or some such? (And same for the `Import` case.)

testing/templates/macro.html
```diff
@@ -1,5 +1,13 @@
...
+
+{{ param }} {{ param }} {{ param }}
+
+{%- endmacro -%}
+
+2
+
+{%- call thrice(s) -%}
+
+3
```
djc: Nit: please add a trailing newline here.

askama_shared/src/generator.rs
larros: I don't think whitespace handling is applicable to ```Extends``` because the only content that is used from a child template is from inside blocks. I did some tests and from what I can see whitespace in the child template outside of blocks don't have any impact on the generation. 

It might be a good idea though to support ```{%-``` and ```-%}``` in the parser to standardize the syntax, even though it doesn't have any impact

larros:
No, I'm not at RustFest. Maybe next time.

askama_shared/src/generator.rs
djc: Ha, good point! Okay, then we can at least ignore it for the purposes of this PR. Or, maybe add a comment explaining this?

djc:
Great, thanks!

djc:
Added some explanatory comments in cb6d450 and 6c7da56.

Issue #40, opened 2017-08-25
Can't include a file that has macro definitions in it


Eijebong:
Hey,

It would be great to be able to have macros in a file and be able to include that file in another.
Currently it's hitting `panic!("no extends or macros allowed in content");`.

djc:
Not sure how you're trying to put to your macro definitions to get that error, but that error will remain in there even if you can import macro definitions from another file. It indicates that you're trying to define a macro somewhere not at the top-level of a template file, is that what you're doing?

Eijebong:
Yeah I have a macros.html file with macros definitions in it and am including it in another template.

Do you think there is a better way to fix that than changing how include currently works and parse the included file at the beginning and include its nodes in the current template nodes ? (parse it here and append nodes to the current nodes somehow) https://github.com/djc/askama/blob/master/askama_derive/src/generator.rs#L17

djc:
Jinja very clearly separates `include` and `import`, as far as I can tell, where `include` basically renders the included file into the including template while `import` makes the macros from the imported template available to the importing template. So your use case seems to need an import mechanism. In terms of implementation, it could perhaps function similarly to the handling for `defmacro`, in the sense that top-level `import` blocks are removed from the AST and instead result in having some macros added to the macro map.

djc:
Thanks to @larros for fixing this!

PR #58, opened 2017-10-15
Add scope when importing macros

larros:
Solves https://github.com/djc/askama/issues/40


larros:
A simpler solution than what I did would be to just have "scope::name" as a string instead of keeping them separated but that would require the creation of owned Strings in the parser instead of everything just being slices of the source.

djc:
No, this is pretty much what I would have done, thanks a lot! I've added some minor line-wrapping in aa18a4c1598b1cf530e37fa73bceb06dad43afdc and merged it into master.

PR #60, opened 2017-11-21
Bump version in README

abdullahibnnadjo:
Update askama version in the README.
Makes the example below compile and work


djc:
Nice catch, thanks!

PR #64, opened 2018-01-17
Content-type inference for Iron support

hajifkd:
I feel that it would be better to set Content-type header for html templates in Iron support.
Just like `askama_shared::input::EscapeMode`, the file extension is checked to determine whether the template is html or not. For html templates, I have added a code to add Content-type header in `modify` method.
I have also added the test to check this works properly.


djc:
It's too bad Iron doesn't have anything like Rocket's `ContentType::from_extension()`, but I guess this is a net improvement. Thanks for contributing!

hajifkd:
Thank you for merging!

PR #66, opened 2018-01-29
`build = "build.rs"` is no longer necessary

pwoolcoc:
For all rust versions >= 1.19, cargo will assume `build = "build.rs"` if it sees a `build.rs` file in the same directory as the `Cargo.toml`


djc:
Nice, thanks!

PR #77, opened 2018-04-12
Isolate proc-macro dependency to the proc macro

### LINKED ISSUE #73: When using Askama at revision 36cbfd4bad8c3f534c8cb13f9ddbc463e1c44ef5, binaries produced by Rust nightly-2018-04-07 dynamically link libstd and libproc_macro which cannot be found
I originally misattributed this issue first as https://github.com/japaric/cross/issues/193 and then as https://github.com/djc/askama/issues/72 but I have now found conclusively that simply using Askama at revision 36cbfd4bad8c3f534c8cb13f9ddbc463e1c44ef5 with Rust nightly-2018-04-07 causes the following problem: The produced binary gets dynamically linked to libstd and libproc_macro which cannot be found, resulting in it being possible to run the binary with `cargo run` but not by direct execution of `./target/debug/hello_askama` (hello_askama is what I named the minimal example that is made based on the Askama README) as in https://github.com/rust-lang/cargo/issues/4651.

```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/hello_askama`
Hello, world!
```

```
$ ./target/debug/hello_askama
./target/debug/hello_askama: error while loading shared libraries: libproc_macro-d3bd4d284648dcab.so: cannot open shared object file: No such file or directory
```

```
$ ldd target/debug/hello_askama
	linux-vdso.so.1 =>  (0x00007fff265d1000)
	libproc_macro-d3bd4d284648dcab.so => not found
	libstd-b70fe684839d0aca.so => not found
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3db94ba000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f3db92a3000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3db8ec3000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f3db9921000)
```

I used a `Cargo.toml` that looks like this:

```
[package]

name = "hello_askama"
version = "0.0.1"
authors = [ "Your name <you@example.com>" ]

[dependencies]
askama = { git = "https://github.com/djc/askama", rev = "36cbfd4bad8c3f534c8cb13f9ddbc463e1c44ef5" }

[build-dependencies]
askama = { git = "https://github.com/djc/askama", rev = "36cbfd4bad8c3f534c8cb13f9ddbc463e1c44ef5" }
```

`build.rs` like in Askama README:

```
extern crate askama;

fn main() {
    askama::rerun_if_templates_changed();
}
```

`templates/hello.htm` like in Askama README:

```
Hello, {{ name }}!
```

`src/main.rs` like in Askama README:

```
#[macro_use]
extern crate askama;

use askama::Template;

#[derive(Template)]
#[template(path = "hello.htm")]
struct HelloTemplate<'a> {
    name: &'a str,
}

fn main() {
    let hello = HelloTemplate { name: "world" };
    println!("{}", hello.render().unwrap());
}
```

And like I said, I was using Rust nightly-2018-04-07.

```
rustup default nightly-2018-04-07
```

dtolnay:
The dependency of askama on askama_shared on syn on proc_macro used to mean libproc_macro would be dynamically linked into any crate using askama. We want only askama_derive to have a runtime dependency on proc macro. This commit moves all proc macro code from askama_shared into askama_derive so that the askama crate no longer dynamically links libproc_macro.

Fixes #73. Closes https://github.com/dtolnay/syn/issues/400.


dtolnay:
(This is a breaking change for askama_shared.)

PR #90, opened 2018-05-26
Fix the base template

david-sawatzke:
```{{ block title%}``` is simply invalid syntax and I have no idea what ```{{ title }}``` is for or how to make it work.

Probably a botched Copy-Paste-Job.

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -103,7 +103,7 @@
 //! <!DOCTYPE html>
 //! <html lang="en">
 //!   <head>
-//!     <title>{{ block title %}{{ title }}{% endblock %} - My Site</title>
+//!     <title>{% block title %}{% endblock %} - My Site</title>
 //!     {% block head %}{% endblock %}
 //!   </head>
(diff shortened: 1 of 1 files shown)


djc:
Why did you close it? :) Looked like thing that needs fixing!

david-sawatzke:
I was unsure if the ```{{ title }}``` tag didn't belong there and wanted to experiment a bit further, since it does work like that in Jinja. But since compilation fails for the base struct and there is no further further documentation, I think it's an oversight.

djc:
Please take a look at the contents of `testing/tests/inheritance.rs`. These are actually tested so should work.

david-sawatzke:
In the tests, the ```{{ title }}``` tag is outside the block, while it's inside the block in the docs for the template. If it's outside the block, it works fine, but if it's inside, the rust compiler complains with:

> error[E0609]: no field `title` on type `&Self`

djc:
@david-sawatzke so is that what you are actually trying to achieve?

PR #91, opened 2018-05-26
[WIP] Make "self" available to templates

Smibu:
This PR makes it possible to call methods of the template context struct in templates.

An unresolved question is how to deal with potentially infinite recursion (causing stack overflow) as demonstrated by the test `test_self` in `methods.rs`.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -147,7 +147,9 @@ impl<'a> Generator<'a> {
     }
 
     fn default<'n>() -> Generator<'n> {
-        Self::new(SetChain::new(), 0)
+        let mut chain = SetChain::new();
+        chain.insert("self");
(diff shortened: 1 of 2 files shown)


djc:
Thanks! I've changed your commit to remove the failing test and the fix and rebased it as 0330699a337f4bd86854eaafccc7128117f10b14. Before it, I've made a different fix limited to exposing `self` only for method calls in ecae242714875524c421945cbaf0ce948dc5754a. I'm going to try and work the risk of stack overflows into the documentation somehow.

djc:
I've described the pitfall here: ac4aaf6503f7e13641cbdd8fba2c4a10b81b1f97. Do you think that makes sense, or is there a better way of describing it?

Smibu:
Thanks! The documentation update seems fine.

Issue #92, opened 2018-05-31
Provide a `Sync` error (aka don't use error-chain)


dathinab:
Errors from `error-chain` are not `Sync`,  which can cause problems in some cases. E.g. if you want to/have to use `Failure` which requires `Send+Sync` bounds.

As `askama` doesn't really have many errors it should be simple.

---
Currently askama has two foreign links and that's it:

1. `::std::fmt::Error` (which is zero-sized)
2. [feature] `::serde_json::Error`  (which is a custom error impl)

I think it might be the best to roll a simple error enum, as people might use either `failure` or `error-chain` or custom errors and a simple error enum is compatible  with all this ways (note that you would have to make
it non-exhaustive)

dathinab:
I also noticed that it does not expose the `ErrorKind` making it impossible
to match the error kinds and convert it to a wrapper implementing Sync.

djc:
Sounds sensible. Would you be able to submit a PR?

dathinab:
Yes, probably not today but in the next 1-2 days.

djc:
Great, thanks!

djc:
Fixed by 12c920d7e34dae4934d52072ab9d8cb271d71058.

PR #93, opened 2018-06-04
Converted error-chain based error to enum based error

dathinab:
- now implements `Send` + `Sync` + `'static`
- still implements `std::error::Error`, `Debug`,
  `Display`, `From<std::fmt::Error>`,
  `From<::serde_json::Error>`

- closes issue #92


djc:
LGTM, thanks!

PR #97, opened 2018-06-29
Slightly improve TemplateInput code

mashedcode:
No new feature here.
I simply read the source and happend to touch some code while trying to understand the reason why things are done the way they're done.
Some changes may be questionable but I thought a PR won't hurt.


djc:
Thanks for looking at the code and submitting your changes! There are quite a few changes here all in a single commit, and I'm not sure I understand the purpose/rationale for what you're trying to do here. Can you explain your changes a bit more and say something about what you wanted to achieve?

The change to carry `parent_type` as a `String` makes sense, I think, and I also like the idea of only keeping the extension in the path (since it's now there anyway even for source-based templates). I'm generally not such a fan of single-use functions.

askama_derive/src/input.rs
```diff
@@ -4,41 +4,22 @@ use quote::ToTokens;
...
-                }
-            }
-        }
-
-        let meta_list = match meta.expect("no attribute 'template' found") {
-            syn::Meta::List(inner) => inner,
-            _ => panic!("attribute 'template' has incorrect type"),
-        };
-
+        let meta_list = meta_list_from(ast);
```
mashedcode: The `TemplateInput::new` method seemed way to complicated and contain a lot of stuff so I moved stuff out.

askama_derive/src/input.rs
```diff
@@ -88,61 +69,76 @@ impl<'a> TemplateInput<'a> {
...
 
         TemplateInput {
             ast,
             source,
             print,
-            escaping,
-            ext,
-            parent,
+            escaping: escaping.unwrap_or(EscapeMode::from(&path)),
+            parent: field_type(ast, "_parent"),
```
mashedcode: This was moved to a function called `field_type` because it took me quite some time to figure out that this gets the type of the `_parent` field at all.
This is another example of single-use you mentioned. It's not likely and "efficient" to reuse the `field_type` function but now the intend is clear and it's better than adding a comment or renaming `parent` to `parent_type` IMHO.

askama_derive/src/input.rs
```diff
@@ -88,61 +69,76 @@ impl<'a> TemplateInput<'a> {
...
             path,
         }
     }
+
+    pub fn source(&self) -> String {
```
mashedcode: I feel like the caller shouldn't care about how it gets the source so this implementation detail got moved here. In a separate function to avoid another clone.

askama_derive/src/input.rs
```diff
@@ -88,61 +69,76 @@ impl<'a> TemplateInput<'a> {
...
         };
 
         TemplateInput {
             ast,
             source,
             print,
-            escaping,
-            ext,
-            parent,
+            escaping: escaping.unwrap_or(EscapeMode::from(&path)),
```
mashedcode: Simplify and make it clearer which value has priority by looking at the code.

askama_derive/src/input.rs
```diff
@@ -88,61 +69,76 @@ impl<'a> TemplateInput<'a> {
...
-            _ => panic!("derive(Template) only works for struct items"),
-        }.next();
-
         let path = match source {
-            Source::Source(_) => match ext {
-                Some(ref v) => PathBuf::from(format!("{}.{}", ast.ident, v)),
-                None => PathBuf::new(),
-            },
-            Source::Path(ref s) => path::find_template_from_path(s, None),
+            Source::Source(_) => PathBuf::from(format!("{}.{}", ast.ident, ext.unwrap())),
```
mashedcode: `ext` is guaranteed to be defined since it's checked above.

askama_derive/src/input.rs
djc: I don't like this part much. I like that the `TemplateInput` is just a well-typed representation of the input AST, that represents all of the `template()` attribute stuff as well as anything the code generator might need from the AST. Then `build_template()` can be responsible for taking that input and working out what needs to be done with it, including getting the source from the `path` if necessary. I'd also prefer to keep that in `lib.rs` for the symmetry with included templates, for which a symmetric operation happens in `find_used_templates()`.

askama_derive/src/input.rs
```diff
@@ -167,6 +163,17 @@ impl From<String> for EscapeMode {
     }
 }
 
+impl<'a> From<&'a PathBuf> for EscapeMode {
```
djc: I don't think using `From` here is idiomatic, it is not really a type conversion. I think it might make sense to add a `from_path()` to an `impl` block for `EscapeMode`, though.

askama_derive/src/input.rs
djc: Yeah, I don't think these changes will make the code easier to read, generally. It would be good to add some documentation to the `TemplateInput` code to clarify the flow, though!

askama_derive/src/input.rs
```diff
@@ -88,61 +69,76 @@ impl<'a> TemplateInput<'a> {
         match (&source, ext.is_some()) {
             (&Source::Path(_), true) => {
                 panic!("'ext' attribute cannot be used with 'path' attribute")
-            }
+            },
```
djc: `rustfmt` style says no commas after `{}` blocks in match arms.

askama_derive/src/input.rs
djc: This should probably use `unwrap_or_else()` (in order to only evaluate the argument in the case that escaping is actually `None`).

askama_derive/src/input.rs
djc: Sure it is defined, but it might still be `None`. I prefer to avoid `unwrap()` on principle if possible. In this case, it might make sense to merge the `match (&source, ext.is_some())` with this `match` in a way that the `ext` can be pushed into the source.

askama_derive/src/input.rs
djc: Adding more indirection doesn't necessarily make it less complicated. I would prefer adding good comments to indicate the structure and goals, instead.

mashedcode: Well, I don't want to start an argument here. So I'll revert this. But I'm a huge fan of functions actually fitting on my screen. (It still doesn't.)

askama_derive/src/input.rs
mashedcode: Thanks!

askama_derive/src/input.rs
mashedcode: I think I meant to say it's guaranteed to not be `None` (to be `Some`). But your last sentence is a good idea.

mashedcode:
Okay, so here're the I think you liked in separate commits.

djc:
Thanks for cleanly splitting the commits, this makes it much easier to review! I like all of these except for the "field type determination" one. I understand that the transformation is a bit convoluted, but I think adding some comments would be a better improvement here than your code changes, which, not unimportantly, add quite a bit to the apparent length of the code.

Also, the way you've done the commits here means the git commit metadata does not contain your a name or a usable email address. Would you mind changing that? I would like to at least be able to contact you at a later time (for example, should a relicensing discussion ever come up).

mashedcode:
In regards to the "field type determination". If you look at it closely the change it that it uses `find` instead of `filter_map` and can therefore return early.

If contacting me is an issue for you due to licencing I can simply sign off my commits to this project so that you can make changes to the license without ever having to contact me.

djc:
I also made some further improvements in 9021c229b72861e4b5fc49260819643eb92d8cb6 and 3ba96cabca53aa660f3dc935e3fd3e58eeb7b142.

mashedcode:
@djc Nice thanks for teaching me rust.

PR #99, opened 2018-06-29
Do not put a trailing comma after a block based match arm

mashedcode:
You said:
> [`rustfmt` style says no commas after `{}` blocks in match arms.](https://github.com/djc/askama/pull/97#discussion_r199143221)

So let's configure it that way. See [match_block_trailing_comma](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#match_block_trailing_comma).

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,4 +1,3 @@
 brace_style = "PreferSameLine"
 fn_args_density = "Compressed"
-match_block_trailing_comma = true
 use_try_shorthand = true


djc:
Stable rustfmt doesn't support configuration, and I don't see any good reason to diverge from standard style on this point.

mashedcode:
IIUC stable rustfmt is deprecated and everyone is using rustfmt-nightly.
Also by removing that line this project would conform to the standard style. See the link in my original post.
It simply seems to be reconfigured here.

djc:
Ah, sorry, I forgot that the config was even there. In this case, please submit a PR to just remove it (or I can take care of it later).

PR #103, opened 2018-07-04
Add possibility to specify template directories

### LINKED ISSUE #81: Forced template root dir
> The path is interpreted as relative to the templates dir in the directory where the originating crate's Cargo.toml resides. 

Why? This forces me a 1 way structure, that works bad with my project tree. Please add an option to change the root dir, or better to add multiple ones.

mashedcode:
Resolves #81 

I didn't do extensive testing on this because I don't actually use askama in production yet. But it should work. I think.

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -393,7 +393,9 @@ fn visit_dirs(dir: &Path, cb: &Fn(&DirEntry)) -> io::Result<()> {
 /// that have templates, to make sure the crate gets rebuilt when template
 /// source code changes.
 pub fn rerun_if_templates_changed() {
-    visit_dirs(&path::template_dir(), &|e: &DirEntry| {
-        println!("cargo:rerun-if-changed={}", e.path().to_str().unwrap());
-    }).unwrap();
+    for template_dir in path::template_dirs().iter() {
+        visit_dirs(template_dir, &|e: &DirEntry| {
+            println!("cargo:rerun-if-changed={}", e.path().to_str().unwrap());
+        }).unwrap();
+    }
 }
--- a/askama_shared/Cargo.toml
+++ b/askama_shared/Cargo.toml
@@ -10,10 +10,12 @@ workspace = ".."
 
 [features]
 default = []
-serde-json = ["serde", "serde_json"]
+serde-json = ["serde_json"]
 iron = []
 rocket = []
 
 [dependencies]
-serde = { version = "1.0", optional = true }
+serde = "1.0"
+serde_derive = "1.0"
 serde_json = { version = "1.0", optional = true }
+toml = "0.4"
--- a/askama_shared/src/path.rs
+++ b/askama_shared/src/path.rs
@@ -3,6 +3,8 @@ use std::fs::File;
 use std::io::Read;
 use std::path::{Path, PathBuf};
 
+use toml;
+
 pub fn get_template_source(tpl_path: &Path) -> String {
     let mut path = template_dir();
     path.push(tpl_path);
@@ -50,6 +52,36 @@ pub fn template_dir() -> PathBuf {
     let mut path = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
     path.push("templates");
     path
+}
+
+#[derive(Deserialize)]
+struct Config {
+    dirs: Option<Vec<String>>,
+}
+
+pub fn template_dirs() -> Vec<PathBuf> {
+    let root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    let askama = root.join("askama.toml");
+    let default = vec![root.join("templates")];
+    if askama.exists() {
+        let mut contents = String::new();
+        File::open(askama).unwrap().read_to_string(&mut contents).unwrap();
+        let config: Config = toml::from_str(&contents).unwrap();
+        if let Some(dirs) = config.dirs {
+            let paths: Vec<PathBuf> = dirs.into_iter().map(|directory| {
+                root.join(directory)
+            }).collect();
+            if paths.len() > 0 {
+                paths
+            } else {
+                default
+            }
+        } else {
+            default
+        }
+    } else {
+        vec![root.join("templates")]
+    }
 }
 
 #[cfg(test)]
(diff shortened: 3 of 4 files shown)


djc:
This looks pretty good! I have a bunch of minor concerns, will do a detailed review tomorrow.

askama/src/lib.rs
```diff
@@ -393,7 +393,10 @@ fn visit_dirs(dir: &Path, cb: &Fn(&DirEntry)) -> io::Result<()> {
 /// that have templates, to make sure the crate gets rebuilt when template
 /// source code changes.
 pub fn rerun_if_templates_changed() {
-    visit_dirs(&path::template_dir(), &|e: &DirEntry| {
+    let trigger_rerun = &|e: &DirEntry| {
```
djc: Why pre-define this here? Did you try just passing the closure inside the loop?

askama_shared/Cargo.toml
```diff
@@ -10,10 +10,12 @@ workspace = ".."
 
 [features]
 default = []
-serde-json = ["serde", "serde_json"]
+serde-json = []
```
djc: So `serde` is now required for deserializing the configuration, but I think `serde_json` should still be optional, as before, no?

askama_shared/src/path.rs
```diff
@@ -52,6 +54,36 @@ pub fn template_dir() -> PathBuf {
...
 }
 
+#[derive(Deserialize)]
+struct Config {
+    dirs: Option<Vec<String>>,
+}
+
+pub fn template_dirs() -> Vec<PathBuf> {
+    let root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    let askama = root.join("askama.toml");
```
djc: I would like to move `"askama.toml"` into a constant, maybe named `CONFIG_FILE_NAME`. Then, please move the code to instantiate a `Config` into a `Config::from_file()` method (which doesn't need any arguments).

Also, here and in other places, please don't override the configuration if it provides an empty list. In that case, it should probably just do nothing (this would allow users to enforce the use of `source`-based templates only, I guess).

askama_shared/src/path.rs
```diff
@@ -52,6 +54,36 @@ pub fn template_dir() -> PathBuf {
...
+            if paths.len() > 0 {
+                paths
+            } else {
+                default
+            }
+        } else {
+            default
+        }
+    } else {
+        vec![root.join("templates")]
```
djc: This seems to be a duplicate of the `default` defined earlier.

askama_shared/src/path.rs
```diff
@@ -5,9 +5,17 @@ use std::path::{Path, PathBuf};
...
+        .find(|dir| dir.join(path).exists())
+        .map(|dir| dir.join(path))
+}
+
 pub fn get_template_source(tpl_path: &Path) -> String {
-    let mut path = template_dir();
-    path.push(tpl_path);
+    let dirs = template_dirs();
+    let path = search_template_in_dirs(&dirs, tpl_path)
+        .expect(&format!("Template file '{}' does not exist.", tpl_path.to_str().unwrap()));
```
djc: Nit: elsewhere I use sentence fragments for errors, not full sentences. Please don't capitalize, and don't end with a period.

askama_shared/src/path.rs
```diff
@@ -5,9 +5,17 @@ use std::path::{Path, PathBuf};
 
 use toml;
 
+fn search_template_in_dirs(dirs: &Vec<PathBuf>, path: &Path) -> Option<PathBuf> {
```
djc: Nit: this can take a slice instead of a `Vec`, so `&[PathBuf]`.

askama_shared/src/path.rs
```diff
@@ -32,34 +32,23 @@ pub fn get_template_source(tpl_path: &Path) -> String {
...
 
 pub fn find_template_from_path(path: &str, start_at: Option<&Path>) -> PathBuf {
-    let root = template_dir();
+    let dirs = template_dirs();
     if let Some(rel) = start_at {
-        let mut fs_rel_path = root.clone();
-        fs_rel_path.push(rel);
-        fs_rel_path = fs_rel_path.with_file_name(path);
+        let root = search_template_in_dirs(&dirs, rel).unwrap();
+        let fs_rel_path = root.join(rel).with_file_name(path);
```
djc: One question is whether the `start_at` paths here are already rooted (that is, are already absolute paths starting with the crate root) or not. Probably need to audit all uses to see whether usage is consistent or not.

askama_shared/src/path.rs
```diff
@@ -32,34 +32,23 @@ pub fn get_template_source(tpl_path: &Path) -> String {
 }
 
 pub fn find_template_from_path(path: &str, start_at: Option<&Path>) -> PathBuf {
-    let root = template_dir();
+    let dirs = template_dirs();
     if let Some(rel) = start_at {
-        let mut fs_rel_path = root.clone();
-        fs_rel_path.push(rel);
-        fs_rel_path = fs_rel_path.with_file_name(path);
+        let root = search_template_in_dirs(&dirs, rel).unwrap();
```
djc: Not sure `unwrap()` makes sense here?

mashedcode: I think since `start_at` is defined this is a path that already existed previously.

askama_shared/src/path.rs
mashedcode: Just preserving existing behavior. If `start_at` is already rooted the relative path should not exist and it should slip through.

djc: Okay, but can we at least replace the `unwrap()` with an `expect()` that clearly states the problem?

askama_shared/src/path.rs
djc: This should also `expect()` instead of `unwrap()`.

djc:
@mashedcode anything else you want to change? Otherwise I'll just merge this and iterate on the design a little bit. Thanks for working on this!

djc:
If you're interested, here's a bunch of the follow-up I had in mind: https://github.com/djc/askama/compare/b5b1589...f74d121. There's probably a bit more, I wanted to store a `Config` in `Input` and move `find_template_from_path()` into `Config`.

PR #104, opened 2018-07-21
Implement Responder trait for actix-web

ryanmcgrath:
Heya!

Wound up needing this and figured I'd throw it over as a pull request. It's built for `actix-web` 0.7 which went out less than 12 hours ago, _but_ I originally built it against `0.6.x` and it should work there too ('s why I include the `mime_guess` crate rather than rely on `actix-web`'s `fs::{}` methods).

Test passes, but I couldn't figure out how you were opting to run against different web frameworks (Iron/Rocket/etc) so it needs to be enabled to run. Hoping this is the ~90% that lets you do the last 10% to make it correct per your package design, or something.

Cheers!

### DIFF
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -22,12 +22,15 @@ default = []
 serde-json = ["askama_shared/serde-json"]
 with-iron = ["iron", "askama_derive/iron"]
 with-rocket = ["rocket", "askama_derive/rocket"]
+with-actix-web = ["actix-web", "askama_derive/actix-web", "mime_guess"]
 
 [dependencies]
 askama_derive = { path = "../askama_derive", version = "0.7.0" }
 askama_shared = { version = "0.7.0", path = "../askama_shared" }
 iron = { version = ">= 0.5, < 0.7", optional = true }
 rocket = { version = "0.3", optional = true }
+actix-web = { version = ">= 0.6", optional = true }
+mime_guess = { version = "2.0.0-alpha", optional = true }
 
 [package.metadata.docs.rs]
 features = [ "serde-json" ]
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -29,7 +29,7 @@ askama_derive = { path = "../askama_derive", version = "0.7.0" }
 askama_shared = { version = "0.7.0", path = "../askama_shared" }
 iron = { version = ">= 0.5, < 0.7", optional = true }
 rocket = { version = "0.3", optional = true }
-actix-web = { version = ">= 0.6", optional = true }
+actix-web = { version = "0.7", optional = true }
 mime_guess = { version = "2.0.0-alpha", optional = true }
 
 [package.metadata.docs.rs]
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -117,6 +117,15 @@ impl<'a> Generator<'a> {
         self.flush_ws(WS(false, false));
         self.writeln("Ok(())");
         self.writeln("}");
+
+        self.writeln("fn file_extension(&self) -> &str {"); 
+        let ext = match self.input.path.extension() {
+            Some(s) => s.to_str().unwrap(),
+            None => "txt",
+        };
+        self.writeln(&format!("{:?}", ext));
+        self.writeln("}");
+
         self.writeln("}");
     }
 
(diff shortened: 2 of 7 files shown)


askama_derive/src/generator.rs
```diff
@@ -234,6 +237,29 @@ impl<'a> Generator<'a> {
         self.writeln("}");
         self.writeln("}");
     }
+    
+    // Implement Actix-web's `Responder`.
+    fn impl_actix_web_responder(&mut self) {
+        //let lifetime = syn::Lifetime::new("'askama", Span::call_site());
```
ryanmcgrath: Bah, I forgot to remove these. They can be taken out,`actix_web::Responder` doesn't need 'em.

askama/Cargo.toml
```diff
@@ -22,12 +22,15 @@ default = []
...
 with-rocket = ["rocket", "askama_derive/rocket"]
+with-actix-web = ["actix-web", "askama_derive/actix-web", "mime_guess"]
 
 [dependencies]
 askama_derive = { path = "../askama_derive", version = "0.7.0" }
 askama_shared = { version = "0.7.0", path = "../askama_shared" }
 iron = { version = ">= 0.5, < 0.7", optional = true }
 rocket = { version = "0.3", optional = true }
+actix-web = { version = ">= 0.6", optional = true }
+mime_guess = { version = "2.0.0-alpha", optional = true }
```
djc: Can we depend on a stable version here, please?

testing/Cargo.toml
```diff
@@ -9,15 +9,17 @@ default = []
 nightly = ["rocket", "rocket_codegen", "askama/with-rocket"]
 
 [dependencies]
-askama = { path = "../askama", version = "*", features = ["with-iron", "serde-json"] }
+askama = { path = "../askama", version = "*", features = ["iron", "serde-json"] }
```
djc: I think this change is probably responsible for the test failures, the feature in `askama` is still called `with-iron` here, right?

testing/Cargo.toml
```diff
@@ -9,15 +9,17 @@ default = []
...
 iron = "0.6"
 rocket = { version = "0.3", optional = true }
 rocket_codegen = { version = "0.3", optional = true }
+actix-web = { version = "0.7", optional = true }
+bytes = { version = "0.4", optional = true }
 serde_json = "1.0"
 
 [build-dependencies]
-askama = { path = "../askama", version = "*", features = ["with-iron", "serde-json"] }
+askama = { path = "../askama", version = "*", features = ["iron", "serde-json"] }
```
djc: This also needs to be reverted, I think.

testing/Cargo.toml
djc: Also, I'd like to have the Cargo manifest changes for the `testing` crate in the earlier `test` commit. I can take care of that if you prefer, though.

ryanmcgrath:
Hmm, it'd probably be a bit before I could circle back to these (and to be honest, all the testing stuff confused me slightly here, so I'd honestly defer to you on it). Let me know if you need anything more from me though? Glad it's (mostly) there~

askama/Cargo.toml
ryanmcgrath: Ah, this `2.0.0-alpha`? That's...

- Unfortunately, what `actix-web` currently relies on
- And currently a choice of theirs to remain at this release until the `mime` crate stabilizes: https://github.com/abonander/mime_guess

I know, I found it weird too.

ryanmcgrath:
One other thing I added over in my fork that I found useful - a method on the `Template` trait that returns the file extension for the template. It sounds weird, but it comes in handy for getting around issues in frameworks where you want to return a variety of different things from a request handler, and need to turn it into an HTTP Response object - thus needing the extension to determine the mimetype.

djc:
I've merged all this stuff with a bunch of tweaks: c860bee5d5934fba6d57cbf815894a5cc8599ad9 for the actix-web `Responder` impl, ed98793b506c66723c713991f40c0065557a35d5 for the tests, and 8a25a1ee65031a687ad7d208eaa2843aed6c9752 to add an `extension()` method to `Template`.

Thanks again!

ryanmcgrath:
Ah, amazing! Thanks again!

ryanmcgrath:
Oh, quick question - when do you plan on throwing this up to crates.io?

djc:
I just released 0.7.1. 👍

Issue #105, opened 2018-07-27
Recursive includes end in endless loop


mashedcode:
Recursive includes don't work. Without those one can't build tree structures.
This is just a note. I'm currently not interested in implementing this.

djc:
Okay, thanks for the feedback! I'll take a look and see if there's an easy fix. Off the top of my head I don't see why it wouldn't work... Do you happen to have an error message for me?

mashedcode:
No error message. But it comes as a surprise to me that you think it should work as 
[handle_include](https://github.com/djc/askama/blob/d95c3c106191f2df091d1d53d6c2f6b00d5581d6/askama_derive/src/generator.rs#L534) generates inline code without generating a separate function.

djc:
Hmm, yeah, that makes sense.

djc:
This might be working after #114.

botika:
It does not solve this issue, but I think it is similar. In the test I will put several recursives included, that is, two more templates and [included.html](https://github.com/djc/askama/blob/master/testing/templates/included.html). I will keep you informed

botika:
@mashedcode The tests work correctly, give me the exact use case please.

mashedcode:
As already mentioned it's for tree data structures such as:
```
struct RecursiveInclude<'a> {
    data: &'a str,
    childs: Vec<RecursiveInclude<'a>>,
}
```

botika:
The same thing happens with macros. But I do not see the behavior in [pallet/jinja](https://github.com/pallets/jinja/blob/master/tests/test_regression.py). And for recursive macros or loops they use a different syntax. It is difficult to detect the recursion loop without this. Anyway, I think this behavior is of the macro or loop because they have their own scope.

josefrcm:
Hello everybody. I have been facing this problem, and I found a solution: call the render method from the template itself.

The struct definition would be something like this:
```
#[derive(Template)]
#[template(path = "recursive_template.html", escape = "none")]
struct RecursiveInclude<'a> {
    data: &'a str,
    children: &'a [RecursiveInclude<'a>],
}
```
And then the template:
```
<li><span>{{ data }}</span>
    <ul>
            {% for c in children %}
            {{ c.render().unwrap() }}
            {% endfor %}
    </ul>    
</li>
```

PR #114, opened 2018-09-07
Fix deep nested imports in macro calls

botika:
I think solve #105


djc:
Ohh, that looks like it was the right fix all along. Very nice, thanks!

djc:
Actually, I don't think it fixes #105? That is about includes, not macro calls.

PR #116, opened 2018-09-13
Fix operator preference at loop

botika:


djc:
Good stuff, thanks for fixing this!

PR #120, opened 2018-09-22
Add capitalize filter

botika:
#18


botika:
The title filter will go faster with a Regex. Can we depend on regex and lazy_static?

djc:
I'd prefer we make `regex` and `lazy_static` optional dependencies, and make the filters that need them compile conditionally on the feature. Does that make sense for you?

Do you think other filters will also require `regex`?

djc:
I'm not aware of a good solution in this space, I'm afraid. 😞

botika:
We can use cfg! macro and const fn. So I will implement it as soon as https://github.com/rust-lang/rust/issues/24111 is solved. In my opinion you can merge it without the regex.

Issue #121, opened 2018-09-25
[Question] Changing delimiters?


lukehsiao:
With the template delimieters as `{{` `}}` `{%`, etc., is there a way to use askama in content that heavily uses these curly braces and `%`? For example, if I wanted to use askama to template a LaTeX section like this:
 
 Trying to template `name.last`
 
 ```tex
 \newbibmacro{name:bold_author}{%
   \ifthenelse{\equal{\namepartfamily}{ {{name.last}} }}%
     {\textbf{\ifblank{\namepartgiven}{}{\namepartgiveni\space}\namepartfamily}}%
     {\ifblank{\namepartgiven}{}{\namepartgiveni\space}\namepartfamily}%
   \ifthenelse{\value{listcount}=1 \AND\value{liststop}=2}%
     {\space and\space}
     {\ifthenelse{\value{listcount}<\value{liststop}}%
       {\addcomma\space}
       {}
     }
 }
 ```
 
 I would assume that this would fail due to the conflicting use of `{}` and `%`.
 
 Using jinja's options in Python, I am able to switch delimiters:
 
 ```python
 block_start_string='~<',
 block_end_string='>~',
 variable_start_string='<<',
 variable_end_string='>>',
 comment_start_string='<#',
 comment_end_string='#>',
 ```
 
 which allows me to make a template that looks like the following and parses correctly:
 
 ```tex
 \newbibmacro{name:bold_author}{%
   \ifthenelse{\equal{\namepartfamily}{ <<name.last>> }%
     {\textbf{\ifblank{\namepartgiven}{}{\namepartgiveni\space}\namepartfamily}}%
     {\ifblank{\namepartgiven}{}{\namepartgiveni\space}\namepartfamily}%
   \ifthenelse{\value{listcount}=1 \AND\value{liststop}=2}%
     {\space and\space}
     {\ifthenelse{\value{listcount}<\value{liststop}}%
       {\addcomma\space}
       {}
     }
 }
 ```
 
 Is similar functionality possible in askama, or am I stuck? Is it impossible to use askama in templates that already use curly braces and percentage signs?

djc:
It is possible, but it has not been implemented. Since a few releases there is a per-crate configuration file that we could stick the configuration in, but wiring that up to the parser would need some work.

Would you be interested in working on that? If not, maybe @botika would like to help?

botika:
Nice to help! You can assign it to me now.

djc:
[...]

PR #122, opened 2018-09-27
Add changing delimiters support

botika:
#121

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -363,7 +363,7 @@ pub trait Template {
 
 pub use askama_derive::*;
 pub use shared::filters;
-pub use shared::{Error, MarkupDisplay, Result};
+pub use shared::{read_config_file, Error, MarkupDisplay, Result};
 
 #[cfg(feature = "with-iron")]
 pub mod iron {
@@ -441,7 +441,8 @@ fn visit_dirs(dir: &Path, cb: &Fn(&DirEntry)) -> io::Result<()> {
 /// that have templates, to make sure the crate gets rebuilt when template
 /// source code changes.
 pub fn rerun_if_templates_changed() {
-    for template_dir in &shared::Config::new().dirs {
+    let file = read_config_file();
+    for template_dir in &shared::Config::new(&file).dirs {
         visit_dirs(template_dir, &|e: &DirEntry| {
             println!("cargo:rerun-if-changed={}", e.path().to_str().unwrap());
         }).unwrap();
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1,6 +1,6 @@
 use super::{get_template_source, Context, Heritage};
 use input::TemplateInput;
-use parser::{self, Cond, Expr, MatchParameter, MatchVariant, Node, Target, When, WS};
+use parser::{Cond, Expr, MatchParameter, MatchVariant, Node, Target, When, WS};
 use shared::filters;
 
 use proc_macro2::Span;
@@ -12,6 +12,8 @@ use std::path::PathBuf;
 use std::{cmp, hash, str};
 
 use syn;
+
+use parser::parse;
 
 pub(crate) fn generate(
     input: &TemplateInput,
@@ -482,7 +484,7 @@ impl<'a> Generator<'a> {
             .config
             .find_template(path, Some(&self.input.path));
         let src = get_template_source(&path);
-        let nodes = parser::parse(&src);
+        let nodes = parse(&src, self.input.syntax);
         {
             // Since nodes must not outlive the Generator, we instantiate
             // a nested Generator here to handle the include's nodes.
--- a/askama_derive/src/input.rs
+++ b/askama_derive/src/input.rs
@@ -2,7 +2,7 @@ use proc_macro2::TokenStream;
 
 use quote::ToTokens;
 
-use shared::Config;
+use shared::{Config, Syntax};
 
 use std::path::PathBuf;
 
@@ -10,7 +10,8 @@ use syn;
 
 pub struct TemplateInput<'a> {
     pub ast: &'a syn::DeriveInput,
-    pub config: Config,
+    pub config: &'a Config<'a>,
+    pub syntax: &'a Syntax<'a>,
     pub source: Source,
     pub print: Print,
     pub escaping: EscapeMode,
@@ -24,7 +25,7 @@ impl<'a> TemplateInput<'a> {
     /// mostly recovers the data for the `TemplateInput` fields from the
     /// `template()` attribute list fields; it also finds the of the `_parent`
     /// field, if any.
-    pub fn new(ast: &'a syn::DeriveInput) -> TemplateInput<'a> {
+    pub fn new<'n>(ast: &'n syn::DeriveInput, config: &'n Config) -> TemplateInput<'n> {
         // Check that an attribute called `template()` exists and that it is
         // the proper type (list).
         let mut meta = None;
@@ -53,6 +54,7 @@ impl<'a> TemplateInput<'a> {
         let mut print = Print::None;
         let mut escaping = None;
         let mut ext = None;
+        let mut syntax = None;
         for nm_item in meta_list.nested {
             if let syn::NestedMeta::Meta(ref item) = nm_item {
                 if let syn::Meta::NameValue(ref pair) = item {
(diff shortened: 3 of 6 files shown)


djc:
Wow, that's fast! Again, thanks for working on this!

I initially suggested changes should be made in the configuration file. Instead, you've built it using attributes. I'm going to guess you did it this way because with my proposal it would be impossible to use different syntax in different test cases within the testing crate.

I think there are two ways to solve this issue:

- Only test different syntaxes at the parser level (using unit tests), instead of end-to-end tests. I think this is a fine way to solve it, though it provides less flexibility to end users.
- Use the configuration file to define named syntaxes and define a default syntax for the crate, while introducing an attribute for a different syntax on a per-template basis. This is a somewhat more involved change, but would be nicely flexible for end users.

Finally, I would like to request that you try to isolate refactoring changes to existing code (or, in fact, any change that's unrelated to the changing the delimiters) into separate commits. So ideally I'd like to have fixing the space in `Node::Call` in a separate commit, moving parts of the parser into an impl in a different commit, and so on. This will make it more explicit what you are refactoring to make this change and it can be a bit more work, but it makes reviewing much easier, increasing chances that we'll be able to find any issues that come up. Let me know if all this makes sense?

botika:
It's a deep refactoring in parser, but basically the functions, `expr_node, cond_block, block_id, match_else_block, when_block, block_match, block_for, block_block, block_macro, block_node, block_comment, parse_template, take_content and parser` to method of the new struct, Parser. 

If those changes have crept in, with the IDE code refactor, I'm sorry. I have fix it. 

It is prepared so that in the `extract` function, previously `TemplateInput::new`, it can pick up configuration from any side. I will implement the second option, but you will detail it with an example of the configuration, which I did not see clearly.

botika:
Now it can be configured by file and by derive with priority in the derive. The configuration file is given the properties `block_start, block_end, expr_start, expr_end, comment_start, comment_end` of 2 characters and the developer has to take care of the combinations of the different "start" braces because the method `take_content`.

botika:
With this last change, is it similar to the idea you had?

djc:
I still want there to be only named syntax (that is, no direct syntax manipulation via attribute fields -- only changing the syntax via an attribute field). I also still want the commit that modifies the parser macros to be split into two parts, one of which is a no-functional change refactoring.

botika:
Now you can create named syntax in `askama.toml`, and add his name to the `default_syntax` attribute without any modifier attribute (`block_start`,..). Modifiers by attribute in addition to this. Ok, now I divide the first commit in two,

botika:
Sorry

botika:
I can remove the modifier attributes but I would like you to explain where the problem is if it is only for test. You already had the cut, if you need anything else let me know.

djc:
So you mentioned having more freedom for the user. I understand how that might be attractive, but as a library maintainer, more freedom for the user means less freedom for me in how to evolve the API. As such, I'd rather be more conservating about exposing API to users, such that maintaining it will be simpler over time.

Additionally, I think Askama is already at the limit of how much attribute fields should be used. It's not great syntax (for example, there's less syntax highlighting at least in my editor) and having long attribute lines doesn't seem very idiomatic. So for this use case of custom syntaxes, which is arguably a bit niche, it seems more than good enough to only allow it to be done in configuration anyway. That has the added benefit to some extent of requiring users to name their syntax, which will make it more obvious what's going on if different syntaxes are used within a single crate.

As for your changes, the commit introducing `Parser` now changes in the `TemplateInput` type to support parser creation, but with this direction, that should likely be rolled back, with `syntax` just becoming a field on `TemplateInput`. 

Could `Parser::parse()` take a reference instead of an owned `self`? Cloning the `Parser` all over the places seems a little ugly.

Can syntaxes be named using `[[syntax.foo]]` rather than having an option `name` inside the syntax block? I think something like that is probably possible with TOML, but I haven't looked deeply into it. I'd prefer not to have a separate `default_syntax` on `Config`, it can just be named entry "default" in the `syntaxes` map. It also seems like we shouldn't need a `Syntax`, an `OptionSyntax` and a `RawSyntax`, although I can see how we might need two rather than three.

Again, thanks for doing this work, I know I'm a bit particular with how I want this to be done -- but as a maintainer, I'm going to be stuck with it for the foreseeable future. If you get tired of it, I can also finish it off myself (though it might take a bit more time). So thanks for working through this with me!

botika:
Ok, I remove the attributes of more, thanks for explaining it to me. 

Due to the lifetimes of the `TemplateInput` properties, it is not possible to copy or clone, since it lives in the heap and the stack at the same time. So it can not be used as an argument of `Parser::parse()`

I do not like clones either, but I have not found a way to call `named_args!` inside the macro `alt!`, `many0!`, etc and even if I did it I would have to clone or copy the `Syntax` struct. Clone could be avoided by a copy if I change the lifetimes, but now they came in the heap from [here](https://github.com/djc/askama/blob/master/askama_derive/src/input.rs#L59). And the main problem with the current implementation of the attributes is that they do not allow to set the lifetime in the stack.
And I was also thinking that we could pair other languages apart from Jinja, with few changes from this struct.

And as for `[[syntax.name]]`, it can be done, but it's a big change in `Config::from_str`, looking for these options by discard in the `Table`, `BTreeMap <String, Value>`. I add the [test with the feature](https://github.com/alexcrichton/toml-rs/blob/master/test-suite/tests/parser.rs#L122).

The three syntaxes were to be able to compose several Syntaxes in another using the overwrite method, if the feature is removed, it remains at 2.

So the main problem that I identify is the lifetime of `TemplateInput` and is a deep refactor. I can put it after these commits, I was already thinking how.

For what I have thought, redo the last commit by removing attributes and removing the 3 Syntax. And another solving the lifetime of `TemplateInput` and all those ugly clones. 
How about?

Thank you very much for your patience, I am learning a lot with Askama.

djc:
The `TemplateInput` should live long enough, though, to just pass it as a reference to everything else? Also ideally you don't pass the full `TemplateInput` to the parser, but instead pass (a reference to) the relevant `Syntax` instance from the `Config` keyed with the attribute field's syntax key (which should also live plenty long enough). Does that make sense?

If doing the other thing with TOML is hard, let's keep that for a follow-up PR to get this all straightened out first.

botika:
The methods for Parser seem good to you? Or try the static function? I do not find another way and I see it correct as long as it is not cloned.

You can use the same lifetime of the [`build_template` function](https://github.com/djc/askama/blob/master/askama_derive/src/lib.rs#L36) for everything else, including `TemplateInput`. In other words, nothing apart from the return of `build_template` must continue to exist. It can be done, but the refactor is deep, I do it?

djc:
Oh, I'd missed that you're passing `mut self` to the `Parser` methods. Is that why you're cloning it around? I think in this case you should be able to just use `&self`. If you do that, I don't see why you'd need deep refactoring. What I want is kind of like this:

- Restore `TemplateInput::new()`, `TemplateInput` will include a new `syntax: Option<&str>` field
- `build_template()` gets a `&Syntax` out of the `Config::syntaxes` map based on `input.syntax`
- `Parser::new()` takes a `&Syntax`
- `Parser` methods take `&self`
- `generator::generate()` also takes a `&Syntax`

Do you think that could work?

botika:
Ok I like it. 

I will have to dismantle the Config a little to do this, it would be `Config::new(& str)`, remove `Config::from_str` and the logic of the `fs: read_to_string` will draw it to `TemplateInput: new`. So the `config` disappears from `TemplateInput` in favor of `dirs: Vec<PathBuf>` and `Parser`. 

The only difference is to change in `TemplateInput`, `config` for `dirs` and let that config die. The idea is to eliminate  the config that we do not need and deserialize without alloc. How about?

djc:
I don't understand why you want to kill `Config` or why you'd have to. I do think `TemplateInput::new()` instantiating the `Config` is a little bit off, and it would be better to move that into `build_template()`, which would then pass `&Config` into `TemplateInput::new()`?

botika:
You can not put the `"&"` in the `self` field of the macro `method!`. The only way to use `call_m!` more than once in the same method is to use `mut self` https://docs.rs/nom/4.0.0/nom/methods/index.html. 
Now I'll do the refactor for` TemplateInput::new(ast, config)`.

botika:
TemplateInput can not borrow Parser since it is yours since it created it in yours constructor. 

Tomorrow I will continue with the few commits that are missing.

botika:
I do not know if well understood. 

The alternative to the `methods!` of Nom are the `named_args!`. I tested and can not pass arguments, I have looked at all the implementations with Nom that I have found and none use any of the 2. 

So I do not see how to pass the Syntax as an argument of `parser::parse()` and get it to `block_match` for example.

In the last refactor I have separated the logic of reading the configuration file of the Config contructor, because the string can only live above Config, if not it gives me the typical error of lifetime, "borrowed...". 

how about? 
I have looked at all the repositories that use Nom and none do this. If you have a small example of how to do it with the `named_args!`, I have run out of options in this topic.

botika:
It is already totally static.

djc:
Ohh, this is looking great! Using `named_args!()` for the parser seems like a very nice solution. I'm going to merge this now and tweak some things just a little bit.

PR #123, opened 2018-10-06
Syntax tweaks

djc:
@botika what do you think? This puts the syntax content validation in the `impl From<RawSyntax> for Syntax`, which seems like a nicer place for it.

Having played with your code a bit more, I'm even more impressed, it's all very well done!


botika:
It has been very good. 

Thank you very much, your words are the best salary I have ever paid. I am very happy with this experience. Thank you very much for creating Askama.

PR #124, opened 2018-10-06
Add changing delimiters documentation

botika:


askama/src/lib.rs
```diff
@@ -55,20 +55,53 @@
 //!   specified extension (`ext`): if the extension is `html`, `htm` or `xml`,
 //!   the `html` escape mode is used; otherwise, no implicit escaping is done.
 //!   Setting an escape mode explicitly overrides the inferred value.
+//! * `syntax` (as `syntax = "foo"`): set the syntax name for a parser defined
+//!   in the configuration file. The default syntax , "default",  is the one
+//!   provided by Jinja.
```
djc: I think you mean Askama here, not Jinja?

askama/src/lib.rs
```diff
@@ -55,20 +55,53 @@
...
 //!
 //! ## Configuration
 //!
 //! At compile time, Askama will read optional configuration values from
 //! `askama.toml` in the crate root (the directory where `Cargo.toml` can
-//! be found). Currently, this only covers the directories to search for
-//! templates.
+//! be found). Currently, this covers the directories to search for templates,
+//! as well as the personalization of syntax.
```
djc: Can we replace "personalization" with "customization" here? And please get rid of the extra empty line.

askama/src/lib.rs
```diff
@@ -55,20 +55,53 @@
...
+//!
 //!
 //! This example file demonstrates the default configuration:
 //!
 //! ```toml
 //! [general]
 //! # Directories to search for templates, relative to the crate root.
 //! dirs = ["templates"]
+//!
```
djc: Remove this empty line, please.

askama/src/lib.rs
```diff
@@ -55,20 +55,53 @@
...
+//! # Default syntax name
+//! default_syntax = "default"
+//!
```
djc: And this one...

askama/src/lib.rs
```diff
@@ -55,20 +55,53 @@
...
+//! ```
+//!
+//! Syntax consists in at least the attribute `name` which has to be unique for all syntax.
```
djc: Let's have only a single line of introduction before having the example, and then explain the constraints after the example -- I think that will be easier to understand.

It's "consists of", not "consists in". "which has to be unique for all syntax" => ", which uniquely names this syntax in the project."

askama/src/lib.rs
```diff
@@ -55,20 +55,53 @@
...
+//! Delimiters can be added to personalized syntax for example: `block_start`, `block_end`,
```
djc: Please expand this to a bullet list, introduced with "The following keys can currently be used to customize template syntax:".

askama/src/lib.rs
```diff
@@ -55,20 +55,53 @@
...
+//! `comment_start`, `comment_end`, `expr_start` and `expr_end`. Delimiters must have exactly 2
+//! characters and that ones ending in `*_start` must have one character in common in the same
```
djc: "must be exactly 2 characters long and start delimiters must all start with the same character."

askama/src/lib.rs
```diff
@@ -55,20 +55,53 @@
...
+//! position. If no delimiter added, default values are used.
```
djc: "If a key is omitted, the value from the default syntax is used."

askama/src/lib.rs
botika: must be exactly 2 characters long and start delimiters must all start **or end** with the same character.

djc: Why do we have that requirement, anyway? Is it that start delimiters must all start, and end delimiters must all end with the same character?

botika: It's because of the take_content function. Only in the start. Because it checks the input 1 to 1 and 2 correct ones are needed. Then the https://github.com/djc/askama/blob/master/askama_derive/src/parser.rs#L137-L150 can fail or give false positives or not parse them.

PR #126, opened 2018-10-09
Add map! filter

botika:
#18

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -385,6 +385,9 @@ use std::fs::{self, DirEntry};
 use std::io;
 use std::path::Path;
 
+#[macro_use]
+pub mod macros;
+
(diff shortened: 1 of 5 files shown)


djc:
I'm not sure I understand exactly what the use case or purpose for this feature is. Is this a need you encountered in some of the templates in your own project?

botika:
`Users on this page: {{ map!(users, username, lower) | join(', ') }}`. Applies a filter on a sequence of objects or looks up an attribute. I understand that there is no other alternative because in the `Vec<HashMap|BTreeMap>` it is not an alternative. There is polish to pass the second attribute or filter and can be done with filter with `|`.

djc:
I'm sorry, I still don't really understand what problem this is supposed to solve, so I'd rather not merge it for now.

PR #127, opened 2018-10-10
make `Template::extension()` static

ubnt-intrepid:
It is useful for calculating the media type without instantiating the context value.


ubnt-intrepid:
@djc Yes. I'll soon fix it.

PR #131, opened 2018-10-17
Add `indent` filter

casey:
Add an `indent` filter with signature `indent(s, width)` that indents `s` with `width` spaces. Not sure if `width` is the best name, `indentation_width` might be better.

### DIFF
--- a/askama_shared/src/filters/mod.rs
+++ b/askama_shared/src/filters/mod.rs
@@ -20,13 +20,14 @@ use escaping::{self, MarkupDisplay};
 // Askama or should refer to a local `filters` module. It should contain all the
 // filters shipped with Askama, even the optional ones (since optional inclusion
 // in the const vector based on features seems impossible right now).
-pub const BUILT_IN_FILTERS: [&str; 18] = [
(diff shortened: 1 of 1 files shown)


askama_shared/src/filters/mod.rs
```diff
@@ -138,6 +139,25 @@ pub fn truncate(s: &fmt::Display, len: &usize) -> Result<String> {
     }
 }
 
+/// Indent lines with `width` spaces
+pub fn indent(s: &fmt::Display, width: &usize) -> Result<String> {
+    let s = format!("{}", s);
```
casey: I used `format!` here, to match the existing functions, but I think that `s.to_string()` would be a little more readable.

djc: Do you want to change it for all the similar filter functions in one go?

casey: Yup, will do.

casey: Done

djc:
Thanks!

casey:
You are most welcome ^_^

PR #133, opened 2018-10-19
Take reference of unescaped expressions directly

djc:


TheWaWaR:
Maybe better add a test case for this.

https://github.com/TheWaWaR/askama/commit/cef25aa2525d6f1dd6d039ed1b7516f7717ef6a5

gregelenbaas:
Looks like this fix got merged in to master for version 0.8.0 but the latest on Crates.io is 0.7.2. When will the 0.8.0 branch be released to cargo? Thank you all for your work on this, askama is awesome!

PR #135, opened 2018-10-20
Fix reference of struct field ownership issue.

TheWaWaR:
fix #133


TheWaWaR:
Oops, 133 is a PR...

djc:
Thanks! I'm going to keep my original fix from #133, which seems a little simpler, but I've added a separate commit containing your test case, which seems just right, to the branch.

PR #138, opened 2018-10-24
Fix off-by-one error with HTML escaping

benjunmun:
Here's a small fix for a corner case I ran into with HTML escaping.

If the second-to-last character of a string should be escaped,
but not the last, the last character was not being included in the
result.


djc:
Very nice! I wonder if this might merit a RustSec advisory -- I've opened https://github.com/RustSec/advisory-db/issues/67 to track that.

PR #140, opened 2018-10-27
Improve bench with literals buffer for optimize ws writer

botika:


djc:
Very nice, thanks! In these performance PRs, would you mind pasting the Criterion output for the change you measured in the PR description? That would be very helpful.

PR #142, opened 2018-10-28
Improve bench with writable buffer

botika:
since https://github.com/djc/askama/commit/7062aabbcbb98bdeacda94071111f0fcc13a918d
```
 Running target/release/deps/all-7a1802b78ef1756b
Big table               time:   [894.00 us 895.23 us 896.52 us]                      
                        change: [-3.3812% -3.0583% -2.6742%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  2 (2.00%) high severe

Teams                   time:   [1.1794 us 1.1828 us 1.1864 us]                   
                        change: [-13.783% -13.247% -12.689%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe
```

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -9,7 +9,7 @@ use quote::ToTokens;
 
 use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
-use std::{cmp, hash, str};
+use std::{cmp, hash, mem, str};
 
 use syn;
 
@@ -41,8 +41,8 @@ struct Generator<'a> {
     skip_ws: bool,
     // If currently in a block, this will contain the name of a potential parent block
     super_block: Option<(&'a str, usize)>,
-    // buffer for literals
-    buf_lit: Vec<&'a str>,
+    // buffer for writable
+    buf_writable: Vec<Writable<'a>>,
 }
 
 impl<'a> Generator<'a> {
@@ -60,7 +60,7 @@ impl<'a> Generator<'a> {
             next_ws: None,
             skip_ws: false,
             super_block: None,
-            buf_lit: vec![],
+            buf_writable: vec![],
         }
     }
 
@@ -106,7 +106,7 @@ impl<'a> Generator<'a> {
             self.handle(ctx, &ctx.nodes, buf, AstLevel::Top);
         }
 
-        self.flush_ws(buf, WS(false, false));
+        self.flush_ws(WS(false, false));
         buf.writeln("Ok(())");
         buf.writeln("}");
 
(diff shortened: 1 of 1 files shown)


djc:
This is great, and thanks for the benchmark output! Left a few nits to review, please have a look.

askama_derive/src/generator.rs
```diff
@@ -576,35 +593,70 @@ impl<'a> Generator<'a> {
...
     }
 
-    // Write literals buffer and empty
-    fn write_buf_lit(&mut self, buf: &mut Buffer) {
-        if !self.buf_lit.is_empty() {
-            buf.writeln(&format!("writer.write_str({:#?})?;", self.buf_lit.join("")));
-            self.buf_lit = vec![];
+    // Write expression buffer and empty
+    fn write_buf_writable(&mut self, buf: &mut Buffer) {
+        if !self.buf_writable.is_empty() {
```
djc: Did you miss my comment about returning early here by inverting the `if` condition?

botika: I did not understand you. I flipped the if below. You want
```
if self.buf_writable.is_empty() {
   return
```
??
Can you put the code, please?

djc: Yes, that is exactly what I want. 👍

botika: Done

djc: I'll elide the `else` clause here.

djc:
BTW, if you're interested in suggestions for further work: 

* More performance work: consider looking into the performance of the `escape` filter, maybe writing a microbenchmark for it. It's possible that there is an external crate that provides the right functionality with better performance, or we might use something that can do SIMD to do smarter/faster searching.
* Other performance work: find some way to avoid the overhead of escaping for some types, in particular numbers. That would probably help performance on the big table benchmark. Could rely on the `num-traits` stuff you added for the other filter.
* Feature work: someone on Twitter asked about internationalization at some point. I would like to support that, but have not looked into it at all. Maybe investigate what support there is in other template engines in Rust (especially Tera, which I think is most featureful and quite similar) and what low-level infrastructure is available in Rust (consider both the traditional gettext-based stuff and maybe also the Fluent stuff that Mozilla has been working on).

djc:
(But of course, if you have other things you want to work on, that's great too!)

djc:
Merged, thanks!

botika:
Ok, cool, I'm watching it. On the other hand, in `self.handle_include` create a new `Generator`, the idea would be to pass it the `buf_writable` of the parent and return it again and remove the calls to `self.write_buf_literal`. Or some other way for parent and child to use the same `buf_writable` and not call `self.write_buf_writable` in `self.handle_include`. You could do it?

djc:
That also sounds like a good idea! I'll likely have little time to work on it myself for the short term.

djc:
(Also, includes are currently not used in the benchmarks, so before you go down that route you should probably add a benchmark that exercises that code.)

Issue #107, opened 2018-08-20
Cannot move out of borrowed content in for loop


iammichiel:
Hi there, 

I am probably not posting this in the right place but I seek help using this library. 
I am trying to iterate over a list of x509Name's using the following code in my template : 

```rust
{% for entry in certificate.subject_name().entries() %}

{% endfor %}
```

`certificate` being an instance of [X509](https://docs.rs/openssl/0.10.11/x86_64-unknown-linux-gnu/openssl/x509/struct.X509.html). I have the following error : `cannot move out of borrowed content`. I am pretty new to rust, this must be something easy but I have not yet full comprehension of the concept of ownership..

djc:
This is a fine place to post this!

So Askama's `Template` trait takes a reference to your context struct and in general converts everything in the template code to references in order to make things work. It's possible that that doesn't work in this case. As a first step, try enabling the `print = "code"` attribute on the `template()` call, so you can see the generated code, then paste that generated code into your source file and comment out the derive call and related `template()` attribute. That should pinpoint to the exact location of the error. If you then paste the generated code here, I can also take another look and see if the code generator can be improved in this respect.

iammichiel:
Thanks for your answer, Rust community is not a lie :)

I tried the `print` attribute as suggested and the following code was generated : 

```rust 
  match &self.certificate {
    Some(cert) => {
        for (_loop_index, entry) in (&cert.subject_name().entries()).into_iter().enumerate() {
            writer.write_str("\n\n                ")?;
        }
    }
    None => {
        writer.write_str("\n        ")?;
    }
}
```

Removing the leading `&` solves the problem but I have no idea how to do this in Jinja like code.

djc:
You can try doing it like this:

```
{% let entries = cert.subject_name().entries() %}
{% for entry in entries %}

{% endfor %}
```

But I'm not sure it'll work... In general, you cannot remove the `&` in the template code.

iammichiel:
Tried that as well, same error. `cannot move out of borrowed content`. My current work-around is creating a structure and extracting data in controller before rendering in the template.

djc:
I think that's probably the best workaround for now. :+1:

bhansconnect:
I am also running into this issue. My code is trying to create an iterator in the template:
```
{% for i in 1..n %}

{% endfor %}
```

lukehsiao:
Just wanted to follow-up on this issue with a similar problem. Given the following structs:

```rust
pub struct Coursework {
[...]

PR #143, opened 2018-10-30
Add not to borrow tag

botika:
#107

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -338,6 +338,20 @@
 //!
 //! Askama supports block comments delimited by `{#` and `#}`.
 //!
+//! ## Borrow
+//! Askama supports "not to borrow" tag by `^` before expression.
+//!
+//! ```text
+//! for i in ^iterator.clone().rev() {
+//!     {{ i }}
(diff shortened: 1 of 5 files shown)


djc:
I'm afraid I don't like this solution very much. If you want to make progress on #107, I think we need some analysis on exactly what the current generated code can handle and where it fails, and what avenues are available to make the latter cases more ergonomic. Adding an unfamiliar sigil to magically fix this problem seems like a hard-to-learn and hard-to-understand solution to the problem.

botika:
I have tried to explain it in the documentation a little better, it is not much better :sweat_smile: . It is for cases in which as in the return of the `clone()` that returns the element with the borrow. In other words, you have to launch a function that returns `&` needs it. Suggestions for documentation are accepted. Without tag I do not know how we can know when the function returns a borrow. What do you think?

djc:
Still don't like it, sorry. 😞 Maybe we should just document the constraints here rather than adding the new operator, with a workaround.

botika:
If much better document and give alternatives to these. Ok, I close

PR #144, opened 2018-10-31
Add bench for escaping and improve it

botika:


testing/benches/all.rs
```diff
@@ -10,8 +10,9 @@ criterion_main!(benches);
 criterion_group!(benches, functions);
 
 fn functions(c: &mut Criterion) {
-    c.bench_function("Big table", |b| big_table(b, &100));
-    c.bench_function("Teams", |b| teams(b));
+    c.bench_function("Big table", |b| big_table(b, &100))
```
djc: Why did you use a chaining structure here rather than just adding a bench function with a separate function call?

testing/benches/all.rs
```diff
@@ -69,3 +70,17 @@ struct Team {
     name: String,
     score: u8,
 }
+
+fn escaping(b: &mut criterion::Bencher) {
+    let string_foo = "foo".to_string();
+    let string_full_escape = "<>&\"\\/";
+    let string_mix_escape = "<foo>bar&foo\"bar\\foo/bar";
```
djc: In terms of benchmark cases, I think we roughly need the following to adequately cover the usage surface of this function:

- Short string vs longer string (like some chunk of "lorem ipsum")
- No escapable characters vs ~5% escapable characters

The (almost) everything case (as in `string_full_escape` here) seems not so important to the performance profile of this function, in my view.

How does that sound?

askama_shared/src/escaping.rs
```diff
@@ -55,26 +55,33 @@ fn escapable(b: u8) -> bool {
 }
 
 pub fn escape(s: String) -> String {
-    let mut found = Vec::new();
-    for (i, b) in s.as_bytes().iter().enumerate() {
-        if escapable(*b) {
-            found.push(i);
+    let found: usize = if let Some(c) = s.chars().next() {
```
djc: What is the purpose of separately testing the first character, here?

askama_shared/src/escaping.rs
```diff
@@ -55,26 +55,33 @@ fn escapable(b: u8) -> bool {
...
+        if escapable(c as u8) {
```
djc: This seems like it could potentially be wrong, but I don't know enough about the details of how Unicode and the Rust `char` type work to be sure. Can you dig into that and convince me? :smile:

askama_shared/src/escaping.rs
```diff
@@ -55,26 +55,33 @@ fn escapable(b: u8) -> bool {
...
-    let max_len = bytes.len() + found.len() * 5;
-    let mut res = Vec::<u8>::with_capacity(max_len);
-    let mut start = 0;
-    for idx in &found {
-        if start < *idx {
-            res.extend(&bytes[start..*idx]);
-        }
-        start = *idx + 1;
-        match bytes[*idx] {
+    let mut res = Vec::with_capacity(s.len() + 5);
```
djc: I'm wondering about the justification about the simplification of the allocation size here. I think the previous version was more accurate (and should thus require fewer extra allocations -- but maybe you won't hit that with the short strings you use for benchmarking now).

askama_shared/src/escaping.rs
```diff
@@ -55,26 +55,33 @@ fn escapable(b: u8) -> bool {
...
+    res.extend(&bytes[0..found]);
+    for c in bytes[found..].iter() {
```
djc: I'm also skeptical of this transformation for longer string sizes. Let's land a more comprehensive set of benchmark cases before you try to optimize, ok?

testing/benches/all.rs
botika: Sound goods

askama_shared/src/escaping.rs
botika: Yes, I can change it for a boolean.

askama_shared/src/escaping.rs
botika: With the change from above it is solved.

askama_shared/src/escaping.rs
botika: The rest I look at it with a longer text as you indicate above. Now I upload the changes.

botika:
It has an alloc less than the original, since the found also had to allocate and is assumed at least one. And now make `found.len()` - 1 less comparisons. Append with `extends` it's the same as with `push`. I've already made the changes, I had no idea about the chars :sweat_smile: , thank you. How does that sound?

askama_shared/src/escaping.rs
```diff
@@ -55,26 +55,25 @@ fn escapable(b: u8) -> bool {
 }
 
 pub fn escape(s: String) -> String {
-    let mut found = Vec::new();
+    let mut found = 0;
+    let mut flag = true;
```
djc: I think we can make `found` an `Option<usize>` here, instead of having the separate `flag` variable?

botika: It's so cool, :smile: thanks 
```
  Running target/release/deps/all-7a1802b78ef1756b
Big table               time:   [846.27 us 847.09 us 848.03 us]
                        change: [-1.3469% -0.7382% -0.1178%] (p = 0.02 < 0.05)
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) high mild
  9 (9.00%) high severe

Teams                   time:   [1.1615 us 1.1647 us 1.1685 us]
                        change: [-1.5642% -0.7306% +0.0928%] (p = 0.09 > 0.05)
Found 18 outliers among 100 measurements (18.00%)
  1 (1.00%) high mild
  17 (17.00%) high severe

Escaping                time:   [6.1220 us 6.1350 us 6.1507 us]
                        change: [-0.2161% +0.5854% +1.4861%] (p = 0.18 > 0.05)
Found 19 outliers among 100 measurements (19.00%)
  4 (4.00%) high mild
  15 (15.00%) high severe
```
Now I do the commit. thanks

botika:
Done. Very good work.

askama_shared/src/escaping.rs
botika: The allocations are one less than the previous version, because the allocations in "found" vector also count.

botika:
Well this is already? There is something else you want to emphasize? 
I want to make it public and be part of the crate `askama_shared`, since all the other alternatives are slower.

botika:
Thanks to @h2o, I do not think it can be done faster in Rust. I would like to expose this function to use it in other crates.

djc:
What do you mean by "thanks to @h2o"? I'm pretty sure there are still a number of opportunities to optimize this code (for example, by using SIMD).

botika:
Yes, at least until https://github.com/rust-lang/rust/issues/27731 is resolved.
 
On the other hand, when I need a `new Vec` I can not adjust the time of life.
Measuring the resulting Vec worsens the performance by having to make 2 passes on the string.

askama_shared/src/escaping.rs
```diff
@@ -47,60 +47,53 @@ where
...
-        match bytes[*idx] {
-            b'<' => {
-                res.extend(b"&lt;");
-            }
-            b'>' => {
-                res.extend(b"&gt;");
-            }
-            b'&' => {
-                res.extend(b"&amp;");
+    if let Some(found) = found {
```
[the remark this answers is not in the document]
botika: Ok, this
```
match found {
   None => s,
   ...
```

botika:
Sizing vector case: https://gist.github.com/botika/f1a940757a126f114cbf5357bbb596a0

djc:
So I had some more ideas last night and came up with this commit: https://github.com/djc/askama/commit/2ffab977f7cb7ecd98cde199e0b870e363319def. Can you have a look? This made the core escaping code just 3% faster for me than your latest (in [this variant](https://github.com/djc/askama/commit/3abe758975492e4b986b3e5c5028e06408796c65)), but is a nice win on the actual templating benchmarks (10% on big table and 22% on teams) -- and it's nicely compact code, I think.

askama_shared/src/escaping.rs
```diff
@@ -47,60 +47,57 @@ where
...
-                res.extend(b"&#x27;");
-            }
-            b'/' => {
-                res.extend(b"&#x2f;");
+    match found {
+        None => s,
+        Some(found) => {
+            let bytes = s.as_bytes();
+            // Allocate s + (6 * 10% of s)
+            let mut res = Vec::with_capacity(s.len() + 6  * s.len() / 10);
```
djc: Actually hardcoding this 10% seems a little too magic, does that make a big difference?

botika: 2-3% in big table and 5-7% escaping bench, We can can look for another way to adjust the size of the vector. Any more complex heuristics?

botika:
https://github.com/djc/askama/pull/144/commits/19a51181cc807527c8e79b43960b8ffe0427ec6a
```
 Running target/release/deps/all-7a1802b78ef1756b
Benchmarking Big table: Collecting 100 samples in estimated 5.8254 s (5050 iterationBig table               time:   [1.1450 ms 1.1572 ms 1.1706 ms]
                        change: [+36.783% +38.188% +39.964%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  6 (6.00%) high mild
  1 (1.00%) high severe

Teams                   time:   [1.1662 us 1.1686 us 1.1713 us]
                        change: [+0.7566% +1.4850% +2.1822%] (p = 0.00 < 0.05)
Found 17 outliers among 100 measurements (17.00%)
  7 (7.00%) high mild
  10 (10.00%) high severe

Escaping                time:   [5.1435 us 5.1575 us 5.1737 us]
                        change: [-3.9436% -3.1941% -2.3427%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) high mild
  5 (5.00%) high severe
```

askama_shared/src/escaping.rs
djc: Okay, so let's just do `s.len() + s.len() / 10`, and maybe add a comment saying that this is a heuristic with a conservative estimate to save on intermediate allocations.

botika: We could do more cases in the bench, especially extreme to find the correct heuristic approach. How about?

djc: Let's skip that for now. This PR is already complex enough while we get the algorithm sorted out, and adding more cases to the benchmark will likely just make us micro-optimize for the specific benchmarks we've selected.

askama_shared/src/escaping.rs
```diff
@@ -59,9 +61,13 @@ macro_rules! escaping_body {
 }
 
 const FLAG: u8 = b'>' - b'"';
-pub fn escape(s: String) -> String {
+pub fn escape<'a, S>(s: S) -> Cow<'a, str>
+where
+    S: Into<Cow<'a, str>>,
+{
+    let s = s.into();
     let mut found = None;
```
botika: Ref

botika:
I do not see how to optimize it, we make a new pull request from https://github.com/djc/askama/tree/escaping

PR #145, opened 2018-11-01
Update generated code example

botika:


djc:
So I wanted to hold off on this to improve the generated code's formatting a little bit, adhering as much as possible (without too much complexity) to `cargo fmt` style. If you want to work on that, that would be great, otherwise I'll get to it myself sometime soonish.

botika:
Ok, no problem, I use this pull request.

PR #146, opened 2018-11-03
Add int and float filters

botika:


djc:
So do you have an actual use case for these, and if so, can you explain it to me a little?

botika:
We have done it because they are listed in #18, but a practical case would be to truncate a float or cast int for operation with float. Per example:
```
{% for i in range %}
   {{ loop.index / (var_f32 | int) }}
   {{ (loop.index | float) / var_f32 }}
{% endfor %}
```

askama_shared/src/filters/mod.rs
```diff
@@ -158,6 +161,14 @@ pub fn indent(s: &fmt::Display, width: &usize) -> Result<String> {
     Ok(indented)
 }
 
+/// Casts number to i32
+pub fn int<T>(number: T) -> Result<i32>
```
djc: So this is called `int`, but it casts to `i32` -- which is kind of a random match. It could just as well be `isize`, `usize`, or `u64`.

askama_shared/src/filters/mod.rs
```diff
@@ -77,6 +78,14 @@ where
     escape(i)
 }
 
+/// Casts number to f32
+pub fn float<T>(number: T) -> Result<f32>
```
djc: Same issue here -- why `f32`, not `f64`?

askama_shared/src/filters/mod.rs
botika: We adjust it to isize.

askama_shared/src/filters/mod.rs
botika: We adjust it to f64
.

djc:
Hmm, so how about we call them `into_isize` and `into_f64` then? I don't like those particular names that much, but I also think using the more generic `int` and `float` names for more-specific types feels wrong.

botika:
With these more specific types we do some more, apart of `into_isize` and `into_f64`? `into_i64` or `into_usize`...?

djc:
If you have an actual use case for these today, we could do that. If you're satisfied with the first two, I'd prefer to keep it limited and further extend the collection as people come up with use cases.

botika:
Done

PR #148, opened 2018-11-04
use UFCS in derived code

### LINKED ISSUE #147: derivation does not work without importing `Template`
From Rust 1.30, it is possible to specify the path directly in `#[derive()]`. Unfortunately, this notation does not work as the derived code assumes that `Template` is imported at the current module: 

```rust
extern crate askama;

#[derive(askama::Template)]
#[template(source = "foo", ext = "txt")]
struct Template {}
```

```
   Compiling askama1 v0.1.0 (/home/ubnt-intrepid/askama1)
error[E0599]: no method named `render_into` found for type `&Template` in the current scope
 --> src/lib.rs:3:10
  |
3 | #[derive(askama::Template)]
  |          ^^^^^^^^^^^^^^^^
  |
  = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope, perhaps add a `use` for it:
  |
4 | use askama::Template;
  |

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: Could not compile `askama1`.

To learn more, run the command again with --verbose.
```

ubnt-intrepid:
fixes #147

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: rust
 rust:
+  - 1.30.0 # minimum Rust version supported by Askama.
   - stable
   - beta
 cache: cargo
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -137,7 +137,7 @@ impl<'a> Generator<'a> {
     fn impl_display(&mut self, buf: &mut Buffer) {
         self.write_header(buf, "::std::fmt::Display", None);
         buf.writeln("fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {");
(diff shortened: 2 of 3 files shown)


ubnt-intrepid:
Removed the custom compilation flag from `build.rs` (it is true that the solution by build script is not smart...)

ubnt-intrepid:
The requirements of Rust 1.30 is only used in `testing` and the crate users can continue to use the older compilers which supports custom derive (I noticed it after pushing changes).
In either case, I think it is better to clarify the requirements of Rust version supported by the crate in the next version.

djc:
Thanks! I've cherry-picked your change as 82a406df795e2645c83d9ec0974f1f0421337d9b. I'll keep "stable" as the supported minimal Rust version in the Travis build matrix for now and just try to mention a minimal version bump in release change logs when applicable.

PR #150, opened 2018-11-05
Simplify escape 

botika:
```
Running target/release/deps/all-a5a3ff4ba11f751e
Big table               time:   [730.60 us 733.71 us 737.34 us]
                        change: [-5.0011% -4.3003% -3.5780%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  3 (3.00%) high mild
  15 (15.00%) high severe

Teams                   time:   [857.94 ns 866.50 ns 876.90 ns]
                        change: [-9.4725% -7.9203% -6.0633%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  10 (10.00%) high severe

Escaping                time:   [5.1099 us 5.1387 us 5.1745 us]
                        change: [-24.333% -23.266% -22.257%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  5 (5.00%) high mild
  5 (5.00%) high severe
```

### DIFF
--- a/askama_shared/src/escaping.rs
+++ b/askama_shared/src/escaping.rs
@@ -55,46 +55,34 @@ pub struct Escaped<'a> {
     bytes: &'a [u8],
 }
 
-enum State {
-    Empty,
-    Unescaped(usize),
+macro_rules! escaping_body {
+    ($state:ident, $i:ident, $fmt:ident, $_self:ident, $quote:expr) => {{
+        if $state < $i {
+            $fmt.write_str(unsafe { str::from_utf8_unchecked(&$_self.bytes[$state..$i]) })?;
+        }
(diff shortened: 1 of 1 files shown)


botika:
in favor of #144

djc:
I've rebased this PR with some minor style tweaks. Thanks for sticking with it!

djc:
(Updating the template-benchmarks-rs repo from 940f90be to 2e028ed9 shows a ~20% improvement, so this is pretty nice. Still a ways to go before we catch up with horrorshow, though I suspect it might not be doing escaping at all.)

PR #151, opened 2018-11-06
Fix generated code and update example

tizgafa:
#145

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -627,7 +627,7 @@ impl<'a> Generator<'a> {
                 buf.writeln(&format!("writer.write_str({:#?})?;", &buf_lit.buf));
             } else {
                 let mut buf_format = Buffer::new(0);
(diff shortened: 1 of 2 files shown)


askama_derive/src/generator.rs
```diff
@@ -649,16 +649,18 @@ impl<'a> Generator<'a> {
...
                 }
 
-                buf.write("write!(writer, ");
-                buf.write(&format!("{:#?}", &buf_format.buf));
-                buf.writeln(", ");
-                buf.write("\t");
+                buf.writeln("write!(");
+                buf.indent();
+                buf.writeln("writer,");
+                buf.write(&format!("{:#?},\n", &buf_format.buf));
```
djc: I think we can use `writeln()` here instead of including `\n` in the format string?

botika: It is so that `buf.start` is `false` in the following instruction.

botika:
Done

PR #153, opened 2018-11-11
Add heuristic for gauge capacity of output String

botika:
```
 Running target/release/deps/all-1cd74d22a4e3eac1
Big table               time:   [725.54 us 726.58 us 727.79 us]
                        change: [-1.5091% -0.8277% -0.1225%] (p = 0.02 < 0.05)
Found 15 outliers among 100 measurements (15.00%)
  3 (3.00%) high mild
  12 (12.00%) high severe

Teams                   time:   [740.75 ns 741.68 ns 742.96 ns]
                        change: [-15.179% -14.519% -13.787%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  3 (3.00%) high mild
  13 (13.00%) high severe
```

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -615,71 +615,68 @@ impl<'a> Generator<'a> {
     fn write_buf_writable(&mut self, buf: &mut Buffer) {
         if self.buf_writable.is_empty() {
             return;
-        } else {
-            if self.buf_writable.iter().all(|w| match w {
-                Writable::Lit(_) => true,
-                _ => false,
-            }) {
-                let mut buf_lit = Buffer::new(0);
-                for s in mem::replace(&mut self.buf_writable, vec![]) {
-                    if let Writable::Lit(s) = s {
-                        buf_lit.write(s);
+        }
+
+        if self.buf_writable.iter().all(|w| match w {
+            Writable::Lit(_) => true,
+            _ => false,
+        }) {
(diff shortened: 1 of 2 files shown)


botika:
Ok, I had already thought, all my pull requests are follow-up :smile: 
I have tried to return a `usize` of `handler` (for the `child`) and `write_buf_writable`, and a property of `Generator` capacity. Do you see any way to generate including without a child?

On the other hand, adjust these values in the configuration file or/and an approximate operator for the expression.

We continue discussing. While I'm approaching `write!` with the arguments v2 and do the `if-else` so that in the cases that it can return and append it to `buf_writable`. 

In order that the code generated is the same as the bench of std_write.

djc:
Yeah -- I'm afraid I'm a bit of an opinionated perfectionist, thanks for putting up with me!

Returning `usize` from `handle()` sounds right.

Generating `include` without a child is a bit tricky. What we could do is doing it similar to how blocks are handled for inheritance: visit the included templates up front before the code generation stage, in `find_used_templates()`. That would require `find_used_templates()` to visit the full templates though rather than just the top-level... but maybe that'll just be the (compile-time) cost of doing business. That way the included template AST nodes will have the same lifetime as the parent's, making the code generation not require a child `Generator` instance.

If you can get it to generate the same code as the `std::write` benchmark, that would be pretty awesome!

PR #154, opened 2018-11-11
Write macro named arguments

botika:


djc:
What's the perf win here?

botika:
```
 Running target/release/deps/all-1cd74d22a4e3eac1
Big table               time:   [709.48 us 710.21 us 711.05 us]
                        change: [-1.2486% -0.5887% +0.0479%] (p = 0.08 > 0.05)
  5 (5.00%) high mild
  8 (8.00%) high severe

Teams                   time:   [847.99 ns 849.00 ns 850.18 ns]
                        change: [-2.1367% -1.2890% -0.5289%] (p = 0.00 < 0.05)
Found 14 outliers among 100 measurements (14.00%)
  5 (5.00%) high mild
  9 (9.00%) high severe
```

botika:
It is minimal because this case is very little reflected in the bench.

askama_derive/src/generator.rs
```diff
@@ -638,18 +642,30 @@ impl<'a> Generator<'a> {
...
-
-                            buf_format.write("{}");
-                            buf_expr.write("&");
-                            buf_expr.write(&match (wrapped, &self.input.escaping) {
-                                (Wrapped, &Html) | (Wrapped, &None) | (Unwrapped, &None) => {
-                                    expr_buf.buf
-                                }
-                                (Unwrapped, &Html) => {
-                                    format!("::askama::MarkupDisplay::from(&{})", expr_buf.buf)
+                            if cache_id.contains_key(&expr_buf) {
```
djc: Let's match against `cache_id.entry()` here, which should be more efficient (single cache lookup if it is present).

askama_derive/src/generator.rs
```diff
@@ -628,6 +631,7 @@ impl<'a> Generator<'a> {
             } else {
                 let mut buf_format = Buffer::new(0);
                 let mut buf_expr = Buffer::new(buf.indent + 1);
+                let mut cache_id = HashMap::with_capacity(self.buf_writable.len());
```
djc: Let's call this `expr_cache`?

askama_derive/src/generator.rs
```diff
@@ -638,18 +642,30 @@ impl<'a> Generator<'a> {
...
+                                            expr_buf.buf
+                                        ));
+                                    }
                                 }
-                            });
-                            buf_expr.writeln(",");
+
+                                buf_expr.writeln(",");
+                                buf_format.write(&format!("{{a{}}}", id));
+                                cache_id.insert(expr_buf.clone(), id);
```
djc: Move the `insert()` and `named += 1` up to right after the line `let id = self.named` so the logic is better localized?

Actually, we can probably move ownership of `expr_buf` into the cache map? In that case, it needs to be down at the bottom, but we should get rid of the `clone()` (and the corresponding `derive()`).

askama_derive/src/generator.rs
botika: Ok

askama_derive/src/generator.rs
botika: Ok

askama_derive/src/generator.rs
botika: and ok

PR #155, opened 2018-11-11
Simplify expr writing

djc:
Some tweaks to make the code hopefully easier to read/follow.


djc:
@botika you'll need to re-approve since there have been new commits after your previous approval.

PR #156, opened 2018-11-11
Fix write expression wrapper

botika:


djc:
Can you please write a test that would have caught this?

botika:
Ok,

djc:
Thank you!

PR #157, opened 2018-11-13
Add assignments documentation

botika:


botika:
No idea, we should repair them and upload a web page to netifly like what Tera does, I can order it, but English is not my specialism.

djc:
I was not talking to about how we deploy them? By deferred assignment, I mean what's tested in the `LetDeclTemplate`.

PR #158, opened 2018-11-13
Update generated code example

botika:


djc:
This reminds me that we should fix that code style regression, can you take care of that?

botika:
Exactly, you mean?

djc:
I mean that the closing `)` from the `write!(` call should go on the next line.

djc:
Oh hey, you have a name now, pleased to meet you. 👍 😁

PR #159, opened 2018-11-14
Move escaping benchmarks into askama_escape crate

djc:


djc:
Thanks for the quick review!

PR #160, opened 2018-11-16
Add avx escape

botika:
Needs the structure inside the crate. 

Bench:
```
    Running /home/lorenz/workspace/CSFM/askama/target/release/deps/all-9ebea23b2b2e40d8
Escaping                time:   [4.8502 us 4.8627 us 4.8762 us]
                        change: [-1.1820% -0.2382% +0.6101%] (p = 0.62 > 0.05)
  9 (9.00%) high mild
  4 (4.00%) high severe

Avx Escaping            time:   [1.2224 us 1.2264 us 1.2307 us]
                        change: [-2.8507% -2.3043% -1.7410%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  5 (5.00%) high mild
  1 (1.00%) high severe
```


botika:
Missing ifunc https://github.com/BurntSushi/rust-memchr/blob/master/src/x86/mod.rs#L30-L56

askama_escape/src/lib.rs
```diff
@@ -50,45 +53,212 @@ pub fn escape(s: &str) -> Escaped {
...
+        }
+    }
+}
+
+#[inline(always)]
+fn _escape(bytes: &[u8], fmt: &mut fmt::Formatter) -> fmt::Result {
+    let mut start = 0;
+
+    for (i, b) in bytes.iter().enumerate() {
+        unsafe {
```
djc: Why do we have the large `unsafe` block here, instead of the smaller one before?

botika: Ok, or directly `unsafe fn`?

djc: I don't understand your question. My concern is that the new `unsafe {}` block covers a lot of code, so it's not immediately obvious which parts are actually unsafe.

botika: is by `str::from_unchecked_utf-8` in `write_char!
`

botika: `write_char` is used in `_avx_escape` which is `unsafe` and returns the double unsafe if I just decorate this instruction `str::from_unch..`
.

botika: I have marked all the unsafe blocks of the macros with comments. This block picks up a lot so I do not repeat myself.

botika:
of the unsafe topic, can you send a pull request to my master? 

Apart from the unsafe, it's finished. 
Subsequently, improvements in the alignment logic can be proposed.

askama_escape/src/lib.rs
botika: of the unsafe topic, can you send a pull request to my master?

botika:
Wait alignment solution, please comment unsafe solution.

djc:
So I was thinking about this last night and I think we should separate out the long and short benchmarks so we can compare the effect of using AVX2 instructions on them separately (based on my reading, I think AVX2 stuff might actually slow down short string escaping). Can you do that? I'll also want to chew a bit on this code since I myself don't have any experience with AVX2 and I want to really understand it before we merge this.

Also I think next we might want to explore searching based on the PCMPESTRI instruction from SSE 4.2, which might be faster than AVX2 according to what's described [here](https://github.com/BurntSushi/rust-memchr/issues/38).

botika:
Ok, I do the separation in the bench. 
[picohttpparser](https://github.com/h2o/picohttpparser/blob/master/picohttpparser.c) is implemented with that I look at it.

botika:
it is a travis joke?

botika:
I'm going crazy with [these](https://github.com/rust-lang/rust/pull/49664) attributes, can you see if they're fine?

botika:
This fact. Any comment?

djc:
So, high level conclusions just from looking at your benchmark data:

- The SSE implementation doesn't really pay for itself in terms of complexity / performance
- AVX2 is faster even for the shortest strings you've tested (5 characters)
- AVX2 is only neglibly faster if there's not a lot to escape

Very good stuff! In light of these findings, I'd like to change the benchmarks a little:

- Lengths of 1 bytes, 10 bytes, and 5 MB (to have more granularity about smaller strings)
- At 10 bytes and 5 MB, add strings that have 3% escaped characters

My question here is if the AVX2 stuff is even worth having, or if it's really only effective at levels of escaped characters that are unlikely to come up in the wild.

(I'm really sorry if we end up not using some of this painstakingly developed code, but I also want to be realistic about the complexity it adds and whether it's worth it in terms of real-world performance.)

djc:
Also, can you squash the code prettification from the last commit into the related earlier commits?

botika:
But if it is exactly the same to the right and the left, I think AppVeyor is not working well.

botika:
There is another alternative to improve e2e, https://github.com/rust-lang/rust/issues/31844. My experience is very short, but I could embolden myself with littles advice.

botika:
I do not know what happens to AppVeyor, I've updated the bench

botika:
We are comparing the efficiency between a truck and a motorcycle to carry a candy. 

It does not make any sense and it is also that this result warned by gitter when you asked me. Now you can not tell that all the work does not work because what I said has happened.

djc:
Yes, you were right. Just to make sure I understand, are you angry and/or frustrated with me for this?

botika:
I'm not angry nor frustrated, but is very unefficient that after talking and knowing about that possible issue you comment on it like a problem or something unexpected.

botika:
Wtf, Bill, what do we do with that memory? I have icons but not in my memory, perfect.

djc:
> Wtf, Bill, what do we do with that memory? I have icons but not in my memory, perfect.

Huh? Not sure what you mean.

botika:
In Windows, AppVeyor, it gives an equality error in two strings with the same bytes. But I still have 4k icons and these useful things. I was messing with Windows. I removed the simd support to windows.

djc:
> I'm not angry nor frustrated, but is very unefficient that after talking and knowing about that possible issue you comment on it like a problem or something unexpected.

It's not a problem per se, but I still didn't know what to expect. That might be because I don't have a lot of experience doing actual performance optimization, even though I know quite a bit about what kinds of things can cause performance issues. So I had expected that AVX2 would be substantially faster at large strings even if there wasn't a lot to escape. Now that it isn't, I'm no longer sure if there's value in it.

What do you think we should do? Do you think adding the SSE and AVX versions of the code will lead to better performance on real-world workloads? For that matter, have you tried and compared this branch on an actual project you've used Askama for?

My inclination is to see if I can figure out a better benchmark, in the sense that it should be a more reflective mix of longer and shorter strings as well as the relative frequencies of characters from the different relative spaces. For example, it occurs to me now with the benchmark results we have that the trick of ignoring byte values from outside the range of escapable characters is very effective, but that means that digits and non-escaped interpunction will have a different effect on the benchmark results -- yet, the current benchmark data does not contain this class of character at all.

botika:
When I do not like sse, but the avx if I like it. It is still very green and does not give the expected performance. 

I will continue investigating vectorized operations with @tizgafa and later we will try again on askama. 

Viewing the results, prioritize askama in the recognition of elements not escapeable and the heuristic of output size. 

How about?

PR #166, opened 2018-12-02
 Add specialization and optimize

botika:
RFC
at nightly
```
Escaping                time:   [1.6821 us 1.6825 us 1.6828 us]
                        change: [-65.188% -65.075% -64.985%] (p = 0.00 < 0.05)
Big table               time:   [283.14 us 283.19 us 283.24 us]
                        change: [-56.530% -56.514% -56.497%] (p = 0.00 < 0.05)
Teams                   time:   [463.01 ns 463.41 ns 463.89 ns]
                        change: [-47.048% -46.984% -46.887%] (p = 0.00 < 0.05)
```


botika:
How about changing the filter `safe` from function to reading in compilation? In order to use the `MarkupDisplay` only when necessary. With it we will avoid the `match` every `fmt`.

botika:
#167

Issue #136, opened 2018-10-21
User-defined escaping


lukehsiao:
**Feature Request**:
Would it be possible to allow users the ability to define a custom escape mapping in the `askama.toml`, similar to how a user can define a custom syntax? 

I'm not entirely sure what this would look like so this issue is just a place to discuss the idea.

lukehsiao:
I believe it should be more general that just triggering on the same characters that are currently escaped in html. 

> On the other hand, the current architecture/interface is sort of built on the notion that different file types need different escaping modes, so that seems the most clean way to plug in. 

This is the idea I think is important. Almost all the templating engines (especially for Rust) mainly look at HTML templates. It would be great to be able to use askama for more general applications. For example, the project I'm looking at right now is templating markdown and LaTeX. In the LaTeX case, I might want to escape special latex characters (e.g. `&` to `\&` or `$` to `\$`, but not escape these when outputting the same strings to the markdown template. It would be nice to have some way of defining this mapping, and being able to give it a specific name like we can do with the syntaxes.

djc:
In that case I think there is a relatively straightforward solution. I propose that we indeed define another table structure in the configuration, which could look something like this:

```rust
[escape.latex]
filter = "latex_mod::escape"
extensions = ["tex"]
```

This will enable the `latex` escape mode automatically for extensions with the extension `tex` (it is an error to create escape modes with overlapping extensions or to define escape modes with names that conflict with the built-in `html` escape mode). Askama will call the named function when content needs to be escaped (that is, when `AsRef<str>`-type content is injected into the template output); it is on the template user to make sure that the escape function or module is in scope of template context types.

Does that sound right?

djc:
This is probably nice-to-have for 0.8, and might not make it.

botika:
We can use https://doc.rust-lang.org/cargo/reference/build-scripts.html#case-study-code-generation for generate `fn escape` and include in `askama_escape/src/lib.rs`. To generate this function, I have built this macros [`new_escape`](https://github.com/rust-iendo/v_htmlescape/blob/any_escape/v_htmlescape/src/lib.rs#L14-L17) 

```
new_escape!(
    HTMLEscape,
    "60->&lt; || 62->&gt; || 38->&amp; || 34->&quot; || 39->&#x27; || 47->&#x2f; || "
);
```
Than through the syntax "`$u8->$quote || `" generate the `Escaped` structure with any pair.

You can write from askama.toml and have the most common, (latex, html, etc,) saved.

[...]

Issue #108, opened 2018-09-01
recognize a template instance and don't bother escaping it


droundy:
A lovely feature would be to figure out how to recognize that an expression is a type that has derived Template, and then not escape its output (assuming both templates involved are html).  I'm not sure how to implement this, but it took me a while to figure out that I needed {{ var|safe }} to avoid having the html escaped.

I can't imagine a scenario where you'd have a type with an html template, and then want that output escaped when the variable is shown in another template.

djc:
Yeah, this is a bit tricky. It's been part of the contract that `Template` implementors also implement `Display`, and so without specialization I'm not sure it's possible to do have both of those features.

A somewhat similar issue comes up with non-stringy types (like numbers) which ideally you wouldn't escape for performance reasons.

One way this maybe could work is if there is a crate-internal trait `SafeToInclude` (bikeshed needed) which is implemented for known-safe types like `Template` and `i*` and `u*` and other things which only defers to `Display` if there is no other implementation.

droundy:
Yeah, specialization does seem it will/would be the "right" way to do this.  I'm not sure how your crate-internal `SafeToInclude` trait would work.  You couldn't define it for all possible types, so you'd still be left with the same problem in determining if it were to work properly, right?

Unless you were to restrict the template expansion to *only* accept expressions with type `SafeToInclude`, and then users would be expected to implement that trait for any non-`Template` types that they want to use in their templates? That does seem practical, and would have the added feature that users could enable their custom types to have a different `Display` appearance from their included appearance, which actually seems like a real plus.  Several of my tiny templates really could have been rust code even more cleanly.

Actually, stepping back, couldn't you just use `Template` itself to handle `SafeToInclude`? If you implemented `Template` on all the standard concrete types, and then required expression values to implement `Template`, then you'd need no extra bikeshedding.  The only downside I see is that you might want to have some concrete types have distinct rendering in multiple formats, e.g. html versus latex, imagining that you supported latex files (with escaping) in the future.  But perhaps that could just work by abusing the `extension` method?

droundy:
I've been wondering if you'd be open to a breaking change on this, obviously with an accompanying version bump?

[...]

PR #167, opened 2018-12-04
Add specialization

botika:
Nightly
```
Big table               time:   [280.96 us 281.06 us 281.20 us]
                        change: [-56.768% -56.745% -56.721%] (p = 0.00 < 0.05)

Teams                   time:   [485.02 ns 486.16 ns 487.31 ns]
                        change: [-43.564% -43.426% -43.294%] (p = 0.00 < 0.05)
```

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1,6 +1,8 @@
 use super::{get_template_source, Context, Heritage};
 use crate::input::TemplateInput;
-use crate::parser::{Cond, Expr, MatchParameter, MatchParameters, MatchVariant, Node, Target, When, WS};
+use crate::parser::{
+    Cond, Expr, MatchParameter, MatchParameters, MatchVariant, Node, Target, When, WS,
+};
 use askama_shared::filters;
 
 use proc_macro2::Span;
@@ -773,14 +775,14 @@ impl<'a> Generator<'a> {
             Expr::MethodCall(ref obj, method, ref args) => {
                 self.visit_method_call(buf, obj, method, args)
             }
-            Expr::RustMacro(name, ref args) => self.visit_rust_macro(buf, name, args),
+            Expr::RustMacro(name, args) => self.visit_rust_macro(buf, name, args),
         }
     }
 
-    fn visit_rust_macro(&mut self, buf: &mut Buffer, name: &str, args: &[Expr]) -> DisplayWrap {
+    fn visit_rust_macro(&mut self, buf: &mut Buffer, name: &str, args: &str) -> DisplayWrap {
         buf.write(name);
         buf.write("!(");
-        self._visit_args(buf, args);
+        buf.write(args);
         buf.write(")");
 
         DisplayWrap::Unwrapped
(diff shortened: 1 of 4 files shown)


botika:
At stable, the average escape time has been improved, remove the false positives, ascii 34 to 62, by means of a static table, it does not use simd. 

At nightly, simd has been integrated along with specialization, reduction of false positives to 1/64, full coverage and optimizations.

djc:
So that's a big improvement on nightly, but it looks like the improvement comes largely from specialization (more so than from the SIMD implementations). I'm still unconvinced about the benefits of SIMD by itself, and also not sure about the complexity here.

Here's some alternative thinking: what if we can make turn `MarkupDisplay` into a trait such that escaping is more pluggable? We should solve #108 and #136 at the same time as allowing you to plug in your custom escaping code.

botika:
I agree, the best improvement in this e2e is offered by the specialization. 

The complexity is high but the performance is 10% below memchr3. It also has a relative coverage of 100% and only works at nightly. In stable it works by static table with what gives better performance for 2 or more false positives than the previous one.

 What do you propose for SIMD? for the results of other implementations, it will not have much better performance. We leave it or what else can I do?

As for the trait, we can be done but we can only implement it on primitive types (usize, i8, str, String, ..), so it will stop working in other structures that imp Display. That or I do not just understand you.

As for the #108 and #136, I do not understand it very well, I had thought about the specialization for these. You can explain a little more about this solution for the this issues.

botika:
I enclose the [benchmarks](https://github.com/rust-iendo/v_htmlescape/tree/master/benches#benches) with throughput , they are the same data that [rust-memchr](https://github.com/BurntSushi/rust-memchr/tree/master/bench/data)

botika:
The main implementation of SIMD, avx, [loses performance below 11 bytes.](https://github.com/rust-iendo/v_htmlescape/blob/master/benches/src/all.rs#L32-L40)

I propose an operator for expressions, such as the `WS`, to use one or the other depending on whether the operator is or not. Also integrate it so that you can choose different types of escape, latex ...., which could also use SIMD with [this abstraction of the main loop](https://github.com/rust-iendo/v_htmlescape/blob/master/src/avx/loops.rs).

As for the issue #136 , I can write a [static table](https://github.com/rust-iendo/v_htmlescape/blob/master/src/utils.rs#L1-L32) at compile time from askama.toml with the [scalar implementation](https://github.com/rust-iendo/v_htmlescape/blob/master/src/utils.rs#L109-L123). With simd it is more complicated, but it can also be done simply by [calculating the 3 minimum ranges](https://github.com/rust-iendo/v_htmlescape/blob/master/src/avx/mod.rs#L16-L22), later like the previous one.

djc:
I'm sorry -- I don't have the energy right now to dig into this, and I sometimes find it hard to understand what exactly you mean; there's a bit of a language barrier still.

By the implementation of #136/#108 I mean that I think it would be useful to parametrize `Template` and `MarkupDisplay` with a type argument that's bound by an `Escaper` trait which has a single `escape()` method (similar to the current `MarkupDisplay::escape()` method). Instead of just a table of characters to escape in `askama.toml`, we would then configure the name of the `Escaper` impl to use, like this:

```toml
[escape]
html = "v_escape::HtmlEscaper"
```

If nothing is configured, it would default to the `askama_escape::HtmlEscaper`.

botika:
Ok, I only added the specialization.

botika:
I know, damn English, I need linguistic immersion. What I'm trying to say: let's merge the specialization and then we fix this issues.

PR #170, opened 2018-12-07
Fix Travis

djc:


botika:
It's a failure of fmt. Now I upload the fix.

PR #173, opened 2018-12-07
2018

djc:
This also updates the Rocket dependencies.


djc:
Thanks for the review!

Issue #174, opened 2018-12-09
[2018 branch] uri! macro from Rocket 0.4 leads to panic in proc-macro


mjanda:
Seems to my that new typer URIs from Rocket 0.4 don't work in askama templates when they contain any parameters.

Simple `uri!`s without parameters work.

`{{ uri!(crate:foo::bar: foo = bar) }}` will lead to:

```
... proc-macro derive panicked
... message: unable to parse template:
```

djc:
(Note that the 2018 branch has been merged into master already.)

Oops, I guess our parsing for Rust macro (as opposed to template macro) arguments is a little too restrictive to handle what's actually allowed in macros (and thus expected to work for users).

To be exact, the [`rust_macro` parser](https://github.com/djc/askama/blob/master/askama_derive/src/parser.rs#L431) uses the [`arguments` parser](https://github.com/djc/askama/blob/master/askama_derive/src/parser.rs#L274), but that only accepts a comma-separated list of simple expressions.

Widening that enough could be tricky... perhaps the Rust macro support should just pass through all the tokens unparsed?

botika:
@djc It's like you say.

I leave you two options, detecting nested parentheses #176 or a more complicated invocation end `);` #178 .

For me the best option is the #178.

djc:
@botika great work! Why do you prefer the latter option? It seems to me that, while the former is maybe a bit more code (and this uglier), it might also be more correct (since ultimately `);` can also appear in macro arguments). Or am I missing something?

botika:
If you have the argument "(" we would have an unexpected error. In the last there is not this problem.

botika:
Thanks for reporting.

mjanda:
Thank you, you're wonderful

PR #176, opened 2018-12-10
add rust macro arguments detect nested parenthesis

botika:
#174


botika:
Or this,  can not put incomplete parentheses, or I can put a `);` to close the macro. I think both ways are good. The one you like the most :smiley_cat:

djc:
So after giving it some thought, let's go with this approach. But as you said, it's possible for parenthesis to appear inside quotes, too. So, could you try to make the code robust against that? This would involve counting whether the current number of single quotes and double quotes is even, and correcting for quote characters being preceded by an odd number of backslash characters.

botika:
close in favor of #178

Issue #175, opened 2018-12-10
Gotham IntoResponse feature?


rakenodiax:
My favorite stack includes both this crate and the `gotham` web framework. It'd be awesome to have an implementation of its `IntoResponse` trait to make rendering that much simpler and ergonomic. I'm happy to work on this feature and submit a pull request, but wanted to make sure this would be of interest as a crate feature.

djc:
Sure, bring it on! Search for "rocket" or "iron" to see all the pieces on how to put it all together.

botika:
Gotham is a hyper wrapper, really functionality is not Gotham but hyper. So it would make more sense in the feature `with-hyper`. The current implementation #179  , removing the references to Gotham, would work for both Hyper and Gotham cases.

rakenodiax:
@botika The distinction is that `gotham::IntoResponse` is not implemented for `Into<hyper::Response<hyper::Body>>`, so if `Template` were to implement the latter instead, `Template::into` would have to be called manually:

```rust
struct FooBar;

impl Into<hyper::Response<hyper::Body>> for Foobar { ... }

fn hello(state: State) -> (State, impl IntoResponse) {
    (state, FooBar{}.into())
}
```

botika:
Ok, thanks

PR #177, opened 2018-12-10
Add optional Gotham support

rakenodiax:
This pull request implements the Gotham web framework's `IntoResponse` trait from #175, allowing items implementing `askama::Template` to be returned directly from Gotham handlers, like so:

```rust
#[derive(Template)]
struct HelloTemplate<'a> {
    name: &'a str,
}

fn hello(state: State) -> (State, HelloTemplate<'static>) {
    (state, HelloTemplate { name: "world" })
}

fn main() {
    gotham::start("127.0.0.1:7878", || Ok(hello));
}
```

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -478,6 +478,36 @@ pub mod actix_web {
     }
 }
 
+#[cfg(feature = "with-gotham")]
+pub mod gotham {
+    extern crate gotham;
+    extern crate hyper;
+    extern crate mime_guess;
+
+    pub use self::gotham::handler::IntoResponse;
+    use self::gotham::helpers::http::response::create_empty_response;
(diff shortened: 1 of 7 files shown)


askama/src/lib.rs
```diff
@@ -478,6 +488,36 @@ pub mod actix_web {
     }
 }
 
+#[cfg(feature = "with-gotham")]
+pub mod gotham {
+    extern crate gotham;
+    extern crate hyper;
+    extern crate mime_guess;
```
botika: @djc In Rust 2018 we should not eliminate these `extern crate`?

rakenodiax: I've removed them in the latest commit

rakenodiax:
I had an issue squashing the commits, and had to recreate the branch. I'll create another pull request

PR #178, opened 2018-12-10
fix rust macro arguments

botika:
#174 

this or #176

### DIFF
--- a/askama_derive/src/parser.rs
+++ b/askama_derive/src/parser.rs
@@ -21,7 +21,7 @@ pub enum Expr<'a> {
     Range(&'a str, Option<Box<Expr<'a>>>, Option<Box<Expr<'a>>>),
     Group(Box<Expr<'a>>),
     MethodCall(Box<Expr<'a>>, &'a str, Vec<Expr<'a>>),
-    RustMacro(&'a str, Vec<Expr<'a>>),
+    RustMacro(&'a str, &'a str),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


askama_derive/src/parser.rs
```diff
@@ -290,6 +291,38 @@ named!(arguments<Input, Vec<Expr>>, do_parse!(
     (args.unwrap_or_default())
 ));
 
+named!(macro_arguments<Input, &str>, alt!(
+    delimited!(char!('('), nested_parenthesis, char!(')')) |
+    map!(delimited!(char!('('), take_until!(");"), tag!(");")),
```
djc: I don't want this `);` pattern in there, I don't think it fits with the rest of the design since it doesn't make sense to me that a template expression ends in `;`.

askama_derive/src/parser.rs
```diff
@@ -290,6 +291,38 @@ named!(arguments<Input, Vec<Expr>>, do_parse!(
...
+        |s| str::from_utf8(&s).unwrap()
+    )
+));
+
+fn nested_parenthesis(i: Input) -> Result<(Input, &str), nom::Err<Input>> {
```
djc: I do think this should be extended to be robust against parentheses inside single or double unescaped quote characters.

botika: Then we will also have to deal with the escape cases' \ '' and "\"

askama_derive/src/parser.rs
botika: and something less obvious like `))` could I get into?

askama_derive/src/parser.rs
botika: Ok, done with escape.

askama_derive/src/parser.rs
botika: ok, without special end

testing/templates/rust-macro-args.html
```diff
@@ -0,0 +1,26 @@
+{{
+        call_a_or_b_on_tail!(
+            (a: compute_len, b: zero),
+                the recursive part that skips over all these
+                tokens doesn't much care whether we will call a
```
botika: The single quotes would fail in this case.

djc: Good point! I'm okay with that, though, because this would also fail if you put this text in a normal Rust macro call.

botika: Ok, this pull request is already finished.

askama_derive/src/parser.rs
```diff
@@ -290,6 +290,53 @@ named!(arguments<Input, Vec<Expr>>, do_parse!(
...
 ));
 
+named!(macro_arguments<Input, &str>,
+    delimited!(char!('('), nested_parenthesis, char!(')'))
+);
+
+fn nested_parenthesis(i: Input) -> Result<(Input, &str), nom::Err<Input>> {
+    let mut nested = 0;
+    let mut last = 0;
+    let mut in_str= false;
```
djc: Nit: space before `=` please.

askama_derive/src/parser.rs
```diff
@@ -290,6 +290,53 @@ named!(arguments<Input, Vec<Expr>>, do_parse!(
...
+    for (i, b) in i.iter().enumerate() {
+        if !(*b == b'(' || *b == b')') || !in_str {
+            match *b {
+                b'(' => nested += 1,
+                b')' => {
+                    if nested == 0 {
+                        last = i;
+                        break;
+                    }
+                    nested -= 1
```
djc: Nit: add `;` after this statement, please.

askama_derive/src/parser.rs
```diff
@@ -290,6 +290,53 @@ named!(arguments<Input, Vec<Expr>>, do_parse!(
...
+                },
+                b'"' => {
+                    if in_str {
+                        if !escaped {
+                            in_str = false;
+                        }
+                    } else {
+                            in_str = true;
```
djc: Nit: dedent this one block.

askama_derive/src/parser.rs
```diff
@@ -290,6 +290,53 @@ named!(arguments<Input, Vec<Expr>>, do_parse!(
...
+                    }
+                },
+                b'\\' => escaped = !escaped,
```
djc: Nit: Surround this with `{}`, put on its own line.

botika: done. need reaproval.

djc:
Thanks!

PR #179, opened 2018-12-10
Implement optional Gotham support

rakenodiax:
This pull request implements the Gotham web framework's `IntoResponse` trait from #175 , allowing items implementing `askama::Template` to be returned directly from Gotham handlers, like so:

```rust
#[derive(Template)]
struct HelloTemplate<'a> {
    name: &'a str,
}

fn hello(state: State) -> (State, HelloTemplate<'static>) {
    (state, HelloTemplate { name: "world" })
}

fn main() {
    gotham::start("127.0.0.1:7878", || Ok(hello));
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ in a for-profit context, please consider supporting my open source work on
 * Construct templates using a familiar, easy-to-use syntax
 * Template code is compiled into your crate for [optimal performance][benchmarks]
 * Benefit from the safety provided by Rust's type system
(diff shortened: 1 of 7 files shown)


djc:
Looks great! Can you check that the new test is actually exercised in CI? I think the `full` feature in the testing crate needs to be expanded.

rakenodiax:
@djc  Added the tests, thanks for the catch!

PR #182, opened 2018-12-12
Allow using brackets for enums in `match`

mcarton:


djc:
Thanks for doing the work on this!

PR #188, opened 2019-01-05
Fix support for tuple

yossyJ:
Thank you for your previous approval!

I fixed it for Jinja style, like so:
`{% for s1, s2 in tuple_strings %}`

And supported trailing comma for something like auto generated templates.

### DIFF
--- a/askama_derive/src/parser.rs
+++ b/askama_derive/src/parser.rs
@@ -276,23 +276,6 @@ named!(target_single<Input, Target>, map!(identifier,
     |s| Target::Name(s)
 ));
 
-named!(target_plural<Input, Target>, do_parse!(
-    args: do_parse!(
-        arg0: ws!(identifier) >>
-        args: many1!(do_parse!(
-            tag!(",") >>
-            argn: ws!(identifier) >>
-            (argn)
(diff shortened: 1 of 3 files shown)


djc:
This makes a lot of sense, but I'd first like to understand why Rust itself doesn't allow this style of loop. Presumably there's some kind of grammar conflict in Rust, and before we make it impossible to express some things in the Askama template language, I want to know what that might be.

yossyJ:
I see. I had thought you might think the Askama template language must be compatible with Jinja.
But, for me, this is not necessary grammar.
Therefore, I have no problem with rejecting this pull request.
How about trailing comma in tuple?

djc:
Yeah, a decision I made early on was that I didn't want to stick to Jinja's syntax exactly, but would rather build a new syntax that's more aligned with Rust.

Definitely want support for trailing comma in tuples!

yossyJ:
O.K.
I fixed it to leave only support for trailing comma.

yossyJ:
Oh! ws required. I'll fix later.

yossyJ:
I fixed it.

PR #189, opened 2019-01-05
Add support for loop.last

yossyJ:
I managed to support for loop.last.
If necessary, please move ::askama::helpers::enumerate to appropriate position.


yossyJ:
That is nice advice!
I fixed it, and removed count() and size_hint() that are unlikely to be used.

djc:
Very nice, thanks!

PR #193, opened 2019-01-08
Add support for #![feature(non_ascii_idents)]

yossyJ:
I hope for non_ascii_idents feature to be supported.


yossyJ:
Thank you for approvals.
The Askama is an excellent template engine!

PR #197, opened 2019-01-10
Custom output formats & escaping

djc:
I still want to tweak the names of things, but this could use a first round of review.

When this is merged I'd like to release 0.8.0 soon after.

@droundy @lukehsiao would like to hear your thoughts!

### DIFF
--- a/askama_shared/src/lib.rs
+++ b/askama_shared/src/lib.rs
@@ -4,6 +4,7 @@ extern crate serde_derive;
 
 use toml;
 
+use std::collections::HashSet;
 use std::env;
 use std::fs;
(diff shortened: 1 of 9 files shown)


askama_shared/src/lib.rs
```diff
@@ -350,4 +384,23 @@ mod tests {
 
         let _config = Config::new(raw_config);
     }
+
+    #[test]
+    fn escape_modes() {
```
lukehsiao: I see this example of adding a new escape mode, is there an example of showing how to define those new escapes?

For example, defining a new custom escape for JS, and then testing that it is properly escaped.

PR #198, opened 2019-01-10
Provide size_hint() method on Template (fixes #181)

### LINKED ISSUE #181: Implement size_hint() method on Template trait
This is pretty much the same as #153, but using the more precise heuristics as described in https://github.com/djc/askama/pull/153#issuecomment-437654909. This would also be nice to have in 0.8.

djc:


djc:
Can you re-approve?

PR #199, opened 2019-01-10
Use include_bytes macro to avoid the need for a build script

de-vri-es:
This PR adds an otherwise unused `include_bytes!(...);` statement in the generated template code for each template in the context map and for each included template. This allows the compiler to detect file level dependencies without the need for a custom build script.

I believe I caught all the relevant file dependencies: extends, imports, includes and of course the template itself (if it is backed by a file, which requires some checks).

This gives more granular file-level dependency information than the build script, although I don't believe that actually gives any practical benefits right now.

Feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,23 +68,8 @@ First, add the following to your crate's `Cargo.toml`:
 
 ```toml
 # in section [dependencies]
-askama = "0.7"
+askama = "0.8"
 
-# in section [build-dependencies]
-askama = "0.7"
-```
-
-Because Askama will generate Rust code from your template files,
-the crate will need to be recompiled when your templates change.
-This is supported by adding a build script, `build.rs`, to your crate.
-It needs askama as a build dependency:
-
-```rust
-extern crate askama;
-
-fn main() {
-    askama::rerun_if_templates_changed();
-}
 ```
 
 Now create a directory called `templates` in your crate root.
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -515,35 +515,8 @@ pub mod gotham {
     }
 }
 
-fn visit_dirs(dir: &Path, cb: &dyn Fn(&DirEntry)) -> io::Result<()> {
-    if dir.is_dir() {
-        for entry in fs::read_dir(dir)? {
-            let entry = entry?;
-            let path = entry.path();
-            if path.is_dir() {
-                visit_dirs(&path, cb)?;
-            } else {
-                cb(&entry);
-            }
-        }
-    }
-    Ok(())
-}
-
-/// Build script helper to rebuild crates if contained templates have changed
+/// Old build script helper to rebuild crates if contained templates have changed
 ///
-/// Iterates over all files in the template directories and writes a
-/// `cargo:rerun-if-changed=` line for each of them to stdout.
-///
-/// This helper method can be used in build scripts (`build.rs`) in crates
-/// that have templates, to make sure the crate gets rebuilt when template
-/// source code changes.
-pub fn rerun_if_templates_changed() {
-    let file = read_config_file();
-    for template_dir in &shared::Config::new(&file).dirs {
-        visit_dirs(template_dir, &|e: &DirEntry| {
-            println!("cargo:rerun-if-changed={}", e.path().to_str().unwrap());
-        })
-        .unwrap();
-    }
-}
+/// This function is now deprecated and does nothing.
+#[deprecated(since="0.8.1", note="file-level dependency tracking is handled automatically without build script")]
+pub fn rerun_if_templates_changed() {}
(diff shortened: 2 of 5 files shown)


djc:
Very nice, thanks for implementing this! I've thought about it myself a few times. Have you checked that this has the desired effect? (That is, touching a template file will trigger recompilation even without a build script.)

One issue here is that I think it should probably include bytes for all templates used, including inherited templates, imported templates and included templates. That might be a bit more work to get right, and will actually need something that I've so far resisted doing, which is walking the entire template AST during the `Context` creation phase.

Are you up for taking a look at that? It basically means adapting `Context::new()` to do a full traversal of the root template and any inherited templates to discover what templates are used, and then including bytes for all of them.

djc:
Update: oh wait, you have handled all that stuff! More power to you!

So, that means only the includes probably don't work correctly right now.

de-vri-es:
I believe I handled includes properly too ;)

I did test locally, and the rebuilds worked as intended. I did not test multi-level includes or extends but from the code I believe it should work correctly recursively already.

de-vri-es:
Hmm, it looks like the path comparison to avoid `include_bytes!()` for the fake path of templates from rust source break on Windows. Or otherwise something else is going wrong on Windows =]

de-vri-es:
I also tested if a rebuild is properly triggered for nested includes, imports and extends now. On Linux, it works as expected. That still leaves the windows issues to fix.

djc:
Did you test includes in a for or if block?

de-vri-es:
I did now. Regardless of the true/false value of the condition, changes to the included file trigger a rebuild (which makes sense, since the include bytes is done compile time for the generated code, not runtime).

I think the windows issue is caused by backslashes being parsed into escape sequences somewhere. Not sure why that happens though.

de-vri-es:
Oh, of course. We're outputting source and there's backslashes in the string literal. Is there a proper way to escape string literals in the output of a procedural macro?

djc:
Not sure, maybe look at the quote crate API?

djc:
Also as follow-up (ideally in a separate commit) get rid of all the build script stuff? That includes in the README, the build script in the test crate and the helper code in the crates that supports the build script API.

de-vri-es:
Right, I was meaning to ask if it should be deprecated, but if it can be removed, awesome :) Will do once the tests all pass.

In the mean time, to give you an idea of what the resulting generated code looks like, this was generated from a template with nested imports, nested extends and a conditional include with a nested include (now using `quote!{}` to fix the generated output if it has escape sequences):
```rust
impl < 'a > ::askama::Template for Hello< 'a > {
    fn render_into(&self, writer: &mut ::std::fmt::Write) -> ::askama::Result<()> {
        include_bytes ! ( "/home/maarten/dev/rust/lummel/templates/base2.html" ) ;
        include_bytes ! ( "/home/maarten/dev/rust/lummel/templates/base.html" ) ;
        include_bytes ! ( "/home/maarten/dev/rust/lummel/templates/import1.html" ) ;
        include_bytes ! ( "/home/maarten/dev/rust/lummel/templates/import2.html" ) ;
        include_bytes ! ( "/home/maarten/dev/rust/lummel/templates/hello.html" ) ;
        write!(
            writer,
            "<!DOCTYPE html>\n<html>\n<head>\n\t<title>{expr0}</title>\n\t\n\t<link rel=\"stylesheet\" type=\"text/css\" href=\"/css/main.css\">\n\t\n</head>\n<body>\n<p>Hello world!</p>\n",
            expr0 = &::askama::MarkupDisplay::from(&self.title),
        )?;
        if self.include_foo {
            writer.write_str("\n")?;
            include_bytes ! ( "/home/maarten/dev/rust/lummel/templates/foo.html" ) ;
            writer.write_str("FOO\n")?;
            include_bytes ! ( "/home/maarten/dev/rust/lummel/templates/foo2.html" ) ;
            writer.write_str("FOO2")?;
            writer.write_str("\n")?;
        }
        writer.write_str("\n</body>\n</html>")?;
        Ok(())
    }
    fn extension() -> Option<&'static str> {
        Some("html")
    }
}
impl < 'a > ::std::fmt::Display for Hello< 'a > {
    }
}
```

de-vri-es:
Force pushed to re-trigger CI, it now passed =)

djc:
Yeah, can you remove the note from the README again? Also I changed my mind, so maybe bring back `rerun_if_templates_changed` with an empty implementation and a `#[deprecated]` attribute?

After that I'm done, promise.

de-vri-es:
Righty, removed the note from the readme and re-added the old function, but deprecated and with updated doc.

djc:
Thanks, great work!

de-vri-es:
Thanks for the review :)

Issue #204, opened 2019-01-25
Responder trait derivation for actix-web


ZizhengTai:
I'm experimenting with askama and actix-web, and noticed the `Responder` trait derivation sometimes doesn't work. Here is an example project:

Cargo.toml:
```toml
[package]
name = "try-askama"
version = "0.1.0"
authors = ["Z"]
edition = "2018"

[dependencies]
actix-web = "0.7"
askama = { version = "0.7", features = ["with-actix-web"] }
```

src/main.rs:
```rust
use actix_web::{App, HttpRequest};
use askama::Template;

#[derive(Template)]
#[template(ext = "html", source = "Hello, {{ name }}!")]
struct HelloTemplate {
    name: &'static str,
}

fn index(_: &HttpRequest) -> HelloTemplate {  // <-- Compiles
// fn index(_: &HttpRequest) -> impl Template { // <-- DOESN'T compile
    HelloTemplate { name: "John" }
}

fn main() {
    App::new().resource("/", |r| r.get().f(index));
}
```

If I try to make the handler return `impl Template` (or anything like `Result<impl Template>`, `Box<Future<impl Template, Error>>`, etc.), I get this error:

```
error[E0277]: the trait bound `impl askama::Template: actix_web::handler::Responder` is not satisfied
  --> src/main.rs:15:42
   |
15 |     App::new().resource("/", |r| r.get().f(index));
   |                                          ^ the trait `actix_web::handler::Responder` is not implemented for `impl askama::Template`
```

botika:
The correct type of return for this case is `actix_web::Responder` .
```rust
use actix_web::{App, HttpRequest, Responder};
use askama::Template;

#[derive(Template)]
#[template(ext = "html", source = "Hello, {{ name }}!")]
struct HelloTemplate {
    name: &'static str,
}


fn index(_: &HttpRequest) -> impl Responder { 
    HelloTemplate { name: "John" }
}

fn main() {
    App::new().resource("/", |r| r.get().f(index));
}
```
As you can see in the [documentation](https://actix.rs/actix-web/actix_web/trait.Responder.html), the correct trait for generate responses for clients is `Responder`.

@djc Can close it?

djc:
@botika good call! @ZizhengTai does that work for you?

ZizhengTai:
Thanks @botika @djc, this should work for me.

PR #205, opened 2019-01-27
Add TemplateResponder trait for integration with actix-web

ZizhengTai:
This is similar to what `actix_web::AsyncResponder` does. It enables us to call `.responder()` on a template to render it into an `actix_web::HttpResponse`. This is advantageous over the current approach (implementing `actix_web::Responder` for `T: Template`) because the current approach doesn't allow returning templates with references to local variables. Please see the added test for what I mean.

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -492,6 +492,21 @@ pub mod actix_web {
         let ctype = get_mime_type(ext).to_string();
         Ok(HttpResponse::Ok().content_type(ctype.as_str()).body(rsp))
     }
+
+    pub trait TemplateResponder {
+        fn responder(&self) -> Result<HttpResponse, Error>;
+    }
+
+    impl<T: super::Template> TemplateResponder for T {
+        fn responder(&self) -> Result<HttpResponse, Error> {
+            let rsp = self
+                .render()
+                .map_err(|_| ErrorInternalServerError("Template render error"))?;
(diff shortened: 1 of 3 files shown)


djc:
Should we remove the current approach, then, or does it address a different use case?

ZizhengTai:
The current approach can be removed. I just don't know if you are fine with
the breaking change.

gregelenbaas:
I don't think the current approach should be removed. I like the addition for flexibility but I think it would cause unnecessary breakage as well as no longer allowing the shorter syntax when local variables do not need to be computed.

djc:
Closing, no longer necessary (see #204).

ZizhengTai:
@djc, I think there is a misunderstanding. This PR isn't doing the same thing as #204. #204 only provides a way to return a template without specifying the concrete return type (but instead with a `impl Responder`), but it doesn't solve the problem that the returned template cannot contain references to local string slices. The example in #204 compiles only because the reference has static lifetime; if you change it to a `'a` (which is what this PR allows for), the code doesn't compile.

Could you please reopen this PR and review it? Thank you.

ZizhengTai:
Any updates on this?

djc:
So as I understand we can only generate code for the new approach, remove the old `respond()` implementation and the code generation for the stuff that's there now, and almost all of the current users' code would stay working. Can someone explain to me why that might not be true?

PR #206, opened 2019-02-05
Make html-escaper also escape .j2, .jinja and .jinja2 files

Weasy666:
The content of for .j2, .jinja and .jinja2 files shouldn't be anything else than HTML and Jinja syntax, so i just used the existing html-escaper for these file types.

I did this because i'm developing with VS Code and am using the [Better Jinja](https://marketplace.visualstudio.com/items?itemName=samuelcolvin.jinjahtml)-Extension which provides syntax highlighting for jinja(2) including HTML, Markdown and YAML templates.

I hope this is ok for you or should i add a dedicated `::askama::Jinja` function which directly calls `::askama::Html`?


Weasy666:
Great!

PR #208, opened 2019-02-14
Fully qualified Result

spease:
This fixes a "wrong number of type arguments" error if the point of inclusion defines its own Result with a specific type.


djc:
Thanks, merged,

Issue #209, opened 2019-02-15
Does not render UTF-8 correctly


ghost:
This template:

```
<!doctype html>
<html>
<head>
<title>Index</title>
</head>
<body>
↰ ..
</body>
</html>
```

results in:

```
â†° .. 
```

Even trying to explicitly set the charset to UTF-8 fails:
```
let body = render_template(&filepath)
            .respond_to(&request)?
            .replace_body(actix_web::Body::Empty);
        Ok(HttpResponse::Ok()
            .content_type("text-html; charset=utf-8")
            .body(body))
```

botika:
The problem is in the implementation of actix_web. Try that template in this example, https://github.com/djc/askama/blob/master/testing/tests/actix_web.rs, and recomment, please.

djc:
I'm skeptical that Askama is the problem here, since it has tests to make sure it handles UTF-8 correctly (though not specifically the Actix-Web test). Also, I don't know what your `render_template()` snippet is supposed to do, but it does not look like that's using Askama.

pmi-spease:
Here's a quick example:

```
#[derive(Template)]
#[template(path = "chart.html")]
struct ChartTemplate<I>
where
    I: IntoIterator<Item = DataPoint> + ::serde::Serialize,
{
    data: I,
}

#[derive(Serialize)]
struct DataPoint {
    x: f64,
    y: f64,
}

impl From<(f64, f64)> for DataPoint {
    fn from((x, y): (f64, f64)) -> Self {
        DataPoint { x, y }
    }
}

fn stream_chart_page(
    data: impl IntoIterator<Item = DataPoint> + serde::Serialize,
) -> BackendResult<impl Responder<Item = HttpResponse>> {
    Ok(ChartTemplate { data })
}
```

djc:
That doesn't actually show any string contents...

pmi-spease:
The test passed, but the issue is how Firefox interprets the content type. It has to be `text/html; charset=utf-8` for it to display the correct characters. This appears to be correct behavior from first pass:
https://www.w3.org/International/articles/http-charset/index

Default behavior with askama is to only show text/html. It looks like this is because of mime_guess - askama uses its output to override actix_web's default, but mime_guess doesn't include a charset:
```impl<T: super::Template> TemplateIntoResponse for T {
        fn into_response(&self) -> Result<HttpResponse, Error> {
            let rsp = self
                .render()
                .map_err(|_| ErrorInternalServerError("Template parsing error"))?;
[...]

PR #219, opened 2019-03-10
Default to UTF-8 mime type for actix-web and gotham

spease:
Fixes issue #209 for actix-web and gotham. Rocket uses its own internal Content function, so I left it alone. Also fixed the tests in askama_testing, which are what I used to validate the change.

NB I had to use if statements instead of match due to limitations of the mime type. I don't know why they're there.


djc:
Thanks for fixing this! I got rid of the big `if`-table in favor of a const array, fixed the formatting and a new test that had appeared since you branched off, and pushed the result as d1b47f8908f7ca71da1f767432bb33cb87bbee14.

PR #221, opened 2019-03-15
Don't force adding "&" to start of collection for TemplateLoop

gregelenbaas:
Currently the TemplateLoop generator always appends `&` to every expression. This is inconvenient as it doesn't allow iteration over collections that are not owned by a struct but are passed by reference. 

For example today this does not work: 

welcome.rs
```
#[derive(Template)]
#[template(path = "apps/welcome/welcome.html")]
struct Landing<'a> {
    locales: &'a HashMap<String, Locale>,
}

pub fn index(req: HttpRequest<AppState>) -> Result<HttpResponse, Error> {

    let locales = &req.state().locales;

    let landing = Landing {
        locales: locales,
    };

    landing.respond_to(&req)
}

```

welcome.html
```
<h1>Welcome!</h1>
<div>
    {% for locale in locales.values() %}
        <p>
            <a hreflang="{{ locale.key }}" href="/{{ locale.key }}/">{{ locale.description }}</a>
        </p>
    {% endfor %}
</div>
```
This code will produce a cannot move out of borrowed context error. 

The current work around is: 
welcome.rs
```
#[derive(Template)]
#[template(path = "apps/welcome/welcome.html", print = "all")]
struct Landing<'a> {
    locales: &'a HashMap<String, Locale>,
}

// TODO: Delete after upgrading askama
struct LocaleAskama<'a> {
    description: &'a str,
    key: &'a str,
}

pub fn index(req: HttpRequest<AppState>) -> Result<HttpResponse, Error> {

    let locales = &req.state().locales;
    let mut locales_parsed = Vec::<LocaleAskama>::new();

    for locale in locales.values() {
        locales_parsed.push(LocaleAskama{
            description: locale.description.as_str(),
            key: locale.key.as_str(),
        });
    } 

    let landing = Landing {
        locales: locales_parsed,
    };

    landing.respond_to(&req)
}
```

The bad thing is that this would be a breaking change as it would require collections that are owned by the template to have the & prefix. I think this should be the correct pattern, I couldn't figure out any other way to check of the template field was a reference or owned. 

Also, I noticed there is no CHANGELOG.md file, I can create one if you think it's a good idea.

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -296,6 +296,10 @@
 //! * *loop.index0*: current loop iteration (starting from 0)
 //! * *loop.first*: whether this is the first iteration of the loop
 //! * *loop.last*: whether this is the last iteration of the loop
+//! 
+//! Loops should be passed references to collections so that they are not 
+//! consumed by the template. If the template struct owns the collection that is 
+//! being passed to a loop, make sure to prefix the collection with an `&`.
 //!
 //! ### If
 //!
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -534,8 +534,38 @@ impl<'a> Generator<'a> {
                 ", _loop_item) in ::askama::helpers::TemplateLoop::new({}) {{",
                 expr_code
             )),
-            _ => buf.writeln(&format!(
+            Expr::Attr(expr, _) => {
+                let expr: &Expr = &**expr;
+                match expr {
+                    Expr::RefVar(_) => buf.writeln(&format!(
+                        ", _loop_item) in ::askama::helpers::TemplateLoop::new((&{}).into_iter()) {{",
+                        expr_code
+                    )),
+                    _ => buf.writeln(&format!(
+                        ", _loop_item) in ::askama::helpers::TemplateLoop::new(({}).into_iter()) {{",
+                        expr_code
+                    )),
+                }
+            }, 
+            Expr::MethodCall(expr, _, _) => {
+                let expr: &Expr = &**expr;
+                match expr {
+                    Expr::RefVar(_) => buf.writeln(&format!(
+                        ", _loop_item) in ::askama::helpers::TemplateLoop::new((&{}).into_iter()) {{",
+                        expr_code
+                    )),
+                    _ => buf.writeln(&format!(
+                        ", _loop_item) in ::askama::helpers::TemplateLoop::new(({}).into_iter()) {{",
+                        expr_code
+                    )),
+                }
+            }, 
+            Expr::RefVar(_) => buf.writeln(&format!(
                 ", _loop_item) in ::askama::helpers::TemplateLoop::new((&{}).into_iter()) {{",
+                expr_code
+            )),
+            _ => buf.writeln(&format!(
+                ", _loop_item) in ::askama::helpers::TemplateLoop::new(({}).into_iter()) {{",
                 expr_code
             )),
         };
@@ -863,6 +893,7 @@ impl<'a> Generator<'a> {
             Expr::NumLit(s) => self.visit_num_lit(buf, s),
             Expr::StrLit(s) => self.visit_str_lit(buf, s),
             Expr::Var(s) => self.visit_var(buf, s),
+            Expr::RefVar(s) => self.visit_ref_var(buf, s),
             Expr::Path(ref path) => self.visit_path(buf, path),
             Expr::Array(ref elements) => self.visit_array(buf, elements),
             Expr::Attr(ref obj, name) => self.visit_attr(buf, obj, name),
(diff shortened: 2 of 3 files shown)


djc:
There's actually a fairly extensive change log here: https://github.com/djc/askama/releases, but maybe it would help to link to it from the README to make it more discoverable?

djc:
Although I agree that this should work, I don't think this is the right direction for a solution. Can you show in more detail what causes the "cannot move out of borrowed context" problem in this case?

gregelenbaas:
Great, I'll update  https://github.com/djc/askama/releases . Definitely not sure if this is the right solution either but will be great to be able to use referenced collections. When I use 0.8 Askama here is what I get:

Code:
```
#[derive(Template)]
#[template(path = "apps/welcome/welcome.html", print = "all")]
struct Landing<'a> {
    locales: &'a HashMap<String, Locale>,
}

pub fn index(req: HttpRequest<AppState>) -> Result<HttpResponse, Error> {

    let locales = &req.state().locales;

    let landing = Landing {
        locales: locales,
    };

    landing.respond_to(&req)
}
```
Template
```
<h1>Welcome!</h1>
<div>
    {% for locale in locales.values() %}
        <p>
            <a hreflang="{{ locale.key }}" href="/{{ locale.key }}/">{{ locale.description }}</a>
        </p>
    {% endfor %}
</div>
```
Askama Output
```
[Lit("", "<h1>Welcome!</h1>\r\n<div>", "\r\n    "), Loop(WS(false, false), Name("locale"), MethodCall(Var("locales"), "values", []), [Lit("\r\n        ", "<p>\r\n            <a hreflang=\"", ""), Expr(WS(false, false), Attr(Var("locale"), "key")), Lit("", "\" href=\"/", ""), Expr(WS(false, false), Attr(Var("locale"), "key")), Lit("", "/\">", ""), Expr(WS(false, false), Attr(Var("locale"), "description")), Lit("", "</a>\r\n        </p>", "\r\n    ")], WS(false, false)), Lit("\r\n", "</div>", "")]
impl < 'a > ::askama::Template for Landing< 'a > {    fn render_into(&self, writer: &mut ::std::fmt::Write) -> ::askama::Result<()> {
        include_bytes ! (
"C:\\Users\\GregElenbaas\\workspace\\walkwiki\\templates\\apps/welcome/welcome.html"
) ;
        writer.write_str("<h1>Welcome!</h1>\r\n<div>\r\n    ")?;
        for (locale, _loop_item) in ::askama::helpers::TemplateLoop::new((&self.locales.values()).into_iter()) {
            write!(
                writer,
                "\r\n        <p>\r\n            <a hreflang=\"{expr0}\" href=\"/{expr0}/\">{expr1}</a>\r\n        </p>\r\n    ",
                expr0 = &::askama::MarkupDisplay::new_unsafe(&locale.key, ::askama::Html),
                expr1 = &::askama::MarkupDisplay::new_unsafe(&locale.description, ::askama::Html),
            )?;
        }
        writer.write_str("\r\n</div>")?;
        Ok(())
    }
    fn extension() -> Option<&'static str> {
        Some("html")
    }
    fn size_hint() -> usize {
        8
    }
}
impl < 'a > ::std::fmt::Display for Landing< 'a > {
    }
}
impl < 'a > ::askama::actix_web::Responder for Landing< 'a > {
    type Item = ::askama::actix_web::HttpResponse;
    type Error = ::askama::actix_web::Error;
    fn respond_to<S>(self, _req: &::askama::actix_web::HttpRequest<S>) -> ::std::result::Result<Self::Item, Self::Error> {
        use ::askama::actix_web::TemplateIntoResponse;
        self.into_response()
    }
```
Compiler Error
```
error[E0507]: cannot move out of borrowed content
  --> src\apps\welcome\mod.rs:11:10
   |
11 | #[derive(Template)]
   |          ^^^^^^^^ cannot move out of borrowed content

error: aborting due to previous error

For more information about this error, try `rustc --explain E0507`.
error: Could not compile `walkwiki`.

To learn more, run the command again with --verbose.
```

djc:
So now can you put the generated source code in your own source code, comment out the `derive` and `template` attributes so that we can see the precise error the compiler generates?

gregelenbaas:
Just checking @djc if you had any feedback

djc:
I suppose the problem here is that you want `(&self.locales).values()` but you're getting `&self.locales.values()` which is different in terms of ownership?

So far I've resisted adding `&` in the template language, and I don't like the idea of adding it. But right now I don't have any good ideas for workarounds for your problem.

gregelenbaas:
I understand not wanting to have to think about "&" in Askama templates. The issue is that I have some long lived collections of strings that are loaded at server start that are used to populate the text for various templates. Askama currently can't loop through collections held by the template struct that are references with lifetime annotations, and not owned by the struct itself. This PR fixes that issue but is breaking. I've been trying to research a way in Rust that we could detect a value is a reference instead of owned at run time. I am going to look into TryFrom but am not sure if that's the right way to go.

djc:
I wonder what we could do here if we add filters to the mix. That way, you could at least write Rust code that adapts your iterator, in Rust code, possibly to some trait defined in askama that doesn't get the mandatory reference treatment. Does that make sense?

zakcutner:
My issue with the current implementation is that it seems `into_iter()` only happens to work with types like `Vec` due to [its implementation under the hood](https://doc.rust-lang.org/src/alloc/vec.rs.html#1876-1883) which actually uses `iter()`.

Conversely, I get the rational behind this and it's definitely a nice idea to not consume iterables used in templates. However, I think that trying to implement `for` loops in a different way to how Rust does will ultimately always cause conflict with the language and confusion for users of this library.

It would be great to get this resolved because, at least for me, it currently presents quite a barrier to using Askama 😞

PR #223, opened 2019-03-17
Improve actix-web integration by creating bytes buffer instead of string

DoumanAsh:
It removes small overhead of converting from `Vec<u8>` to `Bytes`


djc:
Thanks!

PR #230, opened 2019-04-11
Fix doc for using custom filters

notsimon:
Contrary to what is stated in the doc, custom filters aren't accessed by prepending `filters::` in the templates. I fixed the documentation given what I read [in the code](https://github.com/djc/askama/blob/23ff7ad636b4c0e26d3242747d50b7123831a1a8/askama_derive/src/generator.rs#L940).


djc:
Thanks for this!

PR #231, opened 2019-04-15
WIP: Add raw block

fokinpv:
I'm trying to implement `raw` block like the one in the _Jinja_.

```
{% raw %}
  <h1>Hello, {{ name}}</h1> 
{% endraw %}
```

But I think I did it totally not how it supposed to be added, would be great to have an advice how to make it more idiomatic.


djc:
This looks pretty good! I'd prefer to not to have the `Input` type in the AST node type like this. Also, can you please add a test in `testing/tests` that exercises? Can go in `simple.rs` and should be pretty simple to write if you look at the other tests.

Thanks for working on this!

djc:
Looks great, thanks!

PR #232, opened 2019-04-20
Improve parsing of a 'raw' block

fokinpv:
Hello,

I want to improve parsing of a `raw` block. 

There is some hard-code in here
https://github.com/djc/askama/blob/master/askama_derive/src/parser.rs#L846

Added `take_raw_block` function which will take content of a `raw` block.
The minus I see now in this PR that I could not came up with a failure case for this function to return `Err`.


fokinpv:
Bad implementation, closed.

PR #233, opened 2019-04-21
[DNM] Improve raw block

fokinpv:
I want to improve parsing of a `raw` block. 

There is some hard-code in here
https://github.com/djc/askama/blob/master/askama_derive/src/parser.rs#L846

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -345,10 +345,9 @@ impl<'a> Generator<'a> {
                     self.flush_ws(m.ws1);
                     self.prepare_ws(m.ws2);
                 }
-                Node::Raw(ws1, contents, ws2) => {
+                Node::Raw(ws1, contents, _) => {
                     self.handle_ws(ws1);
                     self.buf_writable.push(Writable::Lit(contents));
-                    self.handle_ws(ws2);
                 }
                 Node::Import(ws, _, _) => {
                     if level != AstLevel::Top {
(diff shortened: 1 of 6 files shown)


fokinpv:
The syntax for a block start/end is configurable and can be changed. And whitespaces should be allowed for start/end block elements. Previous approach will not work in these cases.

askama_derive/src/generator.rs
```diff
@@ -345,10 +345,9 @@ impl<'a> Generator<'a> {
                     self.flush_ws(m.ws1);
                     self.prepare_ws(m.ws2);
                 }
-                Node::Raw(ws1, contents, ws2) => {
+                Node::Raw(ws1, contents, _) => {
                     self.handle_ws(ws1);
                     self.buf_writable.push(Writable::Lit(contents));
-                    self.handle_ws(ws2);
```
djc: This should probably be `self.prepare_ws(ws2);` instead of being removed completely?

askama_derive/src/parser.rs
```diff
@@ -838,20 +838,34 @@ named_args!(block_macro<'a>(s: &'a Syntax<'a>) <Input<'a>, Node<'a>>, do_parse!(
     })
 ));
 
+fn take_raw_block<'a>(
```
djc: So not at all clear what you're trying to do here. `block_start` has multiple characters, but you're only checking for the first one, so it doesn't look like this is handling things correctly...

fokinpv: Yes, indeed, will try to make it work correctly.

djc:
Hey, can you explain why you're closing this?

fokinpv:
I thought it is not good implementation and it's better to reopen when I redo this.

PR #235, opened 2019-04-24
docs/recursive: add documentation hints for recursive structures

### LINKED ISSUE #220: recurse by including self
I tried to use askama to follow down a tree like structure by using something like this:

```jinja
{# content of this.xyz #}
{%for item in item.children %}
{%if !item.is_leaf() %}
{% include "this.xyz" %}
{% endif %}
{% endfor %}
```

is this something that is supposed to be supported?

The actual result is a message:

```
help: message: attempt to multiply with overflow
```

drahnr:
Closes #220 sry for the long delay, rl.

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -402,6 +402,29 @@
 //!
 //! Askama supports block comments delimited by `{#` and `#}`.
 //!
+//!
+//! ## Recursive Structures
+//!
+//! Recursive implementations should preferably use a custom iterator and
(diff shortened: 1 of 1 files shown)


djc:
Fails the tests, could you take a look? I'm not quite sure why this happens, but it does seem related to your changes.

drahnr:
Doctests did not pass, fixed.

djc:
Thanks!

PR #236, opened 2019-05-03
Simplify parser with nom's list combinators

kazimuth:
This is a small standalone refactoring I did during my fluent changes. No semantic changes, just a slightly prettier parser using some of nom's combinators. `separated_list!` and `separated_nonempty_list!` do what you'd expect and return `Vecs`; `delimited!` just passes through the result of its middle parser.


djc:
Great, thanks!

PR #237, opened 2019-05-05
I18n runtime

kazimuth:
This PR adds an I18n runtime to `askama_shared`. There's no codegen yet, though.

This is a little wacky because we're reimplementing some features provided by fluent. (See discussion [here](https://github.com/projectfluent/fluent-rs/issues/64#issuecomment-489471933)).

Open questions:
- Should we add `log` or `slog` to register errors? Fluent is designed to degrade gracefully when errors are present.
- Currently, fluent requires allocating a `HashMap` of arguments for every call (that takes arguments). Should we instead add a `thread_local!` `HashMap` and transparently re-use it for each call? In my benchmarks this takes ~250ns off of every call to `localize` (which takes ~500ns on average). However, it's a little weird for Rust.
- I do a little bit of wacky pre-processing when choosing a locale fallback chain for a user, not sure if it's the right design. See my TODO: discuss comment in the code.

### DIFF
--- a/askama_shared/src/i18n_runtime.rs
+++ b/askama_shared/src/i18n_runtime.rs
@@ -46,16 +46,18 @@ impl<'a> StaticParser<'a> {
 
         let mut bundles = HashMap::new();
         let mut available = Vec::new();
-        for (locale, resource) in resources.0.iter() {
+        for (locale, resources) in resources.0.iter() {
             // confusingly, this value is used by fluent for number and date formatting only.
             // we have to implement looking up missing messages in other bundles ourselves.
             let fallback_chain = &[locale];
 
             let mut bundle = FluentBundle::new(fallback_chain);
 
-            bundle
-                .add_resource(resource)
-                .expect("failed to add resource");
+            for resource in resources {
+                bundle
+                    .add_resource(resource)
+                    .expect("failed to add resource");
+            }
             bundles.insert(*locale, bundle);
 
             available.push(*locale);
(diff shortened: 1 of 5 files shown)


djc:
Let's start without logging? We can add it later before we release this thing.

Linking to the Fluent issue on hashmap allocations again: https://github.com/projectfluent/fluent-rs/issues/94. I would argue we should do without TLS for now, we can add it later.

askama_shared/src/i18n_runtime.rs
```diff
@@ -0,0 +1,322 @@
...
+
+use accept_language::parse as accept_language_parse;
+use fluent_bundle::{FluentBundle, FluentResource, FluentValue};
+use std::collections::HashMap;
+
+use super::{Error, Result};
+
+pub use lazy_static::lazy_static;
+
+/// An I18n argument value. Instantiated only by the `{ localize() }` filter.
```
djc: I prefer to have code ordered the other way around: high-level complex dependent API up top, simple private types that don't depend on anything somewhere down the bottom. It's explained a bit more here:

https://deterministic.space/how-to-order-rust-code.html

askama_shared/src/i18n_runtime.rs
```diff
@@ -0,0 +1,322 @@
...
+
+/// Sources; an array mapping Locales to fluent source strings. Instantiated only by the `impl_localize!` macro.
+pub type Sources = &'static [(Locale, &'static str)];
+
+/// Sources that have been parsed. Instantiated only by the `impl_localize!` macro.
+///
+/// This type is initialized in a lazy_static! in impl_localize!,
+/// because FluentBundle can only take FluentResources by reference;
+/// we have to store them somewhere to reference them.
+/// This can go away once https://github.com/projectfluent/fluent-rs/issues/103 lands.
```
djc: So that issue mentions some apparently newly available features/crates -- have you looked into their applicability yet? From looking over the code here, I'm mostly worried about the finer details of fallback chains. (For example, I would say that `en` should prefer `en_US` over `en_AU` in most projects -- though I'm not sure if the Fluent code handles it like this?)

askama_shared/src/i18n_runtime.rs
kazimuth: Huh, I never thought about this. Can do.

askama_shared/src/i18n_runtime.rs
kazimuth: Those crates don't really buy us anything, honestly. `fluent-resmgr` is designed for loading from the filesystem, which we don't do, and `fluent-fallback` does a bunch of unnecessary allocation -- re-building parts of the localization system for every call, which isn't necessary. (It's also only [50 lines of code](https://github.com/projectfluent/fluent-rs/blob/master/fluent-fallback/src/lib.rs), so i don't think it's bad to re-implement it more efficiently.)

...On the other hand, https://crates.io/crates/fluent-locale does exactly what we want here! It wasn't used in the examples before, i didn't realize it existed. Will switch to using it now :)

kazimuth:
Comments addressed 🙂

kazimuth:
Update: I wanna switch the runtime from concat'ing source files to including them separately, one minute...

kazimuth:
Okay, I've just had a thought.

I'm looking through the rest of my i18n code, and it's occurred to me that the runtime + codegen for `impl_localize` aren't actually askama-specific. It's just a system for baking fluent translations into an executable, and then accessing them.

It might make sense to move most of this to its own crate, and add it to askama as a dependency.

djc:
Hah, that sounds like it might make a lot of sense! I'd probably prefer for someone else to maintain that, anyway. 👍

kazimuth:
Closing in favor of https://github.com/kazimuth/baked_fluent/ 🙂

PR #243, opened 2019-05-29
Change askama_escape to require UTF-8 strings

### LINKED ISSUE #242: askama_escape is unsafe.
I *think* this shouldn't be allowed.

```rust
use std::fmt;
use askama_escape::Escaper;

fn main() {
    let s = format!("{}", Test);
    println!("{:?}", String::from_utf8(s.as_bytes().into())); // Guaranteed to always be OK.
}

struct Test;

impl fmt::Display for Test {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        askama_escape::Html.write_escaped_bytes(f, &[0xC3, 0x28])
    }
}
```

quadrupleslap:
This should make `askama_escape` safe, but it changes some parts of the interface, which aren't used by Askama.

Fixes #242.

### DIFF
--- a/askama_escape/src/lib.rs
+++ b/askama_escape/src/lib.rs
@@ -1,5 +1,4 @@
-use std::fmt::{self, Display, Formatter};
-use std::io::{self, prelude::*};
+use std::fmt::{self, Display, Formatter, Write};
 use std::str;
 
 pub struct MarkupDisplay<E, T>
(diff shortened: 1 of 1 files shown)


djc:
Hey, thanks for working on this! In order to make this easier to review, could you please split this in commits where your changes to make the interface safe are not intermixed with unrelated renaming/other refactoring?

quadrupleslap:
Sorry, I don't know what "unrelated renaming" you're referring to — as far as I can tell, all of the renaming is just getting rid of the word `bytes` because I replaced the `&[u8]`s with `&str`s.

djc:
Sorry this has taken me so long, it's been super-busy. But this looks basically solid!

PR #253, opened 2019-07-17
Add support for static function calls

### LINKED ISSUE #196: Static method call
Can I use static method in the template?

Forkk:
This allows static functions to be called from templates using an expression like `self::static_fn(arg)`. This fixes #196.

I haven't tested this outside of the 3 test cases I wrote, so maybe a little more testing should be done before merging this.

### DIFF
--- a/testing/tests/functions.rs
+++ b/testing/tests/functions.rs
@@ -1,0 +1,46 @@
+use askama::Template;
+
+#[derive(Template)]
+#[template(source = "{{ self::simple_test_fn() }}", ext = "txt")]
+struct FnTemplate;
+
+fn simple_test_fn() -> &'static str {
+    "foo"
+}
+
+#[test]
+fn test_fn() {
+    let t = FnTemplate;
+    assert_eq!(t.render().unwrap(), "foo");
+}
+
+
+#[derive(Template)]
+#[template(source = "{{ self::arg_test_fn(arg) }}", ext = "txt")]
+struct FnArgsTemplate{
+    arg: &'static str
+}
+
+fn arg_test_fn(s: &'static str) -> String {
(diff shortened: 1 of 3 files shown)


testing/tests/functions.rs
```diff
@@ -0,0 +1,46 @@
+use askama::Template;
+
+#[derive(Template)]
+#[template(source = "{{ self::simple_test_fn() }}", ext = "txt")]
```
djc: I don't think these examples should need the `self::` prefix, since they should be allowed to just use the neighboring scope from the type impls.

testing/tests/functions.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[test]
+fn test_fn() {
+    let t = FnTemplate;
+    assert_eq!(t.render().unwrap(), "foo");
+}
+
+
+#[derive(Template)]
+#[template(source = "{{ self::arg_test_fn(arg) }}", ext = "txt")]
```
djc: Don't think we need this test, since we're reusing the same argument parsing that's used for other kinds of calls.

testing/tests/functions.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[test]
+fn test_fn_args() {
+    let t = FnArgsTemplate { arg: "foo" };
+    assert_eq!(t.render().unwrap(), "foo!");
+}
+
+
+#[derive(Template)]
+#[template(source = "{% if self::if_test_fn() %}foo{% endif %}", ext = "txt")]
```
djc: I also don't think this test adds much value, since it's just another expression and so quite similar to the first test.

djc:
This has been fixed by #304.

PR #254, opened 2019-07-19
[WIP] Split integrations into separate crates

fabianfreyer:
See discussion in #234. This is a first attempt at splitting out some of the integrations - I'm not sure this is the right way though.

- [x] Split out actix-web
- [ ] Split out gotham
- [ ] Split out rocket
- [ ] Split out iron


djc:
This looks great! I'd like to move the actix tests into askama-actix as well. Also I think as the author of most of the code in askama-actix, I think I should be part of the authors in the crate metadata? Also it looks like askama-actix uses extern crate more than necessary in 2018 Rust.

fabianfreyer:
> This looks great!
Hmm, the tests don't pass yet though.

> I'd like to move the actix tests into askama-actix as well. 
Yep, this sounds like a good idea!

> Also I think as the author of most of the code in askama-actix, I think I should be part of the authors in the crate metadata?
Of course! I hadn't done much in the Cargo.toml yet - just poking around to make sure this approach works for now.

> Also it looks like askama-actix uses extern crate more than necessary in 2018 Rust.
Ah, good point, I'll remove that

djc:
This was finally fixed in #255. Thanks for working on it with me!

PR #266, opened 2019-10-08
Add support for dereference operator

davebrent:
The use case I had for this, was to be able to do something like the following (the same applies to #265):

```
{% macro foo(b) %}
  {% if *b %}
    X
  {% else %}
    Y
  {% endif %}
{% endmacro %}

{% call foo(true) %}
{% call foo(false) %}
```

Which seemed necessary, because of arguments in calls being referenced [here](https://github.com/djc/askama/blob/66274b7c8ef2f47fe479013b9d408ac21a36472d/askama_derive/src/generator.rs#L605)

### DIFF
--- a/askama_derive/src/parser.rs
+++ b/askama_derive/src/parser.rs
@@ -472,7 +472,7 @@ named!(expr_filtered<Input, Expr>, do_parse!(
 ));
 
 named!(expr_unary<Input, Expr>, do_parse!(
-    op: opt!(alt!(tag!("!") | tag!("-"))) >>
+    op: opt!(alt!(tag!("!") | tag!("-") | tag!("*"))) >>
     expr: expr_filtered >>
     (match op {
         Some(op) => Expr::Unary(str::from_utf8(op.0).unwrap(), Box::new(expr)),


djc:
So I've always stuck to not allowing `&` and `*` in templates, and would prefer to keep it that way. Is there some workaround for this?

davebrent:
I'm not sure, in my case, I've worked around it by just removing the condition from my macro (which was probably for the best).

An alternative that did cross my mind, if [the reference](https://github.com/djc/askama/blob/66274b7c8ef2f47fe479013b9d408ac21a36472d/askama_derive/src/generator.rs#L605) was just removed and require arguments passed into macros to be explicitly cloned if the type does not implement `Copy`. So that both situations have a way of working (I think?):

```diff
diff --git a/askama_derive/src/generator.rs b/askama_derive/src/generator.rs
index 7757016..a63adee 100644
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -602,7 +602,7 @@ impl<'a> Generator<'a> {
                 args.get(i)
                     .unwrap_or_else(|| panic!("macro '{}' takes more than {} arguments", name, i)),
             );
-            buf.writeln(&format!("let {} = &{};", arg, expr_code));
+            buf.writeln(&format!("let {} = {};", arg, expr_code));
             self.locals.insert(arg);
         }

diff --git a/testing/templates/deep-kid.html b/testing/templates/deep-kid.html
index a2aa105..bfb533e 100644
--- a/testing/templates/deep-kid.html
+++ b/testing/templates/deep-kid.html
@@ -6,5 +6,5 @@
 {% endblock %}

 {% block content %}
-  {% call libk::thrice(item) %}
+  {% call libk::thrice(item.clone()) %}
 {% endblock %}
diff --git a/testing/templates/macro-if.html b/testing/templates/macro-if.html
new file mode 100644
index 0000000..4760e84
--- /dev/null
+++ b/testing/templates/macro-if.html
@@ -0,0 +1,4 @@
+{%- macro foo(a) -%}
+  {% if a == 1 %}{{ a }}{% endif %}
+{%- endmacro -%}
+{% call foo(1) %}
diff --git a/testing/tests/macro.rs b/testing/tests/macro.rs
index 51fbcdc..a743748 100644
--- a/testing/tests/macro.rs
+++ b/testing/tests/macro.rs
@@ -12,6 +12,16 @@ fn test_macro() {
     assert_eq!(t.render().unwrap(), "12foo foo foo3");
 }

+#[derive(Template)]
+#[template(path = "macro-if.html")]
+struct MacroIfTemplate {}
+
+#[test]
+fn test_macro_if() {
+    let t = MacroIfTemplate {};
+    assert_eq!(t.render().unwrap(), "1");
+}
+
 #[derive(Template)]
 #[template(path = "import.html")]
 struct ImportTemplate<'a> {
```

But I appreciate if you would rather leave this alone and close the PR (its not longer an issue for me).

djc:
Thanks for the response!

PR #269, opened 2019-10-28
added TemplateDyn trait that can be made into a trait object

jbg:
```
#[derive(Template)]
struct MyTemplate {}

let tpl = MyTemplate {};
(&tpl as &TemplateDyn).render();
```


djc:
Merged #270 instead.

PR #270, opened 2019-10-28
Moved no-receiver size_hint method to a separate trait

jbg:


djc:
In this one, why isn't `extension()` causing problems? Is it because of the `Sized` bound?

jbg:
Yes. I could move that into the other trait too (and rename the new trait) if desired.

jbg:
(and in that case the Sized bound would no longer be needed)

djc:
So as an alternative, could we add a `Sized` bound for `size_hint()`, as well?

jbg:
It works, but you have to also add a `Self: Sized` bound to `render()`. PR: #271

askama/src/lib.rs
```diff
@@ -502,6 +502,10 @@ pub trait Template {
     fn extension() -> Option<&'static str>
     where
         Self: Sized;
+    fn dyn_size_hint(&self) -> usize;
+}
+
+pub trait TemplateSizeHint {
```
djc: Let's rename this to `SizedTemplate` and move both `extension()` and `size_hint()` into it?

askama/src/lib.rs
```diff
@@ -502,6 +502,10 @@ pub trait Template {
     fn extension() -> Option<&'static str>
     where
         Self: Sized;
+    fn dyn_size_hint(&self) -> usize;
```
djc: I would say maybe we can just name this `size_hint()` for now? Not sure if the conflict causes issues in the common case.

jbg: If they have the same name then importing `SizedTemplate` and trying to use the receiverless `size_hint()` will give a "multiple applicable items in scope" compile error if `Template` is also imported.

djc: Yeah, I think I'm okay with that.

jbg: Doesn't it basically make the receiverless `size_hint()` unusable? Or is there a way to disambiguate?

(Giving `extension()` the same treatment causes the same compile error in `testing/tests/simple.rs`.)

djc: I think you can do something like `SizedTemplate::extension(&t)` or `<t as SizedTemplate>::extension()`.

jbg: ah yes, it works. I've updated the PR.

djc:
Thanks for sticking with it and working through this with me!

djc:
(Also, I'd be curious to hear a little bit more about what you're using Askama for.)

PR #271, opened 2019-10-28
add Self: Sized bound to size_hint() and render()

jbg:


jbg:
The problem with this approach is that you can no longer use `render()` on a trait object due to the `Sized` constraint; you must use `render_into()` which doesn't need the constraint due to not using `size_hint()`.

djc:
Merged #270 instead.

Issue #275, opened 2019-12-04
Use "and", "or", "not" logical operators


bibicron:
Thank you for this great project.

Is it possible to add the "and", "or", "not" logical operators in addition to "&&", "||", "!" in the IF statements?

This will make the templates more jinja2 compatible.
I am was thinking to use the same template files with Askama for SSR and Nunjucks for CSR.
Nunjucks will not render if will find the "!" operator. The workaround for now is to place the code in the ELSE block.

Thank you

djc:
I think I would be open to this; we can extend the custom syntax support that's already in place to allow changing these operators.

bibicron:
Sound good.
Other updates which will make the code more jinja2 compatible are:
1. Call macros as an expression also
- Jinja2 can call a macro as a block: {%call foo()%}...{%endcall%}
- Jinja2 can call a macro as an expression: {{foo()}}
- Askama can call a macro like this: {%call foo()%}
2. Declare variable with "set"
- Jinja2 declares mutable variable as: {%set foo=bar%}
- Askama declares immutable variable as: {%let foo=bar%}

I can start with the "set" declaration as I already implement it on my local copy.
All these updates will be 100% compatible with the previous syntax.

PR #278, opened 2019-12-12
Added the "set" variable declaration

bibicron:
Added the "set" variable declaration, in addition to the "let" variable declaration.
This will be a mutable variable declaration.
This will make the syntax more jinja2 compatible.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -327,6 +327,12 @@ impl<'a> Generator<'a> {
                 Node::Let(ws, ref var, ref val) => {
                     self.write_let(buf, ws, var, val);
                 }
+                Node::SetDecl(ws, ref var) => {
+                    self.write_set_decl(buf, ws, var);
+                }
+                Node::Set(ws, ref var, ref val) => {
+                    self.write_set(buf, ws, var, val);
+                }
                 Node::Cond(ref conds, ws) => {
                     self.write_cond(ctx, buf, conds, ws);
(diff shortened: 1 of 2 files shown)


bibicron:
Actually I think that my implementation is not optimal.
I think that it is better to combine:
- block_let & block_set into something like block_let_set
- Let() & Set() into LetSet()
- LetDecl() & SetDecl into LetSetDecl()
- write_let() & write_set() into write_let_set()
- write_let_decl() & write_set_decl() into write_let_set_decl()

All these pairs of functions are almost identical and clutter the code. So it is better to combine them.

The only place where I am not sure how to implement this is:
In block_let() from parser.rs, how should I change the "ws(tag("let"))," code to accept "let" or "set"?
Will "ws(alt(tag("let"), tag("set")))," work for that situation?

If this newer approach is fine with you please let me know your preferred names for the new combined functions.

djc:
Actually, as mentioned in #275 I'd like to do this through a named custom syntax. This involves extending the `Syntax` type as defined in `askama_shared`, and passing an instance of `Syntax` down through the parsers to where you want to use it. There should be a `Syntax::jinja()` method that creates the Jinja-like syntax and then you'd have to wire this up to the syntax selection configuration bits.

bibicron:
My bad, I didn't catch that. Probably I went over my head here.
It is a great idea the Syntax part. When implemented it could parse handlebars syntax.
I will close this pull request for now.

PR #286, opened 2020-01-11
upgrade dependencies(actix-web 0.7 -> 2, bytes 0.4 -> 0.5)

### LINKED ISSUE #246: Support actix-web v1.0
The `feature = ["with-actix-web"]` functionality currently depends on actix-web v0.7.0, and won't work with v1.0.

I tried updating the dependency locally, and it looks like the interface of actix-web's `Responder` trait changed in 1.0. ([v1.0] vs. [v0.7])

[v1.0]: https://docs.rs/actix-web/1.0.0/actix_web/trait.Responder.html
[v0.7]: https://docs.rs/actix-web/0.7.19/actix_web/trait.Responder.html

DCjanus:
Close #246

### DIFF
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -23,7 +23,7 @@ serde-json = ["askama_shared/serde_json"]
 serde-yaml = ["askama_shared/serde_yaml"]
 with-iron = ["iron", "askama_derive/iron"]
 with-rocket = ["rocket", "askama_derive/rocket"]
-with-actix-web = ["actix-web", "askama_derive/actix-web", "mime", "mime_guess", "bytes"]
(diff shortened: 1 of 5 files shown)


DCjanus:
I've tried to upgrade `gotham`, met a lot of problem, looks like gotham is no longer in active maintenance phase. Maybe we should remove `gotham` support?

djc:
Awesome, thanks!

djc:
I think the split cannot work or is quite tricky. I'd like to get the upgrade in anyway!

DCjanus:
You are right, I've tried and found I don't know how to resolve that. Reopen this.

DCjanus:
BTW, maybe we should remove `gotham` support, looks like it is no longer in active maintenance.

djc:
As for Gotham, looks like their latest commit to master was 10 days ago - that hardly seems dead to me.

PR #287, opened 2020-01-11
Support escaping in string literals

siiptuo:
Do not attempt to parse escape sequences thoroughly. Instead let the
Rust compiler to check the string literals and provide nice error
messages if necessary.


djc:
Nice, thanks!

PR #290, opened 2020-01-23
Add support for warp

aeons:
Add support for Warp's `Reply` trait in template derivations.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -89,6 +89,20 @@ jobs:
           cargo test --package askama_rocket --all-targets
           cargo clippy --package askama_rocket --all-targets -- -D warnings
 
+  Warp:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          override: true
+          components: clippy
+      - run: |
+          cargo test --package askama_warp --all-targets
+          cargo clippy --package askama_warp --all-targets -- -D warnings
+
   Lint:
     runs-on: ubuntu-latest
     steps:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,8 +8,10 @@ members = [
     "askama_iron",
     "askama_rocket",
     "askama_shared",
+    "askama_warp",
     "testing",
 ]
+
 default-members = [
     "askama",
     "askama_derive",
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -28,6 +28,7 @@ with-iron = ["askama_derive/iron"]
 with-rocket = ["askama_derive/rocket"]
 with-actix-web = ["askama_derive/actix-web"]
 with-gotham = ["askama_derive/gotham"]
+with-warp = ["askama_derive/warp"]
 
 [dependencies]
 askama_derive = { version = "0.9.0", path = "../askama_derive" }
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -471,6 +471,13 @@
 //! signature, with a status code of `500 Internal Server Error`, mime `*/*`, and an empty `Body`.
 //! This preserves the response chain if any custom error handling needs to occur.
 //!
+//! ## Warp integration
+//!
+//! Enabling the `with-warp` feature appends an implementation of Warp's `Reply`
+//! trait for each template type. This makes it simple to return a template from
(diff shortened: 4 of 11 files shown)


djc:
Hey, sorry for the slowish response to this!

I've been running into trouble from having all these web framework integrations included in my template engine crate (even if optionally). This causes issues because the web framework's dependencies leak into my dep graph, causing problems if conflicting dependencies are used by the web frameworks. So, while I'm open to hosting warp integrations in this repository, I think we'd need a different approach to integrations altogether. See more discussion in #234; my attempts so far have run aground on coherence issues.

Having the integration included in the web framework might be a more attractive solution, but I'm not sure if Sean et al would be open to that (do they already ship some templating integrations?). Alternatively, I need a sustainable way to do the integrations that will let me keep integrations up to date without running into dependency conflicts going forward.

aeons:
That makes sense.

This feels like a place where orphan instances would be nice to have.

There are currently no templating integrations in the warp repo, but I can check if they would be interested.

I can't really tell whether the separate crate approach worked out or not? Is that something I should pursue for this?

djc:
If you want to help figuring out, that would be awesome. Currently I'm thinking that it might be better to have per-integration crates that have their own `Template: askama::Template` traits. Those integration crates could then also contain a blanket implementation that implements `Reply` (or equivalent) for traits that implement these. Then there should be some feature flag in the derive crate (and exposed through the API crate) that makes the code generator implement that extra trait.

aeons:
I haven't updated the documentation, and the mime stuff is just copy/pasted, but this seems to work.

djc:
This is great, thanks! I'd be happy to merge this once you fix up the documentation. Also any generic mime stuff can still live in `askama` if you like, that shouldn't be too big of an issue.

djc:
Also, if possible, it would be nice to have the tests in the askama_warp crate.

aeons:
Alright, I'll make the get_mime_type a `pub fn` then.
Looking at tests now.

aeons:
Right now `cargo test` works in both `askama_warp` and `testing`, but when I run `cargo test` in the workspace, it enables the `warp` feature in `askama_derive` when running the tests in `testing`, which makes the compilation fail.

I can't figure out how to not enable that feature when running the workspace tests.

aeons:
Helpfully, I found this issue https://github.com/rust-lang/cargo/issues/6195
:laughing:

djc:
Recognizing that this *should* never come up in the wild, maybe there's a hack that could work where we add a non-additive feature to `askama_derive` that negates the warp integration. The testing crate would then depend on that feature. How do you feel about that?

aeons:
Something like this?
```rust
if cfg!(feature = "warp") && !cfg!(feature = "test") {
  self.impl_warp_reply(&mut buf);
}
```

djc:
Yes, something like this. I'd like to name the feature `no-integrations` though, in anticipation of using it for the other integrations in the same way.

aeons:
That fixes `testing`, but now `askama_warp` fails since `warp::reply::Reply` is not implemented. I think it relates to the same problem you had in the linked issue.

djc:
Have been thinking about this some more. What if we change the workspace's `default-members` to exclude `askama_warp` (and, in the future, other integration crates)? Can we then run `cargo t --workspace` and `cargo t -p askama_warp` successfully in separate invocations?

aeons:
That seems to work, if you just do `cargo test` in the repo root.
`cargo test --workspace` fails with the same errors though.

aeons:
I removed the `--workspace` from both build and test commands.

From `cargo help build`:
> All packages in the workspace are built if the `--workspace` flag is supplied. The
`--workspace` flag is automatically assumed for a virtual manifest.
Note that `--exclude` has to be specified in conjunction with the `--workspace` flag.

It looks like `--workspace` makes it ignore the `default-members`.

djc:
So I've redone all the other integrations as separate crates now. This has caused some churn, but hopefully rebasing your work should be straightforward -- and it provides a clear model on how to add more integrations. Thanks for working on this with me!

djc:
(Also, please squash all your commits on this branch.)

aeons:
This ended out nicely I think. Thanks.

aeons:
The readme still needs to be updated I guess.

djc:
Thanks! I'll make sure to update the documentation before the next release. Are you comfortable with using the wrap integration from git, or would you prefer to have a release soonish?

aeons:
No rush, I'm just playing around with it for now.

PR #296, opened 2020-02-07
fix truncate filter: eliminate panic on len not lie on boundary

danloh:
check boundary, add truncate len per 1 as needed till checking pass


danloh:
also can set a step(1 or -1), but it is a breaking change

djc:
Thanks! Would you mind adding a test for this? It can be a unit test like the ones already in the same file.

PR #302, opened 2020-02-24
Fix askama_shared clippy warning

paolobarbolini:
This fixes the clippy warning I encountered in #301


djc:
Thanks for fixing this!

PR #304, opened 2020-03-15
Support function calls

### LINKED ISSUE #196: Static method call

### LINKED ISSUE #285: Call methods on root context
Is there a syntax for calling methods implemented directly on the 'context'? A bare `function()` doesn't appear to do the trick

Senzaki:
This allows function calls in templates. There are multiple ways a function could be called: static callables (e.g. static functions), methods on template (in an impl block) and function pointer calls (template has an attribute of type e.g. FnOnce(args) -> String).

I tried to be consistent with the ways things are right now: i.e. {{ var }} refers to an attribute of the template, and path::CONST refers to a constant declared in the path module.

Case 1 : `callable` is an attribute of `Template`.

```rust
#[derive(Template)]
#[template(source = "{{ callable(123) }}", ext = "txt")]
struct Template {
    callable: fn(u32) -> String,
}
```

Case 2: `callable` is in the outer context (e.g. static function)

```rust
fn callable(val: u32) -> String {
    format!("{}", val)
}

#[derive(Template)]
#[template(source = "{{ self::callable(123) }}", ext = "txt")]
struct Template;
```

Case 3: `callable` is a method:
Actually, methods are just static functions that take `&self` as an argument.

```rust
#[derive(Template)]
#[template(source = "{{ Self::callable(self, 123) }}", ext = "txt")]
struct Template;

impl Template {
    fn callable(&self, val: u32) -> String {
        format!("{}", val)
    }
}
```

This last case might seem weird at first, but it actually keeps things consistent in the mind of the programmer, i.e. {{ var }} refers to an attribute of the template and {{ path::var }} refers to something else.

I also need to update the documentation to explain this simply.

Fixes #285, fixes #196

EDIT: Removed reference to an issue that this does not actueally address.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -1030,7 +1030,10 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             }
 
             let scoped = match *arg {
-                Expr::Filter(_, _) | Expr::MethodCall(_, _, _) => true,
+                Expr::Filter(_, _)
+                | Expr::MethodCall(_, _, _)
+                | Expr::VarCall(_, _)
+                | Expr::PathCall(_, _) => true,
                 _ => false,
             };
 
@@ -1169,13 +1172,8 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             }
             buf.write(part);
         }
-        buf.write("(");
-        for (i, arg) in args.iter().enumerate() {
-            if i > 0 {
-                buf.write(",");
-            }
-            self.visit_expr(buf, arg);
-        }
+        buf.write("(&");
+        self._visit_args(buf, args);
         buf.write(")");
         DisplayWrap::Unwrapped
     }
@@ -1198,13 +1196,8 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             buf.write("self.");
             buf.write(s);
         }
-        buf.write(")(");
-        for (i, arg) in args.iter().enumerate() {
-            if i > 0 {
-                buf.write(",");
-            }
-            self.visit_expr(buf, arg);
-        }
+        buf.write(")(&");
+        self._visit_args(buf, args);
         buf.write(")");
         DisplayWrap::Unwrapped
     }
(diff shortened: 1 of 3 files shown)


askama_shared/src/generator.rs
```diff
@@ -1160,6 +1162,24 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
 
+    fn visit_path_call(&mut self, buf: &mut Buffer, path: &[&str], args: &[Expr]) -> DisplayWrap {
+        for (i, part) in path.iter().enumerate() {
+            if i > 0 {
+                buf.write("::");
+            }
+            buf.write(part);
+        }
+        buf.write("(");
+        for (i, arg) in args.iter().enumerate() {
```
djc: Is there any reason you didn't use the `_visit_args()` helper here?

askama_shared/src/generator.rs
```diff
@@ -1170,6 +1190,25 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+    fn visit_var_call(&mut self, buf: &mut Buffer, s: &str, args: &[Expr]) -> DisplayWrap {
+        buf.write("(");
+        if self.locals.contains(s) || s == "self" {
+            buf.write(s);
+        } else {
+            buf.write("self.");
+            buf.write(s);
+        }
+        buf.write(")(");
+        for (i, arg) in args.iter().enumerate() {
```
djc: ... and here?

djc:
You can just silence the clippy warnings by attaching `allow()` attributes to the problematic functions. Thanks!

Senzaki:
There were no good reason for me not to use _visit_args indeed! However, I could not figure out why method calls needed to be scoped in _visit_args (e.g. &{ self.a.call() } instead of &self.a.call()), but I suppose if this had to be done for method calls, I should add it for function calls as well?
Anyway, this should work, and thanks for the awesome crate!

djc:
> However, I could not figure out why method calls needed to be scoped in _visit_args (e.g. &{ self.a.call() } instead of &self.a.call()), but I suppose if this had to be done for method calls, I should add it for function calls as well?

I think so too! I don't remember off the top of my head, but I think some kinds of expressions will introduce temporary variables, and the scopes are used to make sure those are cleaned up again.

Thanks for your contribution!

PR #307, opened 2020-03-31
fix: make filesizeformat work with all types that implement humansize::FileSize

DusterTheFirst:
~~I have enabled dynamic dispatch on the filesizeformat filter's parameter as to allow for numbers other than usize to work~~

As per my commit 1:
dynamic dyspatch would not work since file_size takes a generic.
Instead we use a generic function with type B being any type that implements FileSize.

### DIFF
--- a/askama_shared/src/filters/mod.rs
+++ b/askama_shared/src/filters/mod.rs
@@ -125,7 +125,7 @@ where
 
 #[cfg(feature = "humansize")]
 /// Returns adequate string representation (in KB, ..) of number of bytes
-pub fn filesizeformat(b: &dyn humansize::FileSize) -> Result<String> {
+pub fn filesizeformat<B: FileSize>(b: B) -> Result<String> {
     b.file_size(file_size_opts::DECIMAL)
         .map_err(|_| Fmt(fmt::Error))
 }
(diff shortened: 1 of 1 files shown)


DusterTheFirst:
As a side note, is there any reason for preferring dynamic dispatch with filters like `urlencode` over generics? I know that there is a small performance decrese from using dynamic dispatch over generics but it also saves code size. Or is there no real reason for your choice?

djc:
I don't think the use of dynamic dispatch for the filters is particularly thought through. Both seem okay options, and which one is preferable might depend on the breadth of relevant types that implement the trait (for code size reasons). Also the list of filters has often grown through outside contributors, and as I don't have a strong opinion I probably merge both kinds without paying much attention.

DusterTheFirst:
Thanks for the clarification. I agree that there is really no better answer. Dynamic dispatch does help with code size as you had said and I agree thourpughly with your choice to use it on such widely used traits such as the filters that take a Display type. For the file size format filter I feel that, since there are only a few different types that have the trait implemented for, it is safe for code size, and worth the performance benefit to use a generic. 

I just also want to thank your for your work on this project, it is by far my favorite template by system since it is able to be checked at compile time and compiled into the binary without any extra hassle.

djc:
Thanks for the kind words! I'd like to hear more about your use case.

Issue #318, opened 2020-05-15
Do not call `write` with empty strings


cipriancraciun:
While debugging the efficient sub-template rendering I've stumbled upon the following strange behaviour:  the `write` method of `fmt::Write` is called multiple times with an empty string.

Granted this has a minor performance impact especially if buffered writers are used;  however the behaviour is strange.

Bellow the following is the template:
~~~~
{% for index in 0..100 -%}
	{{- "abc" -}}
{%- endfor %}
~~~~

And the resulting code:
~~~~
impl ::askama::Template for TestTemplate {
    fn render_into(&self, writer: &mut ::std::fmt::Write) -> ::askama::Result<()> {
        include_bytes !
("/tmp/test.html")
;
        for (index, _loop_item) in ::askama::helpers::TemplateLoop::new(0..100) {
            write!(
                writer,
                "{expr0}",
                expr0 = &::askama::MarkupDisplay::new_unsafe(&"abc", ::askama::Html),
            )?;
        }
        Ok(())
    }
    fn extension(&self) -> Option<&'static str> {
        Some("html")
    }
    fn size_hint(&self) -> usize {
        0
    }
}
impl ::askama::SizedTemplate for TestTemplate {
    fn size_hint() -> usize {
        0
    }
    fn extension() -> Option<&'static str> {
        Some("html")
    }
}
impl ::std::fmt::Display for TestTemplate {
    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
        ::askama::Template::render_into(self, f).map_err(|_| ::std::fmt::Error {})
    }
}
~~~~

And bellow is the call trace to a custom `fmt::Write`:
~~~~
struct Writer {}

impl fmt::Write for Writer {
    fn write_str (&mut self, _string : &str) -> (Result<(), fmt::Error>) {
        eprintln! ("[>>] {} | {}", _string.len (), _string);
        return Ok (());
    }
}
~~~~
~~~~
[>>] 0 |
[>>] 0 |
[>>] 3 | abc
[>>] 0 | 
[>>] 0 | 
[>>] 3 | abc
[>>] 0 | 
[>>] 0 | 
[>>] 3 | abc
[>>] 0 | 
[>>] 0 | 
[>>] 3 | abc
...
~~~~

As can be seen, there are many calls with an empty string.

djc:
[...]

PR #325, opened 2020-05-24
Update `EscapeWriter` HTML implementation to not output empty strings

cipriancraciun:
As discussed in #318


djc:
Thanks, this is great!

Issue #315, opened 2020-05-15
Ignore whitespace between method name and argument list


cipriancraciun:
The following simple template fails:
~~~~
{{ model.test ("test-from-model") }}
~~~~

Just by removing the whitespace between the `test` and `(...` solves the issue.

djc:
Yeah, I guess. I am unlikely to work on this anytime soon, but if you want to take a crack at implementing I could provide you with some pointers.

cipriancraciun:
OK, could you give me the needed pointers?

(Although I agree it's low priority, given that my coding styles aren't inline with the 99%;  however this weekend perhaps I'll take a look.)

djc:
You'll need to update https://github.com/djc/askama/blob/master/askama_shared/src/parser.rs#L307 (and probably also `expr_path_call()` a few lines down) to eat the extra whitespace.

cipriancraciun:
OK, trying to solve this issue, I've came across the following ambiguity in the language regarding filter application and bitwise-or.

Say I have the following two templates `a|b` and `a | b`:
* the first one `a|b` is interpreted as applying the `b` filter to `a`, when used fails complaining that the filter `b` does not exist;
* the second one `a | b` is interpreted as a bitwise operator between `a` and `b`;

For example in the `NestedFilterTemplate` `x|mytrim|safe` if rewritten as `x | mytrim | safe` fails by saying that `safe` is not a member of the template.

PR #327, opened 2020-05-25
Add support for more whitespace positions within expressions

cipriancraciun:
As reported in #315 at the moment whitespaces aren't allowed in many places where normal Rust would allow them.

This pull request adds support for whitespaces in:
* in function calls: `x ( 1 , 2 )`;
* in path elements: `module :: element`;
* in attributes: `x . y . z`;
* in filter arguments: `x| filter ( 1 , 2 )`;
* before unary operators: `! false` and `- 42`;
* also allow more than a single whitespace;

Also change some tests to include whitespaces in various positions.

As noted in #315 at the moment there is a language ambiguity regarding bitwise or and filter application.  As such at the moment the pipe character `|` should not be preceded by a space when wanting filter application, and must always be preceded when wanting bitwise or.  (This is not a deficiency of the current patch.)

### DIFF
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -113,11 +113,11 @@ where
     E: ParseError<I>,
 {
     move |i: I| {
-        let i = alt::<_, _, (), _>((tag(b" "), tag(b"\t")))(i.clone())
+        let i = many1(alt::<_, _, (), _>((tag(b" "), tag(b"\t"))))(i.clone())
             .map(|(i, _)| i)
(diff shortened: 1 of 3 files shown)


djc:
Similar to #323, I'm okay with the code generator changes but let's only have a specific test that tests some subset of the code generator for whitespace sensitivity.

djc:
Merged as e71e6806496027089b5c3f0dd764bebe9abbf735 and af2fd0b8dc0894fdb96656351bf2015d4df73813. Thanks!

PR #336, opened 2020-06-02
Allow block definitions in if/for/match blocks (fixes #335)

### LINKED ISSUE #335: Blocks nested within {% if %} blocks are not allowed
Nesting blocks within conditional (and possibly looping) blocks results in the error: `blocks ('nested_block_name') are only allowed at the top level of a template or another block`. Further discussion in #85.

djc:


djc:
@jmafc do you want to try this out?

jmafc:
Do I just have to reference the master branch in my Cargo.toml to do that?

djc:
Well, the `deep-blocks` branch for now, but yes, that should do it.

jmafc:
Ok, I tried it, but it still causes the message: `blocks ('meta_descrip') are only allowed at the top level of a template or another block`. The `meta_descrip` is the what `meta_x` stood for in previous examples, and as shown in the example, the block is nested within an `{% else %}`. I'm pretty I got your code because the Cargo.lock shows `source = "git+https://github.com/djc/askama?branch=deep-blocks#79374b96812dc8454b28bccd05757d90e87b104f"` which is your commit.
Also, in order to test I had to disable a call to a macro, because it was causing `message: unable to parse template`. The macro was parsing and processing OK when building with askama 0.8.0. The macro has a for loop to output HTML list items and includes an `if !loop.last` block.
Let me know if you need further details.

djc:
Ah yes, I forgot to strip out the code in the code generator that panics if it encounters a nested block definition. That's fixed with the latest push.

Can you file a separate issue about the macro thing? That seems unrelated, and potentially a regression. Would like have a minimal reproduction of the template syntax for that.

jmafc:
Ok, the `blocks ('meta_descrip') are only allowed at the top level of a template or another block` is no longer reported. I'll file the issue about the macro regression later today.

djc:
And does the template with the if-nested block produce the expected output? :)

jmafc:
Yes it did produce the expected output (eventually, because one of my templates had an unrecognized field, but I've fixed that now).

djc:
Cool, thanks for checking.

PR #344, opened 2020-07-02
Disable default features of `askama_shared` in `askama_derive`

JohnTitor:
3b8bf97cb6da128f020bb557057269661ac89fea disabled default features of `askama_shared` in `askama` but it didn't in `askama_derive`. We'd drop dependencies completely if we could disable there also.


djc:
Ah, thanks! I've pushed out 0.10.1 with this fix.

PR #345, opened 2020-07-08
Add format filter that swaps the first two arguments.

couchand:
Swapping the first two arguments allows a more natural filter usage: `{{ val | fmt("{:?}") }}`.

Happy to bikeshed the name.

### DIFF
--- a/askama_shared/src/filters/mod.rs
+++ b/askama_shared/src/filters/mod.rs
@@ -58,13 +58,14 @@ const ENCODE_SET: &AsciiSet = &NON_ALPHANUMERIC
 // Askama or should refer to a local `filters` module. It should contain all the
 // filters shipped with Askama, even the optional ones (since optional inclusion
 // in the const vector based on features seems impossible right now).
-pub const BUILT_IN_FILTERS: [&str; 24] = [
+pub const BUILT_IN_FILTERS: [&str; 25] = [
     "abs",
     "capitalize",
     "center",
     "e",
     "escape",
     "filesizeformat",
+    "fmt",
     "format",
     "indent",
     "into_f64",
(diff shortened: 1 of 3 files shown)


djc:
I don't see this adding much value -- it seems more likely to be confusing. Sorry!

couchand:
If you don't want to add this, that's fine, but I hope you can at least take a moment to consider how this API is significantly more ergonomic.

For one thing, it better matches the mental model of other filters: data flows in from the left, configuration is passed as a param (see, for instance, `join`). It is also superior for composition with other filters, such as in  `{{ num | to_f64 | fmt("{:.2}") }}`.

djc:
Should have led with that! I hadn't realized the composition advantage. How do you feel about restricting it to a single argument (so no more arguments after the format string)?

couchand:
Limiting it to one argument makes sense, it would get awfully confusing with more than one.

For that matter, I'm not sure there's much point to putting any static text in the format string in the context of a templating language. Would it be crazy to automatically wrap the format in curly braces? 

`{{ num | fmt(":.2") }}`

On the other hand, perhaps there would be a reason to include static text, to support composition with downstream filters:

`{{ price | fmt("${:.2}") | center }}`

Though that suggests a use-case for additional parameters (most useful if named) :

`{{ price | fmt("{symbol}{:.2}", symbol=currency) | center }}`

Although at that point you're probably better off writing a custom filter.

djc:
One of the core design choices in Askama is to make the templating language Rust-like. Eliding the braces seems antithetical to that.

Let's start with the simple two-argument version in this PR and we can discuss extending it once we run into a compelling use case.

couchand:
Updated.  What do you think about the name?  I'd prefer to be explicit about the difference.  Perhaps `format_with` or `format_as` would clarify it?

As an aside, have you considered using `trybuild` to test compile fails?

testing/tests/filters.rs
```diff
@@ -38,6 +38,18 @@ fn filter_format() {
     assert_eq!(t.render().unwrap(), "\"formatted\"");
 }
 
+#[derive(Template)]
+#[template(path = "fmt.html", escape = "none")]
```
djc, line 42: Please inline the source here (so `source = "{{ var|fmt(\"{:?}\") }}"` rather than `path = `).

askama_shared/src/filters/mod.rs
```diff
@@ -137,12 +138,28 @@ pub fn urlencode(s: &dyn fmt::Display) -> Result<String> {
     Ok(utf8_percent_encode(&s, ENCODE_SET).to_string())
 }
 
+/// Formats arguments according to the specified format
+///
+/// The *second* argument to this filter must be a string literal (as in normal
+/// Rust). All arguments are passed through to the `format!()`
```
djc, line 144: This should be updated now that it can only take 2 arguments.

djc:
Thanks for your contribution, much appreciated (especially for sticking with it despite the initial negative feedback)!

couchand:
Totally understand, I should have provided more context from the start.  Thanks!

PR #347, opened 2020-07-12
Add tide integration

jbr:
Hi, and thanks for Askama! I'm excited to use this with the [tide](https://github.com/http-rs/tide) web framework, so I wrote a little integration.

I had a question about the fallibility of `render()` (not render_into) — I tried to figure out how to write a test of the Err variant, and I couldn't figure out what the failure conditions would be, as it's a `std::fmt::Write::write_fmt` into a String. Is it fairly safe to just unwrap the Result returned by render? If so, that would let us use `Into<Body>` which would be preferable. If not, how can I provide inputs that fail in order to test both paths?

Thanks again, and let me know if there are changes I can make to improve this PR!


djc:
Thanks for doing the work on this, will be happy to accept this!

`write_fmt()` just returns `fmt::Result`, which by definition doesn't have any meaning attached to it. If you want to test the failure path, you could probably write a custom `Write` implementation that errors under some particular condition. I think unwrapping is probably bad in this case.

README.md
```diff
@@ -30,7 +30,7 @@ in a for-profit context, please consider supporting my open source work on
 * Construct templates using a familiar, easy-to-use syntax
 * Template code is compiled into your crate for [optimal performance][benchmarks]
 * Benefit from the safety provided by Rust's type system
-* Optional built-in support for Actix, Gotham, Iron, Rocket and warp web frameworks
+* Optional built-in support for Actix, Gotham, Iron, Rocket, warp, and tide web frameworks
```
djc, line 33: Nit: I'd like to keep all the frameworks in alphabetic order (where they already are). This happens in a bunch of other places too, can you please review them?

askama_tide/LICENSE-APACHE
```diff
@@ -0,0 +1,201 @@
+                              Apache License
```
djc, line 1: The license files should be symlinks, as in the other integration crates.

askama_tide/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition = "2018"
+description = "Tide integration for Askama templates"
+keywords = ["markup", "template", "jinja2", "html", "tide", "http-types"]
+homepage = "https://github.com/djc/askama"
+repository = "https://github.com/djc/askama"
+documentation = "https://docs.rs/askama"
+license = "MIT OR Apache-2.0"
+workspace = ".."
+readme = "README.md"
+
```
djc, line 14: Nit: remove one of the empty lines here, please.

askama_shared/src/generator.rs
```diff
@@ -298,6 +301,28 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+        let ext = self
+            .input
+            .path
+            .extension()
+            .and_then(|s| s.to_str())
+            .unwrap_or("txt");
+
+        self.write_header(buf, "std::convert::TryInto<::askama_tide::Body>", None);
+        buf.writeln(
+            "type Error = ::askama::Error;\n\
```
djc, line 314: This should probably import `Error` and `Result` from `askama_tide` rather than `askama` directly (I know this is potentially inconsistent with what's currently there for other integrations, but I think it's the right direction for the future).

book/src/integrations.md
```diff
@@ -50,3 +50,12 @@ Enabling the `with-warp` feature appends an implementation of Warp's `Reply`
...
 a Warp filter. See [the example](https://github.com/djc/askama/blob/master/askama_warp/tests/warp.rs)
 from the Askama test suite for more on how to integrate.
+
+## Tide integration
+
+Enabling the `with-tide` feature appends `Into<tide::Response>` and
+`TryInto<tide::Body>` implementations for each template type. This
+provides the abiltity for tide apps to build a response directly from
+a template, or to append a templated body to an already-existing
+Response. See [the example](https://github.com/djc/askama/blob/master/askama_tide/tests/tide.rs)
```
djc, line 60:
```suggestion
provides the ability for tide apps to build a response directly from
a template, or to append a templated body to an existing
`Response`. See [the example](https://github.com/djc/askama/blob/master/askama_tide/tests/tide.rs)
```

askama_tide/Cargo.toml
jbr: resolved in ceb9bbe5f37f62123c34d288831553476a640be6

README.md
jbr: I made this change in 3712761ac45ef1771d9f17a9afcc3cf57c064919 and also alphabetically sorted the calls and functions in generator.rs

askama_tide/LICENSE-APACHE
jbr: resolved in 413d4d4ce9fd197c69f896fe6d1eefea6ff96c01

askama_shared/src/generator.rs
jbr: makes sense to me! resolved in f0af044bc0cc564014a46e5b497d6a86a7507afa, and switched to re-exporting all of tide and askama instead of cherry-picking exports

djc:
Oh, forgot one very important thing... Can you add a CI job that tests the tide integration? Just copy one of the other (stable) integrations, should be straightforward. And please squash that change into the commit that has the other tide additions.

djc:
Thanks!

djc:
It's on crates.io now!

PR #352, opened 2020-08-03
tide: use set_error for correct error handling

jbr:
Shortly after askama integrated tide, tide 0.12 introduced the ability to store an error on a Response, allowing applications to inspect, downcast, log, and render the original error instead of serializing to a string. This PR uses Response::set_error to propagate askama errors


djc:
Thanks!

PR #359, opened 2020-09-16
fix into_f64 and into_isize

yshui:
askama passes references to the filters (e.g. &f64), and &f64 doesn't
implement NumCast, only f64 does.

So, the filters should take &T instead of T.

Signed-off-by: Yuxuan Shui <yshuiv7@gmail.com>


djc:
Can you fix up the tests? Ideally we'd add an integration test for this too, to show that it works correctly.

yshui:
Done

djc:
Just a little formatting and then it can be merged. 👍

yshui:
Formatted

djc:
Thanks!

PR #362, opened 2020-10-06
Disable tide default features

mkj:
In particular this avoids enabling the tide logger feature.


djc:
Looks like this needs at least one feature to be enabled!

mkj:
Sorry about that, should be right now.

djc:
Published an update as 0.11.1. Thanks!

PR #370, opened 2020-10-26
Fix links and actions which were broken since the default branch was renamed from master to main

JakeChampion:


djc:
Oops! Thanks for fixing these!

JakeChampion:
@djc great work on the project! :-D

PR #374, opened 2020-11-02
Use `syn::Error::to_compile_error` and add trybuild ui test

### LINKED ISSUE #368: Consider using `compile_error!` instead of panicking
I just tried out askama for the first time and my IDE (using rust-analyzer) reported that my code failed to compile because `proc-macro derive panicked`. This error message is not very helpful as I had to go to the command line and ask cargo what the real problem was. If the derive macro issued a `compile_error!(msg)` instead of panicking, the error message would've been correctly reported by rust-analyzer. This is also [already built into syn](https://docs.rs/syn/1.0.45/syn/struct.Error.html#method.to_compile_error) to make life easier.

msrd0:
This PR only contains one trybuild ui test, others can be added later if desired by putting more files into the `tests/ui/` folder.

Fixes #368

### DIFF
--- a/askama_derive/Cargo.toml
+++ b/askama_derive/Cargo.toml
@@ -25,5 +25,4 @@ warp = []
 [dependencies]
 askama_shared = { version = "0.10.4", path = "../askama_shared", default-features = false }
 proc-macro2 = "1"
-quote = "1"
 syn = "1"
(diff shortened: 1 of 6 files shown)


djc (approved):
This is great, thanks!

msrd0:
Looks like the test requires an operating system that uses forward slashes, do you want me to disable the test on windows?

djc:
Yeah, that seems like a fine solution for this.

PR #384, opened 2020-11-13
Fix urlencode to be usable for URL segments and query keys/values

JakubValtar:
Hey!

This PR makes `urlencode` filter behave like [`encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) and adds `urlencode_path`, which works the same, but preserves `/`.

Previous code behaved like [`encodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI), which should be run on a whole URI and preserves reserved characters which have a special meaning, like `&=?#`. This is typically not what you want: each segment should be urlencoded before joining them into the final URL, so that any reserved characters are safely encoded as data and do not interfere with the URI structure.

Jinja uses an overload, it preserves `/` when encoding strings, but encodes `/` (and replaces spaces by `+`) when encoding key/value pairs for query strings. I don't think encoding spaces in query strings as `+` instead of `%20` is mandatory, but I can add it. I didn't know what would be the best way to handle the slashes, so I added `urlencode_path` filter, which preserves slashes. There might be a better way to do this, like having one filter with a parameter, but I'm not that familiar with what `askama` can do with the filter syntax, so take this only as a proposal.

Another option would be to switch the names and use `urlencode` for slash preserving version (which is closer to the current functionality) and have `urlencode_query` which encodes slashes as well, but it seems that preserving slashes is going to be used less often, so I chose the longer name for it.

Examples
```
<a href="/route/page?key={{ "value/with#=reserved&?chars" | urlencode }}>Link</a>
<a href="/entities/{{ "id!with:symbols" | urlencode }}>Entity</a>
<a href="/files/{{ "path/to/file/ěščřéíáý" | urlencode_path }}>File</a>
```

If there are any use cases which are not covered or are broken by this, please mention them : )

### DIFF
--- a/askama_shared/src/filters/mod.rs
+++ b/askama_shared/src/filters/mod.rs
@@ -47,7 +47,7 @@ const URLENCODE_PATH_SET: &AsciiSet = &URLENCODE_SET.remove(b'/');
 // Askama or should refer to a local `filters` module. It should contain all the
 // filters shipped with Askama, even the optional ones (since optional inclusion
 // in the const vector based on features seems impossible right now).
-pub const BUILT_IN_FILTERS: [&str; 25] = [
+pub const BUILT_IN_FILTERS: [&str; 26] = [
     "abs",
     "capitalize",
     "center",
@@ -70,6 +70,7 @@ pub const BUILT_IN_FILTERS: [&str; 25] = [
     "upper",
     "uppercase",
     "urlencode",
+    "urlencode_path",
     "wordcount",
     "json", // Optional feature; reserve the name anyway
     "yaml", // Optional feature; reserve the name anyway
--- a/askama_shared/src/filters/mod.rs
+++ b/askama_shared/src/filters/mod.rs
@@ -30,29 +30,14 @@ use percent_encoding::{utf8_percent_encode, AsciiSet, NON_ALPHANUMERIC};
 use super::Result;
 
 #[cfg(feature = "percent-encoding")]
-// urlencode char encoding set, escape all characters except the following:
-// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI#Description
-const ENCODE_SET: &AsciiSet = &NON_ALPHANUMERIC
-    .remove(b';')
-    .remove(b',')
-    .remove(b'/')
-    .remove(b'?')
-    .remove(b':')
-    .remove(b'@')
-    .remove(b'&')
-    .remove(b'=')
-    .remove(b'+')
-    .remove(b'$')
-    .remove(b'-')
+// Urlencode char encoding set. Only the characters in the unreserved set don't
+// have any special purpose in any part of a URI and can be safely left
+// unencoded as specified in https://tools.ietf.org/html/rfc3986.html#section-2.3
+const URLENCODE_SET: &AsciiSet = &NON_ALPHANUMERIC
     .remove(b'_')
     .remove(b'.')
-    .remove(b'!')
-    .remove(b'~')
-    .remove(b'*')
-    .remove(b'\'')
-    .remove(b'(')
-    .remove(b')')
-    .remove(b'#');
+    .remove(b'-')
+    .remove(b'~');
 
 // This is used by the code generator to decide whether a named filter is part of
 // Askama or should refer to a local `filters` module. It should contain all the
@@ -132,10 +117,19 @@ pub fn filesizeformat<B: FileSize>(b: &B) -> Result<String> {
 }
 
 #[cfg(feature = "percent-encoding")]
-/// Returns the the UTF-8 encoded String of the given input.
+/// Encodes the argument for use in URL path or query.
+///
+/// Percent-encodes all characters except ASCII letters, digits, and `_.-~`.
+///
+/// Use this filter to encode URL path segments, query keys and query values
+/// before assembling the final URL.
+///
+/// ```ignore
+/// <a href="/page?text={{ val | urlencode }}">Link</a>
+/// ```
 pub fn urlencode(s: &dyn fmt::Display) -> Result<String> {
     let s = s.to_string();
-    Ok(utf8_percent_encode(&s, ENCODE_SET).to_string())
+    Ok(utf8_percent_encode(&s, URLENCODE_SET).to_string())
 }
 
 /// Formats arguments according to the specified format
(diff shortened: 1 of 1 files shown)


djc:
This looks like some great work! My main question here is about backwards compatibility: how many people are expecting `urlencode` to do the thing that you are making it do now, how many are relying on the old behavior, and does either one make it obvious that it broke? While we can note this in the release notes, I don't think relying on people reading the release notes is an ideal option. On the other hand, some folks (including myself) may have expected the existing `urlencode` to do what you're making it do now. Do you have a take on this?

My current feeling is that it might be safer/less surprising to keep the `urlencode` behavior as is, and add a new filter for the other behavior, but it sounds like you might be better at reasoning about what would happen to folks already using this (potentially wrongly).

JakubValtar:
<p>Backwards compatibility is a good question. I'm not sure how many people know about this filter, because it is not mentioned in the book.</p>
<p>The documentation says <code>Returns the the UTF-8 encoded String of the given input</code>, which could honestly mean anything. I think most people would expect it to make the input safe to use as part of an URL, the same as other templating engines. I don't think anyone would expect to use it to encode full URLs, which is a bit odd thing to do; how did you get URL with spaces and unicode characters in the first place?</p>
<p>However, all of this is only my assumption, and maybe I'm wrong about this. I tried to search GitHub for Askama templates using <code>urlencode</code>, but found only Jinja templates, so I have no data to support this. I at least looked through the most popular crates which depend on Askama, and didn't find any which uses the <code>urlencode</code> filter.</p>
<hr>
<p>I did a quick survey of other templating engines:</p>
<ul>
<li><code>Handlebars</code> has only HTML escape helper, but 3rd party helper pack includes <a href="https://github.com/clay/handlebars#urlencode--code--tests-"><code>urlencode</code></a>, which behaves like <code>encodeURIComponent</code> and encods slashes</li>
<li><code>Tera</code> also uses <code>Jinja</code> syntax. It has <a href="https://tera.netlify.app/docs/#urlencode" rel="nofollow"><code>urlencode</code></a> filter which preserves slashes, and <a href="https://tera.netlify.app/docs/#urlencode-strict" rel="nofollow"><code>urlencode_strict</code></a>, which encodes slashes, both work as <code>encodeURIComponent</code></li>
<li><code>Jinja</code> itself has <a href="https://jinja.palletsprojects.com/en/2.11.x/templates/#urlencode" rel="nofollow"><code>urlencode</code></a> which works like <code>encodeURIComponent</code> and conditionally encodes slashes</li>
<li><code>Nunjucks</code> has <a href="https://mozilla.github.io/nunjucks/templating.html#urlencode" rel="nofollow"><code>urlencode</code></a> which works like <code>encodeURIComponent</code> and encodes slashes</li>
<li><code>PHP</code> has <a href="https://www.php.net/manual/en/function.urlencode.php" rel="nofollow"><code>urlencode</code></a> function, which behaves like <code>encodeURIComponent</code> and encodes slashes</li>
</ul>
<p>I'm inclined to mirror Tera and use the same names for the filters. If we do this, this PR would break code for people who use <code>urlencode</code> to encode full URLs and assembled query strings (which is not a good idea anyway):</p>
<pre><code>&lt;a href="{{ "https://example.com/page.html" | urlencode }}&gt;Broken Link&lt;/a&gt;
&lt;a href="{{ "/page?key=value" | urlencode }}&gt;Broken Link&lt;/a&gt;
</code></pre>
<p>However, encoding filepaths will still work and be safer in case the path contains any reserved characters:</p>
<pre><code>&lt;a href="/files/{{ "path/to/file/ěš %# íá ?&amp;" | urlencode }}&gt;File&lt;/a&gt;
</code></pre>
<hr>
<p>TLDR: I think changing <code>urlencode</code> to behave like <code>encodeURIComponent</code> while preserving slashes, and adding a slash encoding version (maybe as <code>urlencode_strict</code> like in Tera) can be justified.</p>
<p>Alternatives would be:</p>
<ul>
<li>keep the current <code>urlencode</code> filter and add these as new filters, however I think the current name and functionality is deceptive</li>
<li>deprecate <code>urlencode</code> filter and add these as new filters, this will notify everyone to check their templates and migrate to one of the new filters, but I'm not sure if it's possible to deprecate (and we would have to use a less fitting name than <code>urlencode</code>)</li>
</ul>

djc:
Wow, thanks for doing the research on this! Your proposal makes sense to me. Can you make sure both variants have documentation that makes the difference clear?

JakubValtar:
Ok, I will update the PR to match Tera and improve the docs.

JakubValtar:
I renamed the filters and updated the docs (rendered below). Let me know how it's looking : )

## `urlencode`

Percent-encodes the argument for safe use in URI; does not encode `/`

This should be safe for all parts of URI (paths segments, query keys, query values). In the rare case that the server can't deal with forward slashes in the query string, use [`urlencode_strict`], which encodes them as well.

Encodes all characters except ASCII letters, digits, and `_.-~/`. In other words, encodes all characters which are not in the unreserved set, as specified by [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.3), with the exception of `/`.

```none,ignore
<a href="/metro{{ "/stations/Château d'Eau" | urlencode }}">Station</a>
<a href="/page?text={{ "look, unicode/emojis ✨" | urlencode }}">Page</a>
```

To encode `/` as well, see [`urlencode_strict`](./fn.urlencode_strict.html).

[`urlencode_strict`]: ./fn.urlencode_strict.html

## `urlencode_strict`

Percent-encodes the argument for safe use in URI; encodes `/`

Use this filter for encoding query keys and values in the rare case that the server can't process them unencoded.

Encodes all characters except ASCII letters, digits, and `_.-~`. In other words, encodes all characters which are not in the unreserved set, as specified by [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.3).

```none,ignore
<a href="/page?text={{ "look, unicode/emojis ✨" | urlencode_strict }}">Page</a>
```

If you want to preserve `/`, see [`urlencode`](./fn.urlencode.html).

djc:
This is great, thanks!

djc:
I've pushed out askama_shared 0.10.5 with this fix.

JakubValtar:
Amazing, thanks for a quick review, Dirkjan : )

PR #388, opened 2020-11-21
Disable nom default features

xfix:
This removes unnecessary `lexical` feature reducing amount of crates Askama depends on.

Unfortunately this doesn't disable `bitvec` feature because `std` depends on `bitvec/std`, and currently Cargo doesn't have support for weak dependencies on stable (https://github.com/rust-lang/cargo/issues/8832).


djc:
Thanks!

djc:
I guess you might conceivably split out nom's indirect std feature dependencies out into a dep-std feature (enabled in default), so downstream crates like Askama have more control.

PR #390, opened 2020-12-01
Fixed implicit borrow of expressions

vallentin:
This does not introduce any breaking changes<sup>\*</sup>.

The change fixes so expressions are now borrowed as a whole, instead of only the first value. Which would cause compile errors, for types that don't implement e.g. `Add<T> for &T`.

In short expressions now expand into `&(...)` instead of `&...`.

_<sup>\*</sup> Technically, it would be considered a breaking change. However, I'd say it's highly unlikely someone has implemented e.g. `Add<T> for &T` and not `Add<T> for T`. To be able to rely on something that could be considered inconsistent behavior._

-----

### Example 1

Minimal example:

```jinja
{{ true && true }}
```

Which expands like this:

**Before:** `&true && true`
**After:** `&(true && true)`

Resulting in the following compile error:

```rust
error[E0308]: mismatched types
   |
11 | #[derive(Template)]
   |          ^^^^^^^^ expected `bool`, found `&bool`
```

This isn't and was never an issue for e.g. `{{ 1 + 2 }}`, as the various operators e.g. `Add<&T> for T` are implemented for various types. However, this is not the case for short-circuit operators.

To have a more complete example of when this applies. I personally have a few macros, which accept booleans to decide on what it expands to.

```jinja
{% macro foo(b) %}
    {% if b.clone() %}
        ...
    {% endif %}
{% endmacro %}

{% call foo(true && true) %}
```

Which expands like this:

**Before:** `let (b) = (&true && true);`
**After:** `let (b) = (&(true && true));`

_The fact that `.clone()` is needed, is a whole other can of worms._

-----

### Example 2

Another example where this also manifests itself is:

```rust
#[derive(Template)]
#[template(path = "foo.html.jinja")]
struct Foo {
    time: SystemTime,
    dur: Duration,
}
```

```jinja
{{ self.time + self.dur }}
```

Once again only [`Add<Duration> for SystemTime`](AddDurTime) exists, so this also resulted in a compile error:

[AddDurTime]: https://doc.rust-lang.org/std/ops/trait.Add.html#impl-Add%3CDuration%3E-2

```rust
error[E0369]: cannot add `Duration` to `&SystemTime`
```

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -945,7 +945,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                     let expression = match wrapped {
                         Wrapped => expr_buf.buf,
                         Unwrapped => format!(
-                            "::askama::MarkupDisplay::new_unsafe(&{}, {})",
+                            "::askama::MarkupDisplay::new_unsafe(&({}), {})",
                             expr_buf.buf, self.input.escaper
                         ),
                     };
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -717,10 +717,11 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             }
             names.write(arg);
 
-            values.write("&");
+            values.write("&(");
             values.write(&self.visit_expr_root(args.get(i).ok_or_else(|| {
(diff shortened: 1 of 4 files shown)


djc:
LGTM, thanks!

PR #391, opened 2020-12-01
Fixed parsing precedence and associativity

### LINKED ISSUE #381: The operator parser gets precedences wrong
Consider the expression `a + b == c`. If we changed the code generated from `{lhs} {op} {rhs}` to include parentheses around the entire thing, we get `(a + (b == c))`, which is incorrect, because _(i)_ the `+` and `==` operators combined are not associative (and I think that's true for any two different operators), so need to be evaluated left to right, and _(ii)_ the `==` operator has higher precedence than `+`. It only currently works because askama emits the code unmodified enough for the rust compiler to correct the mistake.

vallentin:
Fixes #381

This pull request fixes the parser, such that it produces a correct AST in relation to precedence and associativity.

As mentioned in the issue, given that Askama emits the expressions unmodified, then in that regard there isn't any _noticeable_ change. However, with a correct AST it prevents any potential future problems.

In short `a + b == c` was previously parsed as `(a + (b == c))` and now is correctly parsed as `((a + b) == c)`.

I also added a few test cases to check that the parser upholds precedence and associativity. (If you want the test cases split into multiple functions, then mention it and I'll split them.)

### DIFF
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -1080,7 +1080,7 @@ mod tests {
     fn check_ws_split(s: &str, res: &(&str, &str, &str)) {
         let node = super::split_ws_parts(s.as_bytes());
         match node {
-            super::Node::Lit(lws, s, rws) => {
+            Node::Lit(lws, s, rws) => {
                 assert_eq!(lws, res.0);
                 assert_eq!(s, res.1);
                 assert_eq!(rws, res.2);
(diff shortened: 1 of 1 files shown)


djc:
Awesome, thanks for fixing this! I'll merge as soon as this passes CI.

vallentin:
Fixing the issue.

PR #392, opened 2020-12-01
Improved if statement generation to avoid issues with implicit borrows

vallentin:

Having an `if` within a `macro` is cumbersome as the parameters are referenced, requiring `if b` to become `if b.clone()`. This becomes even more cumbersome if macros call other macros, requiring various amounts of `.clone()` to be sprinkled at various places.

I changed `write_cond` to utilize Rust's automatic dereferencing rules to coerce e.g. `&bool`, `&&bool`, `&&&bool`, etc. to `bool`. This improves dealing `if b`, as now `.clone()` is no longer needed.

-----

Minimal example to highlight the issue previously:

```jinja
{% macro foo(b) %}
    {% if b %}
    {% endif %}
{% endmacro %}

{% macro bar(b) %}
    {% call foo(b) %}
{% endmacro %}

{% call foo(true) %}
{% call bar(true) %}
```

Which _essentially_ expands to:

```rust
let b = &true;
if b {}
let b = &&true;
if b {}
```

Causing compile errors as `if` expects `bool`:

```rust
error[E0308]: mismatched types
   |
22 | #[derive(Template)]
   |          ^^^^^^^^ expected `bool`, found `&bool`
error[E0308]: mismatched types
   |
22 | #[derive(Template)]
   |          ^^^^^^^^ expected `bool`, found `&&bool`
```

_I also added a new test case, to ensure this behavior persists._

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -518,8 +518,15 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                         buf.dedent()?;
                         buf.write("} else if ");
                     }
+                    // The following syntax `*(&(...) as &bool)` is used to
+                    // trigger Rust's automatic dereferencing, to coerce
+                    // e.g. `&&&&&bool` to `bool`. First `&(...) as &bool`
+                    // coerces e.g. `&&&bool` to `&bool`. Then `*(&bool)`
+                    // finally dereferences it to `bool`.
+                    buf.write("*(&(");
                     let expr_code = self.visit_expr_root(expr)?;
                     buf.write(&expr_code);
+                    buf.write(") as &bool)");
                 }
                 None => {
                     buf.dedent()?;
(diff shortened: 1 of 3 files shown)


djc:
Thanks, looks great!

vallentin:
@djc I had another idea, which was to have a "`Truthy`" trait. The same automatic dereferencing rules apply to traits, so instead of generating `*(&(expr) as &bool)` then instead it could simply generate `expr.truthy()`.

```rust
trait Truthy {
    fn truthy(&self) -> bool;
}

impl Truthy for bool {
    #[inline]
    fn truthy(&self) -> bool {
        *self
    }
}
```

The _potentially_ nice thing about that instead, would be that it allows to do stuff like this:

```rust
impl Truthy for String {
    #[inline]
    fn truthy(&self) -> bool {
        !self.is_empty()
    }
}
```

Which would make Askama's behavior match more closely to the behavior of web languages and Jinja in Python. Then again, being explicit with `if !s.is_empty()` is probably better. So I'm just throwing the idea out there! What do you think?

djc:
It's a nice idea, but for this crate I'd rather stick to the more explicit Rust variant.

vallentin:
Yes of course, I agree. I just wanted to share the idea.

Issue #377, opened 2020-11-06
Whitespace issue with if conditions


ajeetdsouza:
Askama seems to be including whitespace from inside an if condition even when the condition is not fulfilled. Here is my template file:

```jinja
{%- if some_condition -%}
# WARNING

{% endif -%}

# FILE START
```

What I expected is:

- When `some_condition` is `true`:

  ```
  # WARNING
  
  # FILE START
  ```

- When `some_condition` is `false`:

  ```
  # FILE START
  ```

What I got is:

- When `some_condition` is `true`: same as expected output
- When `some_condition` is `false` (notice the blank lines before `FILE START`):

  ```
  
  
  # FILE START
  ```

Is this a bug? If not, is there a way to get the behaviour I described?

djc:
It sure sounds like a bug! Let me look at some code.

djc:
So the code generation for if statements is here:

https://github.com/djc/askama/blob/main/askama_shared/src/generator.rs#L502

I don't immediately see anything obviously wrong. Maybe you can contribute a test case and/or use the debugging facility (add `print = "code"` to your `template` attribute) to look at the generated code and see what's going on?

ajeetdsouza:
@djc I've added the test. Using `print = "code"` makes the bug more apparent - it's adding the newlines outside the if condition rather than inside:

```rust
fn render_into(&self, writer: &mut dyn ::std::fmt::Write) -> ::askama::Result<()> {
    include_bytes !
("/home/ajeetdsouza/ws/askama/testing/templates/cond-whitespace.txt") ;
    if self.show_preamble {
        writer.write_str("preamble")?;
    }
    writer.write_str("\n\nintroduction\n\nconclusion")?;
    Ok(())
}
```

PR #393, opened 2020-12-02
Allow paths to start with `::`

### LINKED ISSUE #331: Allow paths to start with `::`
[Discovered while working on #315 and #327.]

At the moment `{{ std::string::String::new()}}` works but `{{ ::std::string::String::new() }}` doesn't.

vallentin:
Fixes #331

I wanted to avoid changing `Path` and `PathCall` and the parameters of the visit functions. So instead I decided to just prepend `"::"` to the path parts `Vec`. If you'd rather have I add a `bool` to `Path` and `PathCall` I can do that instead.

### DIFF
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -1139,6 +1139,25 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_root_path() {
+        let syntax = Syntax::default();
+        assert_eq!(
+            super::parse("{{ std::string::String::new() }}", &syntax).unwrap(),
+            vec![Node::Expr(
+                WS(false, false),
+                Expr::PathCall(vec!["std", "string", "String", "new"], vec![]),
+            )],
+        );
+        assert_eq!(
+            super::parse("{{ ::std::string::String::new() }}", &syntax).unwrap(),
+            vec![Node::Expr(
+                WS(false, false),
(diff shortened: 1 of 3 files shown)


askama_shared/src/parser.rs
```diff
@@ -312,10 +312,12 @@ fn expr_var_call(i: &[u8]) -> IResult<&[u8], Expr> {
 }
 
 fn path(i: &[u8]) -> IResult<&[u8], Vec<&str>> {
+    let root = opt(value("::", ws(tag("::"))));
```
djc, line 315: How about we use an empty string instead of "::"? That way, the code generator can remain as it is (a fair bit simpler).

djc (commented):
Nice! I think there's a simpler way to solve this...

askama_shared/src/parser.rs
vallentin: Damn... Brain fart! 😄

vallentin:
@djc can I write to you on Twitter, or do you have another better way for sending you a direct message?

djc:
I'm everywhere! Twitter is fine, direct email, Tokio or Rust Discord, Askama Gitter, Mozilla Matrix, Rust Zulip... pick your poison.

vallentin:
Alright, I'll send you a message in a bit. Just finishing up resolving #377.

PR #394, opened 2020-12-03
Fixed whitespace issue when generating if statement

### LINKED ISSUE #377: Whitespace issue with if conditions

vallentin:
Fixed #377
Closes #378

I've fixed the issue of whitespace being generated incorrectly in if statements.

I've tested all permutations from 1 to 4 branches. For which I created a Python script that generated all the permutations. I included the script as well, as it allows you to generate and test an arbitrary number of branches. (_I'll use it to check `match`, `macro`, along with nested constructs later._)

I included the test cases for 1 and 2 branches.


vallentin:
Forgot to re-generate the test file.

djc:
Once again, thank you!

Issue #329, opened 2020-05-25
Add support for redefining the same variable with `let` (just as Rust permits)


cipriancraciun:
[Discovered while working on #315 and #327.]

The following statements, although allowed by Rust, are disallowed by `askama`:
~~~~
{% let x = 1 %}
{% let x = 2 %}
~~~~

The failure is:
~~~~
cannot assign twice to immutable variable
help: make this binding mutable: `mut x`
~~~~

djc:
I'm guessing this should be easy to fix.

max-frai:
Any news on this? It's very important when splitting template into small parts with ability to `include` them and setting some variables before include.

djc:
I haven't had a chance to work on it, and it's not much of a priority for me right now. I might be able to mentor you if you want to dig in, or you can pay me for a few hours of work on this, if that's a viable option for you.

vallentin:
I wonder if we should include Jinja's `{% set ... = ... %}` and use it to differentiate between defining and assigning. We could also make the `set` keyword optional, I'm just suggesting it to be more in line with Jinja.

In short:

- `{% let x = ... %}` would always (re)define the variable, i.e. result in `let x = ...;`
- `{% set x = ... %}` would (re)assign the variable, i.e. result in `x = ...;`
- Any use of `{% set x = ... %}` would result in the already declared variable to become mutable, i.e. `let mut x = ...;`
- If `{% set x = ... %}` is not accompanied by a previous `{% let x %}` or `{% let x = ... %}`, then it will implicitly generate `let mut x;`

Example Template:

```jinja
{% let x = "foo" %}
{% let x = "bar" %}

{% set y = "foo" %}
{% set y = "bar" %}

{% let z %}
{% set z = "foo" %}
{% set z = "bar" %}
```

Then the above template would generate the following Rust:

```rust
let x = "foo";
let x = "bar";

let mut y = "foo";
y = "bar";

let mut z;
z = "foo";
z = "bar";
```

djc:
I don't like the idea of `set` and `let` having different semantics (which I think would not be obvious). I also don't like introducing mutability here -- I feel like `let` bindings in Askama templates should always be immutable. I am open to shadowing, following general Rust semantics. We could also just make `set` an alias for `let`, which seems innocuous enough and might be helpful for people porting code from another template engine.

vallentin:
Alright, I'll implement that sometime soon, when I'm able to.

- [ ] Shadowing
[...]

PR #396, opened 2020-12-07
Reworked implicit borrowing to be more relaxed

### LINKED ISSUE #107: Cannot move out of borrowed content in for loop

### LINKED ISSUE #376: Allow macro blocks in included templates
I'm trying to have a `nav.html` template included in several other templates via `{% include "nav.html" %}` and was very supprised that I couldn't use macros in that template because _macro blocks only allowed at the top level_. Please allow macros to appear in included templates, this removes a lot of tedious and error-prone duplication.

vallentin:
Fixes #376 (can't compare `&str` with `str`, while `import`/`include` remains the same by design)

This PR changes `SetChain` into `MapChain`, where the value represents an optional variable which the local variable "points to". This improves the generated Rust code, as existing variables are no longer continuously generated for every macro call. This means that variables are no longer continuously borrowed, e.g. making it possible to `s == "foo"` in a macro.

No test cases were modified, and this is now a non-breaking change.

-----

**Old Explanation:**

To be able to reuse variables in nested macro calls, I ended up reworking `SetChain` into `MapChain`. Such that the each local can have an optional variable (or expression) which it points to. Then I added `resolve_var` to be able to follow and resolve a chain if variables, e.g. `"param"` -> `"a"` -> `"b"` -> `"self.c"`.

This avoid that previously when `let (..) = (..)` would be defined in the local scope, it would either (a) cause values to be moved if non-copyable or (b) force all values to be referenced. This would cause every level of nested macro call, to increase the amount of indirection (i.e. 3 depth macro call == `&&&T`). This would makeb it harder to use `.clone()`, but now it does not exceed `&T` so using `.clone()` is _always_ possible.

Given that [`HashSet<T>`] internally uses a `HashMap<T, ()>`, then a `SetChain` can still be created using, e.g. `type SetChain<'a, T> = MapChain<'a, T, ()>`. Which implies that they are still identical. I have not changed "`SetChain`"'s functionality, I've only added `get` and `get_skip` as well as `resolve_var` which I implemented specifically for `MapChain<'_, &str, Option<String>>`.

_Additionally, now empty `let () = ();` would be generated more frequently, since variables are forwarded. So these are also culled from the output._

[`HashSet<T>`]: https://doc.rust-lang.org/std/collections/struct.HashSet.html

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -10,7 +10,7 @@ use proc_macro2::Span;
 
 use quote::{quote, ToTokens};
 
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::path::PathBuf;
 use std::{cmp, hash, mem, str};
 
@@ -20,7 +20,7 @@ pub fn generate<S: std::hash::BuildHasher>(
     heritage: &Option<Heritage>,
     integrations: Integrations,
 ) -> Result<String, CompileError> {
-    Generator::new(input, contexts, heritage, integrations, SetChain::new())
+    Generator::new(input, contexts, heritage, integrations, MapChain::new())
         .build(&contexts[&input.path])
 }
 
@@ -34,7 +34,7 @@ struct Generator<'a, S: std::hash::BuildHasher> {
     // What integrations need to be generated
     integrations: Integrations,
     // Variables accessible directly from the current scope (not redirected to context)
-    locals: SetChain<'a, &'a str>,
+    locals: MapChain<'a, &'a str, LocalMeta>,
     // Suffix whitespace from the previous literal. Will be flushed to the
     // output buffer unless suppressed by whitespace suppression on the next
     // non-literal.
@@ -56,7 +56,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         contexts: &'n HashMap<&'n PathBuf, Context<'n>, S>,
         heritage: &'n Option<Heritage>,
         integrations: Integrations,
-        locals: SetChain<'n, &'n str>,
+        locals: MapChain<'n, &'n str, LocalMeta>,
     ) -> Generator<'n, S> {
         Generator {
             input,
@@ -73,7 +73,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     }
 
     fn child(&mut self) -> Generator<'_, S> {
-        let locals = SetChain::with_parent(&self.locals);
+        let locals = MapChain::with_parent(&self.locals);
         Self::new(
             self.input,
             self.contexts,
@@ -592,7 +592,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                         buf.write("(");
                         for (i, param) in params.iter().enumerate() {
                             if let MatchParameter::Name(p) = *param {
-                                self.locals.insert(p);
+                                self.locals.insert_with_default(p);
                             }
                             if i > 0 {
                                 buf.write(", ");
@@ -606,9 +606,9 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                     buf.write("{");
                     for (i, param) in params.iter().enumerate() {
                         if let Some(MatchParameter::Name(p)) = param.1 {
-                            self.locals.insert(p);
+                            self.locals.insert_with_default(p);
                         } else {
-                            self.locals.insert(param.0);
+                            self.locals.insert_with_default(param.0);
                         }
 
                         if i > 0 {
@@ -719,21 +719,49 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         let mut names = Buffer::new(0);
         let mut values = Buffer::new(0);
         for (i, arg) in def.args.iter().enumerate() {
-            if i > 0 {
-                names.write(", ");
-                values.write(", ");
-            }
-            names.write(arg);
-
-            values.write("&(");
-            values.write(&self.visit_expr_root(args.get(i).ok_or_else(|| {
+            let expr = args.get(i).ok_or_else(|| {
                 CompileError::String(format!("macro '{}' takes more than {} arguments", name, i))
-            })?)?);
-            values.write(")");
-            self.locals.insert(arg);
+            })?;
+
+            match expr {
+                // If `expr` is already a form of variable then
+                // don't reintroduce a new variable. This is
+                // to avoid moving non-copyable values.
+                Expr::Var(name) => {
+                    let var = self.locals.resolve_or_self(name);
+                    self.locals.insert(arg, LocalMeta::with_ref(var));
+                }
+                Expr::Attr(obj, attr) => {
+                    let mut attr_buf = Buffer::new(0);
+                    self.visit_attr(&mut attr_buf, obj, attr)?;
+
+                    let var = self.locals.resolve(&attr_buf.buf).unwrap_or(attr_buf.buf);
+                    self.locals.insert(arg, LocalMeta::with_ref(var));
+                }
+                // Everything else still needs to become variables,
+                // to avoid having the same logic be executed
+                // multiple times, e.g. in the case of macro
+                // parameters being used multiple times.
+                _ => {
+                    if i > 0 {
+                        names.write(", ");
+                        values.write(", ");
+                    }
+                    names.write(arg);
+
+                    values.write("(");
+                    values.write(&self.visit_expr_root(expr)?);
+                    values.write(")");
+                    self.locals.insert_with_default(arg);
+                }
+            }
         }
 
-        buf.writeln(&format!("let ({}) = ({});", names.buf, values.buf))?;
+        debug_assert_eq!(names.buf.is_empty(), values.buf.is_empty());
+        if !names.buf.is_empty() {
+            buf.writeln(&format!("let ({}) = ({});", names.buf, values.buf))?;
+        }
+
         let mut size_hint = self.handle(own_ctx, &def.nodes, buf, AstLevel::Nested)?;
 
         self.flush_ws(def.ws2);
@@ -794,13 +822,13 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         buf.write("let ");
         match *var {
             Target::Name(name) => {
-                self.locals.insert(name);
+                self.locals.insert_with_default(name);
                 buf.write(name);
             }
             Target::Tuple(ref targets) => {
                 buf.write("(");
                 for name in targets {
-                    self.locals.insert(name);
+                    self.locals.insert_with_default(name);
                     buf.write(name);
                     buf.write(",");
                 }
@@ -823,16 +851,16 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
 
         match *var {
             Target::Name(name) => {
-                if !self.locals.contains(name) {
+                if !self.locals.contains(&name) {
                     buf.write("let ");
-                    self.locals.insert(name);
+                    self.locals.insert_with_default(name);
                 }
                 buf.write(name);
             }
             Target::Tuple(ref targets) => {
                 buf.write("let (");
                 for name in targets {
-                    self.locals.insert(name);
+                    self.locals.insert_with_default(name);
                     buf.write(name);
                     buf.write(",");
                 }
@@ -1371,12 +1399,12 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     }
 
     fn visit_var(&mut self, buf: &mut Buffer, s: &str) -> DisplayWrap {
-        if self.locals.contains(s) || s == "self" {
+        if s == "self" {
             buf.write(s);
-        } else {
-            buf.write("self.");
-            buf.write(s);
+            return DisplayWrap::Unwrapped;
         }
+
+        buf.write(&self.locals.resolve_or_self(&s));
         DisplayWrap::Unwrapped
     }
 
@@ -1387,7 +1415,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         args: &[Expr],
     ) -> Result<DisplayWrap, CompileError> {
         buf.write("(");
-        if self.locals.contains(s) || s == "self" {
+        if self.locals.contains(&s) || s == "self" {
             buf.write(s);
         } else {
             buf.write("self.");
@@ -1422,13 +1450,13 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     fn visit_target(&mut self, buf: &mut Buffer, target: &'a Target) {
         match *target {
             Target::Name(name) => {
-                self.locals.insert(name);
+                self.locals.insert_with_default(name);
                 buf.write(name);
             }
             Target::Tuple(ref targets) => {
                 buf.write("(");
                 for name in targets {
-                    self.locals.insert(name);
+                    self.locals.insert_with_default(name);
                     buf.write(name);
                     buf.write(",");
                 }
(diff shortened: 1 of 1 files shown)


vallentin:
I've sent you a message on Twitter, in case you haven't noticed.

djc:
I think I like this direction but in order to be able to review/understand it well I feel like I need more guidance, and perhaps I'd also want to aim for splitting up the initial commit here into smaller changes. Let's maybe start by leaving new tests out of this PR for now so I can focus on reviewing the logic changes and the changes to existing tests. Also, please squash intermediate commits like `Reverted TemplateLoop` and `Updated indent` that changes things from earlier commits into their earlier commits (one option here is the the git-absorb git extension).

First, how do you see this affecting backwards compatibility for existing templates?

Second, what is the goal of the change from `SetChain` to `MapChain`? What additional information is it tracking, and how?

vallentin:
> to review/understand it well I feel like I need more guidance

The _actual_ change to the generator isn't _that_ crazy. It's overall just:

1. Removing a bunch of `"&"`
2. The places where it does `"&"`, do so if `!is_copyable()`
3. Reuse `Expr::Var` when possible, instead of emitting a `let ...` for every macro call

But tell me how I can help or feel free to comment on all the lines you feel necessary, and I'll comment. _(Maybe wait until I've rebased.)_

-----

> Second, what is the goal of the change from `SetChain` to `MapChain`? What additional information is it tracking, and how?

In short, consider the following template:

```jinja
{% macro bar(y) %}
    {{ y }}
    {% if y == "bar" %}
    {% endif %}
{% endmacro %}

{% macro foo(x) %}
    {{ x }}
{% endmacro %}

{% call foo(self.item) %}
```

Then before it would essentially generate:

_(Which would result in a compile error for the if)_

```rust
let x = &self.item;
let y = &x;
write(&y);
if &y == "bar" {}
```

Whereas now it essentially generates:

```rust
write(&self.item);
if self.item == "bar" {}
```

All in all, now you're able to do `y == "bar"` in `macro bar`, as `self.item` is propagated all the way, and whether it is borrowed, is now decided when it is used.

So in short, to be able to know which variable a local variable _points to_, that information is stored in the value part of the `HashMap`. To reuse the same example `x` points to `self.item`, `y` points to `x`. So when `y` is finally emitted, then it is resolved back to `self.item`.

So since that information needed to be part of every variable in every scope, then I thought the simplest way would be to change `HashSet<T>` to `HashMap<K, V>`. Functionally `MapChain<T, ()>` is the equivalent to `SetChain<T>`.

-----

> Backwards compatibility for existing templates?

Technically, it's a breaking change.

However, only 4 of the existing test cases were affected, 1 of which I introduced 11 days ago (#393). The affected test cases, are functions that had a borrowed literal as a parameter (e.g. `&u8`). Technically, if you did `{{ self::foo(self.i) }}` then it would require `.clone()` as anything accessed through `self` is still borrowed.

So realistically, I don't think it will be an issue, and if someone did encounter issues, then adding `.clone()` or inversely `.as_ref()` is enough to fix those few cases.

_Though, maybe the previous non-breaking changes should be released as a patch (v0.10.6), and then after we can bump the minor version (v0.11.0) and include this change._

-----

> I'd also want to aim for splitting up the initial commit here into smaller changes

Maybe it will be possible to split the `write_loop` change. But I'm unsure how much else of the generator changes can be split.

> squash intermediate commits like `Reverted TemplateLoop` and `Updated indent` that changes things from earlier commits into their earlier commits

👍

-----

> Let's maybe start by leaving new tests out of this PR for now so I can focus on reviewing the logic changes and the changes to existing tests

Sure, I wanted to rewrite those into the existing test cases, in a more _approachable_ manner anyways. As it is indeed a bit much, I just wanted to test every case I could think of, to ensure they wouldn't break later.

So, do I just drop the commit for the new test cases? (I'll keep a local copy, to rework into the other test cases for later then)

djc:
Ok, that all sounds good. Definitely split off the new test cases for a later PR. In fact, if it's not too hard to extract from the other changes, I'd prefer to focus this PR only on the change around the `MapChain`, that is (if I understand correctly) to "resolve" Askama-local variables at (template) compile time, and then move the rest of the implicit borrowing changes out to yet another PR.

vallentin:
I've updated the PR. Now it only includes the `SetChain` to `MapChain` change. Which also means it now is a non-breaking change.

In short:
- No tests have been modified, nor affected
- The filters remain as they were prior to this PR (i.e. they still take literals by reference)
- Calling is back to using a references (which is why it's a non-breaking change now)

_I've updated the PR description to reflect the changes._

askama_shared/src/generator.rs
```diff
@@ -719,21 +719,58 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+            match expr {
+                // If `expr` is already a form of variable then
+                // don't reintroduce a new variable. This is
+                // to avoid moving non-copyable values.
+                Expr::Var(name) => {
+                    let var = self
+                        .locals
+                        .resolve_var(name)
+                        .map(ToString::to_string)
+                        .unwrap_or_else(|| format!("self.{}", name));
```
djc, line 735: I feel like this is both more compact and easier to understand. What do you think?

```suggestion
                    let var = match self.locals.resolve_var(name) {
                        Ok(s) => s.to_string(),
                        None => format!("self.{}", name)),
                    };
```

askama_shared/src/generator.rs
```diff
@@ -719,21 +719,58 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+                    let mut attr_buf = Buffer::new(0);
+                    self.visit_attr(&mut attr_buf, obj, attr)?;
+
+                    let var = self
+                        .locals
+                        .resolve_var(&attr_buf.buf)
+                        .map(ToString::to_string)
+                        .unwrap_or(attr_buf.buf);
+                    self.locals.insert(arg, Some(var));
+                    continue;
```
djc, line 748: The `continue` here isn't needed, I think?

askama_shared/src/generator.rs
```diff
@@ -1523,51 +1567,73 @@ impl Buffer {
...
                 None => false,
             }
     }
 
+    fn get(&self, key: &K) -> Option<&V> {
+        self.get_skip(key, 0)
+    }
+
+    /// Returns `Some` if `key` exists, while `skip` represents
+    /// the amount of previous scopes to skip in reverse.
```
djc, line 1612: The explanation here in the latter part of the sentence doesn't make sense to me. Can you explain it some other way? Conceptually, what is this for?

askama_shared/src/generator.rs
```diff
@@ -1523,51 +1567,73 @@ impl Buffer {
...
+        self.scopes
+            .iter()
+            .rev()
+            .skip(skip)
+            .filter_map(|set| set.get(&key))
+            .next()
+            .or_else(|| match self.parent {
+                Some(set) => set.get(key),
+                None => None,
+            })
```
djc, line 1623: This is pretty opaque to me. Maybe name some intermediate values (which might also make it more compact)?

askama_shared/src/generator.rs
```diff
@@ -1523,51 +1567,73 @@ impl Buffer {
...
+    }
+
     fn is_current_empty(&self) -> bool {
         self.scopes.last().unwrap().is_empty()
     }
 
-    fn insert(&mut self, val: T) {
-        self.scopes.last_mut().unwrap().insert(val);
+    fn insert(&mut self, key: K, val: V) {
+        let old_val = self.scopes.last_mut().unwrap().insert(key, val);
```
djc, line 1631: Maybe add a comment documenting why the asserted invariant should hold?

askama_shared/src/generator.rs
```diff
@@ -1576,6 +1642,30 @@ where
     }
 }
 
+impl MapChain<'_, &str, Option<String>> {
+    /// Given a variable `key` it resolves all the way
+    /// back to the initial the variable.
+    fn resolve_var<'a>(&'a self, key: &'a str) -> Option<&'a str> {
```
djc, line 1648: What's the argument against just calling this `resolve()`?

askama_shared/src/generator.rs
```diff
@@ -1576,6 +1642,30 @@ where
...
+        let mut key = key;
+        let mut var = None;
+        for i in 1.. {
+            let prev_var = self.get_skip(&key, i);
+            match prev_var {
```
djc, line 1653:
```suggestion
            match self.get_skip(&key, i) {
```

askama_shared/src/generator.rs
```diff
@@ -34,7 +34,7 @@ struct Generator<'a, S: std::hash::BuildHasher> {
     // What integrations need to be generated
     integrations: Integrations,
     // Variables accessible directly from the current scope (not redirected to context)
-    locals: SetChain<'a, &'a str>,
+    locals: MapChain<'a, &'a str, Option<String>>,
```
djc, line 37: I feel like making `MapChain` abstract over the key type here isn't actually useful, and doing the same thing for the value argument makes it nice and general at the cost of having to deal with these nested options. Do you think we can flatten the abstraction here? It really doesn't need to be general anyway.

askama_shared/src/generator.rs
vallentin: Needs to be `Some(s)` and you had a `)` to many at the end. But I've applied the change :)

askama_shared/src/generator.rs
vallentin: Hmm, weird. Must be left over from when I originally only had 1 arm, and didn't use a catchall pattern. Removed.

askama_shared/src/generator.rs
vallentin: Originally, `resolve_var` was a helper function, so I kinda just kept the name. But I'll rename it to `resolve()` :)

askama_shared/src/generator.rs
vallentin: Do you mean something like this (1):

```rust
let scores = self.scopes.iter().rev().skip(skip);
scores
    .filter_map(|set| set.get(&key))
    .next()
    .or_else(|| self.parent.and_then(|set| set.get(key)))
```

or (2):

```rust
let scores = self.scopes.iter().rev().skip(skip);
let var = scores.filter_map(|set| set.get(&key)).next();
var.or_else(|| self.parent.and_then(|set| set.get(key)))
```

It's more compact (less lines), but personally I find it harder to read. Unless, that wasn't what you meant? :)

djc: Do you find both of them harder to read, or just the latter?  I think I prefer option (1) (but I'm still a bit iffy on what this is actually intended to achieve).

askama_shared/src/generator.rs
vallentin: So I have a related question. I've looked a bit into #329 (shadowing). My initial impression is that locals need to have an `initialized: bool`, to be able to know if it should shadow or assign. So I'm considering changing `Option<String>` into `LocalMeta`.

So for this PR, I could change it to:

```rust
struct LocalMeta {
    references: Option<String>,
}
```

Then in the PR for shadowing, I can introduce `initialized`.

```rust
struct LocalMeta {
    references: Option<String>,
    initialized: bool,
}
```

What do you think about that?

djc: Sounds about right. BTW, does the references value need to be `String`? Why can't it be `&str`?

askama_shared/src/generator.rs
vallentin: Wait, option (1) is fine, yeah.

Originally, I stored variable names in the `Option<String>` and not the resolved expression. So previously, if you did `{% call foo(a) %}` and then the macro parameter was also called `a`. Then to be able to resolve it, it required being able to `get()` while "skipping" the last `N` scopes (to be able to escape the macro's scope).

However, now that it stored the "resolved expression", then I'm feeling it might actually be possible to remove `get_skip` entirely, and then just use `get()`. I'll look into it.

askama_shared/src/generator.rs
vallentin: Alright, I'll change it to that.

It can't because in `write_call` it's used to store the result of `visit_attr`. Though, if `self.` doesn't need to be prefixed, then I might be able to change it to `Cow<'a, str>`.

djc: Let's leave it as `String` for now, probably not a big deal anyway.

askama_shared/src/generator.rs
vallentin: Hmm, we could also remove the `assert` because Rust itself will error when attempting to compile the generated template. Basically, the cases are e.g. `{% macro f(a, a) %}` and `{% let (a, a) = ... %}` which both results in a compile error along the lines of "identifier `a` used more than once" .

I'll remove it. Because if we want to add `"identifier {:?} used more than once"` to the assert, then `insert` requires `where K: Debug + Clone/Copy`.

vallentin:
- Refactored into using `struct LocalMeta`.
- Removed `get_skip` and `resolve_var`.
  - In `write_call` the expression is resolved up front. So since it's resolved then `get()` always get's the actual expression, so no need to `resolve_var` and traverse backwards anymore.

askama_shared/src/generator.rs
```diff
@@ -1371,11 +1411,20 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+
+        match self.locals.get(&s) {
+            Some(meta) => match &meta.refs {
+                Some(expr) => buf.write(expr),
+                None => buf.write(s),
+            },
+            None => {
+                buf.write("self.");
+                buf.write(s);
+            }
```
djc, line 1427: Also looks like essentially the same pattern as above.

askama_shared/src/generator.rs
```diff
@@ -592,7 +592,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                         buf.write("(");
                         for (i, param) in params.iter().enumerate() {
                             if let MatchParameter::Name(p) = *param {
-                                self.locals.insert(p);
+                                self.locals.insert(p, LocalMeta::default());
```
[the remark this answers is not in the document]
vallentin: I propose `insert_with_default`, and I assume you mean adding it to `MapChain` and not `LocalMeta`, i.e. `self.locals.insert_with_default(p)`?

askama_shared/src/generator.rs
```diff
@@ -719,21 +719,61 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+                Expr::Attr(obj, attr) => {
+                    let mut attr_buf = Buffer::new(0);
+                    self.visit_attr(&mut attr_buf, obj, attr)?;
+
+                    let var = match self.locals.get(&attr_buf.buf.as_str()) {
+                        Some(meta) => match &meta.refs {
+                            Some(expr) => expr.clone(),
+                            None => name.to_string(),
+                        },
+                        _ => attr_buf.buf,
```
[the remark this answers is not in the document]
vallentin: Maybe that can turn into a `resolve` method after all 😅

askama_shared/src/generator.rs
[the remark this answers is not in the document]
djc: Err, yeah, of course.

vallentin:
I tried changing the `resolve` methods to return `Cow`. However, it would just introduce an additional clone for `Expr::Attr`. Further changing `LocalMeta` to use `Cow` made the rest more cumbersome.

djc (approved):

djc:
Looks good! In the future, please consider the following in terms of ordering:

* Items that depend on another item should appear before the method they depend on
* `&mut self` methods generally should be defined before `&self` methods

In the new `LocalMeta` methods, you did it the other way around.

vallentin:
Hmm, `LocalMeta` only has a single `&self` method. Are you talking about the `insert` and `insert_with_default` I added to `MapChain`? 😅 

> Items that depend on another item should appear before the method they depend on

So in relation to that, you referring to swapping `insert` and `insert_with_default` around?

> `&mut self` methods generally should be defined before `&self` methods

I assume, you're just mentioning this so I know for the future. Because a quick glance, and I don't see anywhere that I switched it around. To be fair, I didn't really think about that, because my personal preference for ordering, is different than that. 😅 

However, I'll try to remember that for the future.

djc:
Yeah, sorry, the `MapChain` methods (apparently I keep mixing these up!).

> So in relation to that, you referring to swapping `insert` and `insert_with_default around`?

Yup.

> I assume, you're just mentioning this so I know for the future. Because a quick glance, and I don't see anywhere that I switched it around. To be fair, I didn't really think about that, because my personal preference for ordering, is different than that. 😅

What is your preference for ordering, and why?

vallentin:
> Yup.

I'll try to remember this in the future. But otherwise, just tell me to reorder if I forget.

> What is your preference for ordering, and why?

Actually, "preference" might not be the correct word. Now that I'm thinking about it, it's probably more correct to say "habit" and "what I'm used too".

Generally, I order stuff by visibility, i.e. `pub`, `pub(crate)`, `pub(super)` and lastly private items. After that, I order methods in relation to their "kind", i.e. I usually do:


- Static methods
- Constructors, i.e. static methods returning some form of `Self` or `Result<Self, _>`
   - e.g. `new()`, `with_...()`
- Builder methods
- Methods that consume `self`, e.g. I have a `Worker` with a `join(self)` method for some multithreading stuff
- `&mut self` methods excluding setters
- Pairwise getter and setter for fields
- Other utility methods, like `is_empty`, `len`, etc
- Lastly private methods

I'd actually write use declarations in the opposite order (in e.g. [generator.rs]), i.e. `std`, then dependencies, then `crate::`.

[generator.rs]: https://github.com/djc/askama/blob/c7697cbd406dce0962618e99a6b78116b14fdb1c/askama_shared/src/generator.rs#L1-L15

However, ultimately I always attempt to follow the style of the repo I'm making a PR for. :)

djc:
That order is actually pretty close to what I'm aiming for, but I do like to order functions (and items in general) such that depending items come before depended upon items. Usually the former are more complex and need more changes and are more likely to be called on by "external" callers.

vallentin:
> such that depending items come before depended upon items

Of course, and honestly that actually makes a lot of sense in terms of readability. Say `insert` was sufficiently complex, you might not "notice" the subsequent small `insert_with_default`. But if it comes beforehand, then it _stands out more_. Though this is of course from the perspective of reading through the code.

My only _annoyance_ would be that in the documentation, I'd prefer `insert` to come before `insert_with_default`. However, rustdoc would follow the order of the items in the code. 😅

clouds56:
I'm not sure if this is the right place, but I cannot figure out how to make this run
```
{% for i in v.iter().skip(offset).take(size.clone()) %}
{% endfor %}
```
it would generate
```rust
for (i, _loop_item) in ::askama::helpers::TemplateLoop::new((&self.v.iter().skip(&self.offset).take(&{
    self.size.clone()}
)).into_iter())
```
I think it should not borrow when the expression is an function call

djc:
Have you tried with latest main?

vallentin:
Currently `v.iter().skip(offset).take(size.clone())` is not supported, neither the latest release nor the latest commit on main. The issue in particular is `.skip(offset)`, currently all fields and variables are implicitly referenced and cannot be deferenced.

This is related to "copy literals" (issue #404), and a possible solution and workaround should be available soon.

-----

If you use the latest commit on main, then there is a rather simple workaround, by using a method on your template.

```rust
impl SomeTemplate {
    fn iter_skip_take(&self) -> impl Iterator<Item = &...> {
        self.v.iter().skip(self.offset).take(self.size)
    }
}
```

```jinja
{% for i in self.iter_skip_take()  %}
{% endfor %}
```

```toml
[dependencies]
askama = { git = "https://github.com/djc/askama", rev = "b880799" }
```

PR #401, opened 2020-12-15
Improved template loop generation

### LINKED ISSUE #107: Cannot move out of borrowed content in for loop

### LINKED ISSUE #333: for-loop on values of a BTreeMap inside a template does not compile
Hi Askama team,

I'm hitting the an issue with for-loops inside template on `(some BTreeMap).values()`.

## Preliminary: what DOES work
Consider first the following piece of code, which works, demonstrating that the `values()` function of `BTreeMap` can be used inside an Askama template, as expected:

`main.rs`:
```rust
use std::collections::BTreeMap;
use askama::Template;

#[derive(Debug)]
struct Foo {
  foo: u64,
}

fn make_map() -> BTreeMap<u64, Foo> {
    let mut map : BTreeMap<u64, Foo> = BTreeMap::new();
    map.insert(23, Foo { foo: 51} );
    map
}

#[derive(Template)]
#[template(path = "template.txt")]
struct MyTemplate<'a> {
  map : &'a BTreeMap<u64, Foo>
}

fn main() {
    println!("Demo of weird stuff with BTreeMap in Askama template");
    let map = make_map();

    for foo in map.values() {
        println!("{:?}", foo)
    }

    let rendered = (MyTemplate {
        map : &map
    }).render().unwrap();
    println!("{}", rendered);
}
```

`template.txt`:
```
{{ map.values().count() }}
```

This works just fine, and the template renders the number of values, which is 1.

## But iterating does not work

Now add the following to the template:
```
{% for foo in map.values() %}
  {{ format!("{:?}", foo) }}
{% endfor %}
```

This won't compile:
> move occurs because value has type `std::collections::btree_map::Values<'_, u64, Foo>`, which does not implement the `Copy` trait

This surprises me, since in the main we also have a similar for-loop on the values, which work just fine.

## Workaround: collect references to values in a vector of references

I did find the following workaround: in the `MyTemplate` struct, add field
```rust
foos : &'a Vec<&'a Foo>
```

In `main.rs`, allocate a vector of references to the map values:
```rust
let foos : Vec<& Foo> = map.values().collect();
...
MyTemplate {
     ...,
     foos : &foos
}...
```

and iterate on `foos` in the template. This works, but is cumbersome.

Thanks in advance for considering this report!

vallentin:
Fixes #107
Fixes #333
Closes #221

No test cases were modified, and this is a non-breaking change.

This PR improves the generation of `TemplateLoop`, such that it has a higher chance to produce Rust code that will actually compile, based on assumptions that can be drawn from the context of the expression used. _See the comments in the code, for what those assumptions are._

-----

**Old Explanation (from #396 related to this change):**

> Based on the context of where it is used, what can it be assumed to be?
>    - With loops the assumption can be made that:
>      - `*Call(..)` and `Index(..)` likely return something that implements `IntoIterator` so don't borrow
>      - If not a `*Call(..)` but `self.` is used, then assume it needs to be borrowed
>      - Otherwise, assume that `IntoIterator` is implemented and the value can be moved

### DIFF
--- a/testing/tests/loops.rs
+++ b/testing/tests/loops.rs
@@ -67,3 +67,39 @@ fn test_for_range() {
         "foo (first)\nfoo (last)\nbar\nbar\nfoo\nbar\nbar\n"
     );
 }
+
+#[derive(Template)]
+#[template(source = "{% for i in [1, 2, 3] %}{{ i }}{% endfor %}", ext = "txt")]
+struct ForArrayTemplate {}
+
+#[test]
+fn test_for_array() {
+    let t = ForArrayTemplate {};
+    assert_eq!(t.render().unwrap(), "123");
+}
+
+#[derive(Template)]
+#[template(
+    source = "{% for i in [1, 2, 3].iter() %}{{ i }}{% endfor %}",
+    ext = "txt"
+)]
+struct ForMethodCallTemplate {}
+
+#[test]
+fn test_for_method_call() {
+    let t = ForMethodCallTemplate {};
+    assert_eq!(t.render().unwrap(), "123");
+}
+
+#[derive(Template)]
+#[template(
+    source = "{% for i in [1, 2, 3, 4, 5][3..] %}{{ i }}{% endfor %}",
+    ext = "txt"
+)]
+struct ForIndexTemplate {}
+
+#[test]
+fn test_for_index() {
+    let t = ForIndexTemplate {};
+    assert_eq!(t.render().unwrap(), "45");
+}
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -667,8 +667,26 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                 ", _loop_item) in ::askama::helpers::TemplateLoop::new({}) {{",
                 expr_code
             )),
+            Expr::Array(..) => buf.writeln(&format!(
+                ", _loop_item) in ::askama::helpers::TemplateLoop::new({}.iter()) {{",
+                expr_code
+            )),
+            // If `iter` is a call then we assume it's something that returns
+            // an iterator. If not then the user can explicitly add the needed
+            // call without issues.
+            Expr::MethodCall(..) | Expr::PathCall(..) | Expr::Index(..) => buf.writeln(&format!(
+                ", _loop_item) in ::askama::helpers::TemplateLoop::new(({}).into_iter()) {{",
+                expr_code
+            )),
+            // If accessing `self` then it most likely needs to be
+            // borrowed, to prevent an attempt of moving.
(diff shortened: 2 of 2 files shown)


vallentin:
Forgot to include "(fixes #107)" in the commit message, guess you can add it while merging :)

djc (commented):
For this change, it would actually be good just to add some simple test cases that exercise the new heuristics, for example, one test case per match arm you've added. Sound good?

I also prefer to rebase PRs, in which case I don't get to rewrite the commit message, so please just add the issue numbers. Should be easy with `git commit --amend && git push -f`.

vallentin:
I've added a few tests.

-----

I wanted to add a test using `repeat().take()` and/or `zip()`, but that won't work until I've made a PR (again) for `is_copyable` and not-borrowing literals. However, before I do that, I'll look into it a bit more, as I think there's a few cases that would be non-breaking changes.

```
#[derive(Template)]
#[template(
    source = "{% for i in ::std::iter::repeat("a").take(5) %}{{ i }}{% endfor %}",
    ext = "txt"
)]
struct ForPathCallTemplate {}

#[test]
fn test_for_path_call() {
    let t = ForPathCallTemplate {};
    assert_eq!(t.render().unwrap(), "aaaaa");
}
```

testing/tests/loops.rs
```diff
@@ -67,3 +67,39 @@ fn test_for_range() {
         "foo (first)\nfoo (last)\nbar\nbar\nfoo\nbar\nbar\n"
     );
 }
+
+#[derive(Template)]
+#[template(source = "{% for i in [1, 2, 3] %}{{ i }}{% endfor %}", ext = "txt")]
+struct ForArrayTemplate {}
```
vallentin, line 73: Note this one is similar to [`ArrayTemplate` in `simple.rs`][arr]. Though I feel like `simple.rs` is more of a "showcasing" of features and semantics, so should be keep both or remove `ForArrayTemplate`?

[arr]: https://github.com/djc/askama/blob/810d5ad5067085cc5e4be19903b5804b1ff13430/testing/tests/simple.rs#L252-L263

djc: Let's keep both.

djc (approved):
This is awesome, thanks!

PR #402, opened 2020-12-15
Added set alias for let

vallentin:
Related to #329

Added a `set` alias for `let` to be more in line with Jinja.


djc (commented):
Thanks for doing this! Let's also add a note in the book mentioning this is supported as an alternative, with the exact same semantics as `let`?

vallentin:
> add a note in the book

I've added a paragraph :)

There's a few issues in relation to the book. So I had just added it to the back burner, to update and include the `set` alias.

djc (approved):
👍

PR #408, opened 2020-12-18
Added support for parsing nested comments

vallentin:
I added support for nested comments, i.e. `{# foo {# bar #} baz #}`. Now Askama is more on par with Rust.

I sometimes comment out chunks of a template when testing stuff. If that chunk already contains comments, then that becomes harder.

- [x] Updated parser
- [x] Added comments test case
- [x] Updated book

Assuming no one has comments like `{# {# #}`, then this is a non-breaking change.

### DIFF
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -2,7 +2,7 @@ use nom::branch::alt;
 use nom::bytes::complete::{escaped, is_not, tag, take_until};
 use nom::character::complete::{anychar, char, digit1};
 use nom::combinator::{complete, map, opt, value};
-use nom::error::ParseError;
+use nom::error::{Error, ParseError};
 use nom::multi::{many0, many1, separated_list0, separated_list1};
 use nom::sequence::{delimited, pair, tuple};
 use nom::{self, error_position, Compare, IResult, InputTake};
@@ -1007,19 +1007,37 @@ fn block_node<'a>(i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], Node<'a>>
     Ok((i, contents))
 }
 
+fn block_comment_body<'a>(mut i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], &'a [u8]> {
+    let mut level = 0;
+    loop {
+        let (end, tail) = take_until(s.comment_end)(i)?;
+        match take_until::<_, _, Error<_>>(s.comment_start)(i) {
+            Ok((start, _)) if start.as_ptr() < end.as_ptr() => {
+                level += 1;
+                i = &start[2..];
+            }
+            _ if level > 0 => {
+                level -= 1;
+                i = &end[2..];
+            }
+            _ => return Ok((end, tail)),
+        }
+    }
+}
+
 fn block_comment<'a>(i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], Node<'a>> {
     let mut p = tuple((
         |i| tag_comment_start(i, s),
         opt(tag("-")),
-        take_until(s.comment_end),
+        |i| block_comment_body(i, s),
         |i| tag_comment_end(i, s),
     ));
-    let (i, (_, pws, inner, _)) = p(i)?;
+    let (i, (_, pws, tail, _)) = p(i)?;
     Ok((
         i,
         Node::Comment(WS(
             pws.is_some(),
-            inner.len() > 1 && inner[inner.len() - 1] == b'-',
+            tail.len() > 1 && tail[tail.len() - 1] == b'-',
         )),
     ))
 }
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -1329,6 +1329,23 @@ mod tests {
             )],
         );
     }
+
+    #[test]
+    fn test_parse_comments() {
+        let s = &Syntax::default();
+        assert_eq!(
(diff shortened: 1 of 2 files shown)


askama_shared/src/parser.rs
```diff
@@ -1011,15 +1011,33 @@ fn block_comment<'a>(i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], Node<'
     let mut p = tuple((
         |i| tag_comment_start(i, s),
         opt(tag("-")),
-        take_until(s.comment_end),
+        |mut i| {
```
djc, line 1014: Let's pull this out into a named function, please?

askama_shared/src/parser.rs
```diff
@@ -1011,15 +1011,33 @@ fn block_comment<'a>(i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], Node<'
...
+                let nested_start: IResult<&[u8], _> = take_until(s.comment_start)(i);
+
+                match nested_start {
+                    Ok((start, _)) if start.as_ptr() < end.as_ptr() => {
+                        i = block_comment(start, s)?.0;
+                    }
+                    _ => {
+                        i = end;
+                        before_end = inner;
+                        break;
```
djc, line 1027: Could just replace `break` with `return Ok((end, inner));` here?

askama_shared/src/parser.rs
```diff
@@ -1011,15 +1011,33 @@ fn block_comment<'a>(i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], Node<'
...
+                    }
+                }
+            }
+            Ok((i, before_end))
+        },
         |i| tag_comment_end(i, s),
     ));
-    let (i, (_, pws, inner, _)) = p(i)?;
+    let (i, (_, pws, before_end, _)) = p(i)?;
```
djc, line 1035: This seems like an unrelated change, let's pull it out into a separate commit. (I'm not convinced of its merits, TBH.)

djc (commented):
Looks great, some style nits.

askama_shared/src/parser.rs
vallentin: Previously `inner` contained the whole comment. Now it only contains the last part "before the end". So for `{# foo #}` then `inner` would contain ` foo `. However for `{# foo {# bar #} baz #}` it would only contain ` bar `.

So I felt it was more appropriate to rename it to `before_end` instead of `inner`. To avoid future confusion.

askama_shared/src/parser.rs
vallentin: Actually had that originally, and removed it 😅

askama_shared/src/parser.rs
djc: Okay, let's call it `tail`.

askama_shared/src/parser.rs
djc: Why? :)

askama_shared/src/parser.rs
djc: > However for `{# foo {# bar #} baz #}` it would only contain `bar`.

Did you mean it would only contain `baz`?

askama_shared/src/parser.rs
vallentin: Personally, I prefer a function because readability. But I removed it, because I was in the mindset of minimizing the change. However, I'm adding it again 😅

askama_shared/src/parser.rs
vallentin: Yes, of course. Typo 😅

askama_shared/src/parser.rs
```diff
@@ -1007,19 +1007,33 @@ fn block_node<'a>(i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], Node<'a>>
     Ok((i, contents))
 }
 
+fn block_comment_body<'a>(mut i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], &'a [u8]> {
+    loop {
+        let (end, inner) = take_until(s.comment_end)(i)?;
```
djc, line 1012: Can we transfrom the recursion to just iteration? I generally prefer to avoid recursion where possible.

Also let's not name the `nested_start` value, we can just match it directly?

vallentin: It's a bit cumbersome to change it to iteration. Because then we'd need e.g. a `Vec` or `counter` to handle nested comments in nested comments. In this case I don't think recursion, is that bad. Because what are the odds that people have many levels of nested comments, that will actually cause significant issues?

> can just match it directly

Yes, but it needs a type hint. So then it will need to look like this:

```rust
match take_until::<_, _, Error<&[u8]>>(s.comment_start)(i) {
```

djc: Having a counter seems fine with me? I'm not that worried about stack overflows, I just find recursion usually harder to reason about.

I still prefer not naming the expression, even if it involves turbofishing. :)

vallentin: Using a `counter` was simpler than expected 😅

djc (approved):
Thanks for obliging me! Feel free to merge once CI is done.

vallentin:
No problem! Just as a note, I haven't forgotten about #396, I just need to look more into the `get_skip` stuff.

djc:
No problem!

PR #409, opened 2020-12-21
Disable askama's default features in web framework crates

Randati:
Currently it's not possible to disable `askama`'s default features when one of the web framework specific crates is used.

If `Cargo.toml` looks like this, `askama_tide` will enable `askama`'s default features.
```toml
askama = { version = "*", default-features = false, features = ["with-tide"]}
askama_tide = "*"
```

This change will disable `askama`'s default features in `askama_{actix,gotham,iron,rocket,tide,warp}`.

As far as I know, this should only break code that disables default features, but uses them anyway.

### DIFF
--- a/askama_actix/Cargo.toml
+++ b/askama_actix/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 
 [dependencies]
 actix-web = { version = "3", default-features = false }
-askama = { version = "0.10", path = "../askama", features = ["with-actix-web", "mime", "mime_guess"] }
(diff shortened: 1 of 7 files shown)


Randati:
This has some changes to how some tests/clippy are run, since this change surfaced an existing clippy problem in a `#[cfg(not(feature = "config"))]` code path. But I believe that this increased the code coverage as the `#[cfg(feature = "config")]` path is tested too.

djc (approved):
LGTM, thanks!

PR #411, opened 2020-12-23
Added support for shadowing variables

### LINKED ISSUE #329: Add support for redefining the same variable with `let` (just as Rust permits)

vallentin:
Fixes #329

Considering that it wasn't previously possible to shadow variables. Then this qualifies as a non-breaking change, as it only extends new functionality.

- [x] Updated generator, specifically adding `initialized` to `LocalMeta`
- [x] Added test case
- [x] Updated book

In short, if a `LocalMeta` is `initialized` then any subsequent `{% let ... %}` shadows and generates a new variable. The variable is literally redefined by `insert`ing a new one. Thereby making it possible to e.g. initialize and shadow in multiple nested branches.

To be consistent with Rust, `{% let (a, b) = ... %}` always shadows, and can't be used to assign (which it couldn't before either.)


askama_shared/src/generator.rs
```diff
@@ -876,16 +876,34 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
 
         match *var {
             Target::Name(name) => {
-                if !self.locals.contains(&name) {
+                let meta = self.locals.get(&name).cloned();
```
djc, line 879: Why do we have `.cloned()` here? It doesn't seem necessary...

djc:
Looks good, just a minor question!

askama_shared/src/generator.rs
vallentin: Otherwise the subsequent `self.write_buf_writable(buf)` is not possible, as `self` would be borrowed.

Though I guess I can move `meta.is_none()` and turn it into a variable, next to `let shadowed` if you'd rather have that? :)

djc: Let's just keep it. 👍

djc (approved):

PR #412, opened 2020-12-23
WIP: make askama_rocket and askama_derive ready for rocket 0.5

flo-l:
I tried it locally, seems to work for me. But I'm not sure if I got the lifetimes right..

Anyhow, this PR should only be merged after Rocket releases 0.5.

Thank you for askama, I like it a lot :)

### DIFF
--- a/askama_rocket/Cargo.toml
+++ b/askama_rocket/Cargo.toml
@@ -15,4 +15,4 @@ edition = "2018"
 
 [dependencies]
 askama = { version = "0.10", path = "../askama", default-features = false, features = ["with-rocket", "mime", "mime_guess"] }
-rocket = { version = "0.4", default-features = false }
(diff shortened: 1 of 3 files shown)


djc:
Can you use a git dependency on Rocket here, so we can actually test this in CI?

shritesh:
FWIW, I've pointed Rocket to git and updated the test to use the new API at https://github.com/shritesh/askama/commit/b9f7a58cee2be9beb52449bedab06ee685287b06. 
Thank you for the PR.

LEXUGE:
As a remainder, the rocket has recently released 0.5 rc release, it might be a good point to update `askama_rocket` as well.

djc:
If someone is more invested than this PR's author, maybe just start a new PR?

djc:
This was pulled into #495, closing.

PR #407, opened 2020-12-18
Some README updates (fixes #406)

djc:

PR #416, opened 2020-12-26
Updated book to be on par with readme

vallentin:
Applied the changes from #407 to the book. Additionally, removed the 2 dead link references.


djc (approved):
Nice, thanks!

PR #418, opened 2020-12-29
Added error to hint enabling serde features

vallentin:
If `serde-json` or `serde-yaml` is not enabled, while attempting to use the `json` or `yaml` filters, then now an error suggesting to enable them appear. This would help people that don't realize you have to enable the feature(s).

Example:

```jinja
{{ "foo"|json }}
```

Before:

```text
error[E0425]: cannot find function `json` in module `askama::filters`
```

After:

```text
error: the `json` filter requires the `serde-json` feature to be enabled
```

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -1184,6 +1184,15 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             return Ok(DisplayWrap::Unwrapped);
         }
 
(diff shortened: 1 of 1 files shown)


djc (approved):
Nice!

vallentin:
This is of course a non-breaking change, given that `"json"` and `"yaml"` always resolve to `::askama::filters` and never `filters::`.

PR #419, opened 2020-12-29
Fixed yaml filter generation

vallentin:
While doing #418, I saw that `"yaml"` was missing from `visit_filter`.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -1193,7 +1193,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             return Err("the `yaml` filter requires the `serde-yaml` feature to be enabled".into());
         }
 
-        const FILTERS: [&str; 4] = ["safe", "escape", "e", "json"];
+        const FILTERS: [&str; 5] = ["safe", "escape", "e", "json", "yaml"];
         if FILTERS.contains(&name) {
             buf.write(&format!(
                 "::askama::filters::{}({}, ",
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -1193,7 +1193,8 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             return Err("the `yaml` filter requires the `serde-yaml` feature to be enabled".into());
         }
 
-        if name == "escape" || name == "safe" || name == "e" || name == "json" {
+        const FILTERS: [&str; 4] = ["safe", "escape", "e", "json"];
+        if FILTERS.contains(&name) {
             buf.write(&format!(
                 "::askama::filters::{}({}, ",
                 name, self.input.escaper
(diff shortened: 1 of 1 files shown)


askama_shared/src/generator.rs
```diff
@@ -1184,7 +1184,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             return Ok(DisplayWrap::Unwrapped);
         }
 
-        if name == "escape" || name == "safe" || name == "e" || name == "json" {
+        if name == "escape" || name == "safe" || name == "e" || name == "json" || name == "yaml" {
```
djc, line 1187: I think we should probably extract this list out into a constant or function or something, to avoid the repetition. Would you mind adding that in a separate commit?

vallentin: Just to be clear, you mean doing e.g. `FILTERS.contains(&name)`, right?

djc: Yup. Though this would be a newly defined `FILTERS`, right?

vallentin:
In Jinja, the `json` filter is actually called [`tojson`]. Should we add an alias?

[`tojson`]: https://jinja.palletsprojects.com/en/2.11.x/templates/#tojson

djc:
Adding an alias is fine with me, sure.

askama_shared/src/generator.rs
```diff
@@ -1184,7 +1184,8 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             return Ok(DisplayWrap::Unwrapped);
         }
 
-        if name == "escape" || name == "safe" || name == "e" || name == "json" {
+        const FILTERS: [&str; 5] = ["safe", "escape", "e", "json", "yaml"];
+        if FILTERS.contains(&name) {
```
vallentin, line 1188: or did you want it in `filters`, like with `filters::BUILT_IN_FILTERS`?

askama_shared/src/generator.rs
```diff
@@ -1197,13 +1198,11 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
-                DisplayWrap::Wrapped
-            } else {
-                DisplayWrap::Unwrapped
-            },
-        )
+        Ok(if FILTERS.contains(&name) {
+            DisplayWrap::Wrapped
+        } else {
+            DisplayWrap::Unwrapped
+        })
```
djc, line 1205: I usually write these as a `match` these days, since they're more compact and even a little bit easier to read IMO. What do you think?

djc (approved):

askama_shared/src/generator.rs
djc: Nah, I think this is fine for now.

askama_shared/src/generator.rs
vallentin: So like this? Either is fine with me, so if you prefer with `match` then sure :)

```rust
Ok(match FILTERS.contains(&name) {
    true => DisplayWrap::Wrapped,
    false => DisplayWrap::Unwrapped,
})
```

djc: Yup!

vallentin:
Done and conflicts resolved :)

djc (approved):

PR #420, opened 2020-12-29
Added tojson alias

vallentin:
Added [`tojson`] alias to be on par with Jinja.

[`tojson`]: https://jinja.palletsprojects.com/en/2.11.x/templates/#tojson

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -1170,7 +1170,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     fn visit_filter(
         &mut self,
         buf: &mut Buffer,
-        name: &str,
+        mut name: &str,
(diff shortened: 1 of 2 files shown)


vallentin:
If you prefer `name` to remain immutable, I can change the `if` to:

```rust
let name = match name {
    "tojson" => "json",
    _ => name,
};
```

vallentin:
Also, I know the language tags _doesn't do anything_, given that mdbook doesn't support Jinja.

However, it is possible to add support to highlighting other languages. So I just add them whenever I encounter them, to make my Markdown linter stop screaming at me.

I just need to check, if it's possible to avoid having to include a copy of highlight.js in the repo. To be able to add Jinja highlighting.

djc (approved):
Looks fine as is!

PR #421, opened 2021-01-01
Resolved Clippy lints

vallentin:
The [Rust 1.49.0 release] yesterday, or specifically the [Clippy release] with the addition of [`clippy::unnecessary-cast`] caused the scheduled CI task to fail.

[Rust 1.49.0 release]: https://blog.rust-lang.org/2020/12/31/Rust-1.49.0.html
[Clippy release]: https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-149
[`clippy::unnecessary-cast`]: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast

### DIFF
--- a/askama_shared/src/filters/mod.rs
+++ b/askama_shared/src/filters/mod.rs
@@ -518,9 +518,9 @@ mod tests {
     #[test]
     #[allow(clippy::float_cmp)]
     fn test_into_f64() {
-        assert_eq!(into_f64(&1).unwrap(), 1.0 as f64);
-        assert_eq!(into_f64(&1.9).unwrap(), 1.9 as f64);
-        assert_eq!(into_f64(&-1.9).unwrap(), -1.9 as f64);
+        assert_eq!(into_f64(&1).unwrap(), 1.0_f64);
+        assert_eq!(into_f64(&1.9).unwrap(), 1.9_f64);
+        assert_eq!(into_f64(&-1.9).unwrap(), -1.9_f64);
         assert_eq!(into_f64(&(INFINITY as f32)).unwrap(), INFINITY);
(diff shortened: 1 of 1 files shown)


askama_shared/src/filters/mod.rs
```diff
@@ -518,21 +518,21 @@ mod tests {
     #[test]
     #[allow(clippy::float_cmp)]
     fn test_into_f64() {
-        assert_eq!(into_f64(&1).unwrap(), 1.0 as f64);
-        assert_eq!(into_f64(&1.9).unwrap(), 1.9 as f64);
-        assert_eq!(into_f64(&-1.9).unwrap(), -1.9 as f64);
+        assert_eq!(into_f64(&1).unwrap(), 1.0_f64);
```
djc, line 521: I know clippy suggested the `_f64` suffix, but I think type suffixes without underscores are more idiomatic. What do you think?

djc (commented):
Thanks for fixing these! (This is exactly why I have scheduled CI run on Friday mornings... I was fixing up Quinn and trust-dns first.)

askama_shared/src/filters/mod.rs
vallentin: If you prefer without, then I can remove the underscores. :)

In this case I actually prefer with. It's less because of those cases, but more in the case of hexadecimals, e.g. `0xabcdi32` vs `0xabcd_i32`. So for consistency, I always use underscore.

In my own projects I even enforce some pedantic lints, so I'm used to underscores in literals given [`unreadable_literal`](https://rust-lang.github.io/rust-clippy/master/index.html#unreadable_literal).

djc (approved):

PR #422, opened 2021-01-01
Updated workflow to show each command

vallentin:
I've changed the multi-line commands to single-line commands, such that each get their own collapsible section.

When I went to fix #421, it caught me off guard, that it was `cargo fmt` that failed, as instead of inspecting it closer, I just jumped to my own terminal and did `cargo fmt` yielding no issues. You'd have to expand twice to be able to see what commands that step includes.

| Closed | Expanded |
| :------: | :---------: |
| ![closed] | ![expanded] |

[closed]: https://user-images.githubusercontent.com/17464404/103444285-6bde3800-4c67-11eb-817a-5ff57f94f956.png
[expanded]: https://user-images.githubusercontent.com/17464404/103444297-7d274480-4c67-11eb-8b47-697cb8fcb239.png

Whereas now, it will separate e.g. `cargo fmt` and `cargo clippy`, making it more apparent which one failed.

image

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -45,9 +45,8 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
-      - run: |
-          cargo test --package askama_actix --all-targets
-          cargo clippy --package askama_actix --all-targets -- -D warnings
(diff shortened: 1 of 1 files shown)


djc (approved):
Fair point!

PR #423, opened 2021-01-02
Removed implicit borrowing of literals, calls, and more

### LINKED ISSUE #404: Version 0.10 places Copy value literals behind a reference where 0.9 did not
I have a large collection of templates that compiles perfectly fine with version 0.9 of askama, but when I try to upgrade to 0.10, I get a bunch of type errors because a bunch of `Copy` types that were previously passed as values now get placed behind a reference

vallentin:
Fixes #404
Partially fixes the [code commented](https://github.com/djc/askama/pull/396#issuecomment-752028318) by @clouds56 (i.e. it requires using `offset.clone()`)

The `truncate`, `indent`, `into_f64`, and `into_isize` filters have been changed from taking `&usize` to `usize`.
This is a partial breaking change, as while e.g. `{{ "hello"|truncate(3) }}` is still valid.

Then given that fields are still implicitly borrowed, then `{{ "hello"|truncate(self.len) }}` would no longer be valid. However, given that calls aren't borrowed implicitly, then now explicitly cloning e.g. `{{ "hello"|truncate(self.len.clone()) }}` is possible as a workaround.

Which also means the comment by @clouds56 is possible. Before neither `{% for i in v.iter().skip(5) %}` nor `{% for i in v.iter().skip(self.offset.clone()) %}` were possible. Now both are possible!

Related: I added the [test case previously mention] in #401.

[test case previously mention]: https://github.com/djc/askama/pull/401#issuecomment-745440470

------

If desired, we could add special cases to `visit_filter`/`_visit_args` to allow, e.g. `{{ "hello"|truncate(self.len) }}`. Thus that in those cases, using fields won't be referenced.

------

The only remaining test case that broke by this change, was [`test_root_path_func_call`], which I added a month ago in #393.

[`test_root_path_func_call`]: https://github.com/djc/askama/blob/c29ecd68714bddf5e27a9e347c902faa23b2a545/testing/tests/simple.rs#L293-L300

This doesn't work, given that `ToString::to_string` expects e.g. `&i32` while it would be given a `i32`.

```jinja
{{ ::std::string::ToString::to_string(123) }}
```

However, there is a workaround that does work within the template itself, by defining a temporary variable.

```jinja
{% let i = 123 %}
{{ ::std::string::ToString::to_string(i) }}
```

-----

Side note, it wasn't possible to use the [`abs` filter] before, given that expects an owned value. No change is needed to make it work now.

[`abs` filter]: https://github.com/djc/askama/blob/c29ecd68714bddf5e27a9e347c902faa23b2a545/askama_shared/src/filters/mod.rs#L318-L325

### DIFF
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -51,6 +51,45 @@ pub enum Expr<'a> {
     RustMacro(&'a str, &'a str),
 }
 
+impl Expr<'_> {
+    /// Returns `true` if enough assumptions can be made,
+    /// to determine that `self` is copyable.
+    pub fn is_copyable(&self) -> bool {
+        self.is_copyable_within_op(false)
+    }
+
+    fn is_copyable_within_op(&self, within_op: bool) -> bool {
+        use Expr::*;
+        match self {
+            BoolLit(_) | NumLit(_) | StrLit(_) | CharLit(_) => true,
+            Unary(.., expr) => expr.is_copyable_within_op(true),
+            BinOp(_, lhs, rhs) => {
+                lhs.is_copyable_within_op(true) && rhs.is_copyable_within_op(true)
+            }
+            // The result of a call likely doesn't need to be borrowed,
+            // as in that case the call is more likely to return a
+            // reference in the first place then.
+            VarCall(..) | PathCall(..) | MethodCall(..) => true,
+            // If the `expr` is within a `Unary` or `BinOp` then
+            // an assumption can be made that the operand is copy.
+            // If not, then the value is moved and adding `.clone()`
+            // will solve that issue. However, if the operand is
+            // implicitly borrowed, then it's likely not even possible
+            // to get the template to compile.
+            _ => within_op && self.is_attr_self(),
+        }
+    }
+
+    /// Returns `true` if this is an `Attr` where the `obj` is `"self"`.
+    pub fn is_attr_self(&self) -> bool {
+        match self {
+            Expr::Attr(obj, _) if matches!(obj.as_ref(), Expr::Var("self")) => true,
+            Expr::Attr(obj, _) if matches!(obj.as_ref(), Expr::Attr(..)) => obj.is_attr_self(),
+            _ => false,
+        }
+    }
+}
+
 pub type When<'a> = (
     WS,
     Option<MatchVariant<'a>>,
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -1132,6 +1132,19 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_numbers() {
+        let syntax = Syntax::default();
+        assert_eq!(
+            super::parse("{{ 2 }}", &syntax).unwrap(),
+            vec![Node::Expr(WS(false, false), Expr::NumLit("2"),)],
+        );
+        assert_eq!(
+            super::parse("{{ 2.5 }}", &syntax).unwrap(),
+            vec![Node::Expr(WS(false, false), Expr::NumLit("2.5"),)],
+        );
+    }
+
+    #[test]
     fn test_parse_var_call() {
         assert_eq!(
             super::parse("{{ function(\"123\", 3) }}", &Syntax::default()).unwrap(),
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -1,7 +1,7 @@
 use nom::branch::alt;
 use nom::bytes::complete::{escaped, is_not, tag, take_until};
 use nom::character::complete::{anychar, char, digit1};
-use nom::combinator::{complete, map, opt, value};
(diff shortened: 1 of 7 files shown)


askama_shared/src/parser.rs
```diff
@@ -51,6 +51,44 @@ pub enum Expr<'a> {
...
+    /// Returns `true` if enough assumptions can be made,
+    /// to determine that `self` is copyable.
+    pub fn is_copyable(&self) -> bool {
+        fn is_copyable(expr: &Expr, within_op: bool) -> bool {
+            use Expr::*;
+            match expr {
+                BoolLit(_) | NumLit(_) | StrLit(_) | CharLit(_) => true,
+                Unary(.., expr) => is_copyable(expr, true),
+                BinOp(_, lhs, rhs) => is_copyable(lhs, true) && is_copyable(rhs, true),
+                // FIXME: The parser should produce `NumLit(_)` instead of `Attr(NumLit(_), _)` for floats
```
vallentin, line 64: Should I just resolve this in this PR? As in a prior commit.

vallentin:
Alternatively, we could also just force that `_visit_args` always borrows when used from the context of `visit_filter`. However, I haven't looked into that, so unsure whether or how much it might complicate the logic.

Pro: That way there would be no breaking changes.
Con: It might confuse users when reasoning about the implicit borrowing.

askama_shared/src/parser.rs
djc: Yeah, that would be great!

askama_shared/src/parser.rs
```diff
@@ -51,6 +51,44 @@ pub enum Expr<'a> {
     RustMacro(&'a str, &'a str),
 }
 
+impl Expr<'_> {
+    /// Returns `true` if enough assumptions can be made,
+    /// to determine that `self` is copyable.
+    pub fn is_copyable(&self) -> bool {
+        fn is_copyable(expr: &Expr, within_op: bool) -> bool {
```
djc, line 58: I don't really like the nested method here. Let's just have two methods on `Expr`, with the "inner" one called `is_copyable_within_op()` (and private)?

djc (commented):
IMO calling `to_string()` as a top-level expression is pretty pointless, since expressions will rely on `Display` anyway (so it's unnecessary). 

This workaround isn't very discoverable, I think, so it doesn't seem very useful:

```
{% let i = 123 %}
{{ ::std::string::ToString::to_string(i) }}
```

I think this is a good change, though I still feel it probably warrants a breaking version number.

> Alternatively, we could also just force that _visit_args always borrows when used from the context of visit_filter. However, I haven't looked into that, so unsure whether or how much it might complicate the logic.
> 
> Pro: That way there would be no breaking changes.
> Con: It might confuse users when reasoning about the implicit borrowing.

Yeah, it's clever but it's probably too hard to reason about.

askama_shared/src/parser.rs
vallentin: Sure. I sometimes still write it like that, as in the past I encountered a few cases where it would produce more efficient binaries in the end. Unsure if that's still the case. But yeah, in this case it doesn't really matter. :)

vallentin:
> IMO calling `to_string()` as a top-level expression is pretty pointless, since expressions will rely on `Display` anyway (so it's unnecessary).

I agree, I just took the example from the test case. The same would apply for user defined function, e.g. `foo(i: &i32)`, then that now would fail with `{{ self::foo(123) }}`. But a workaround would be `{% let i = 123 %} {{ self::foo(i) }}`. However, in this case it being a user defined function, would make it a lot easier to just change `&i32` ti `i32`.

> I think this is a good change, though I still feel it probably warrants a breaking version number.

Should we get a patch release out first, with all the prior non-breaking changes?

> Yeah, it's clever but it's probably too hard to reason about.

It would boil down to that filters borrow literals/everything, whereas _anything else_ doesn't borrow literals.

On one hand, I agree that it would be confusing. On the other hand, it technically only means that it's the behavior of e.g. function calls that have been changed to not borrow literals.

But yes, that would still be confusing and potentially inconsistent behavior that a user would have trouble reasoning with.

djc:
How much stuff do you still have on your list of things to fix? I'm inclined to just keep going for a while -- I don't think "breaking" releases are that bad, especially if they don't actually break much in practice.

vallentin:
When this is merged, I can finally delete my fork, as this is the last thing committed on my fork! 😅 

> How much stuff do you still have on your list of things to fix?

I'll need to sort out my notes. It's a bit jumbled right now, between issues, stuff that can be refactored, and missing Jinja features.

For instance, sent you a message in relation to the `~` operator. This could easily be added, and wouldn't be a breaking change of course.

> I don't think "breaking" releases are that bad

Sure. I meant more, if you wanted the patches and minor features added first. Then we could wait with merging this. Though I'm fine with either.

djc (approved):

PR #426, opened 2021-01-05
 Fixed precedence of filters and unary operators

### LINKED ISSUE #424: Filters have precedence over unary operators
Performing e.g. `{{ -2|abs }}` is parsed as `{{ -(2|abs) }}` resulting in `-2` being rendered.

vallentin:
Fixes #424 and improves the filters parser test case


djc (approved):
LGTM!

PR #432, opened 2021-01-11
Updated book to include urlencode and urlencode_strict

vallentin:
Forgot to include these in #429.

### DIFF
--- a/book/src/filters.md
+++ b/book/src/filters.md
@@ -233,6 +233,34 @@ Output:
 HELLO
 ```
 
+### urlencode
+
+Percent-encodes the value for safe use in URI; does not encode `/`.
+
+```jinja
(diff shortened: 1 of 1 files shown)


book/src/filters.md
```diff
@@ -233,6 +233,34 @@ Output:
 HELLO
 ```
 
+### urlencode
+
+Percent-encodes the value for safe use in URI; does not encode `/`.
```
djc, line 238: The example here does encode the slash. Which is correct?

vallentin: I just it from the docs, and rendered the examples.

I'll look into whether `urlencode` has always been buggy, or if some upstream change caused it.

djc: It's probably due to https://github.com/djc/askama/pull/384.

PR #433, opened 2021-01-11
Changed so uppercase variables passthrough as constants

vallentin:
Resolves #432

Fixes so that variables including an uppercase letter, doesn't get prefixed `self.`, i.e. `{{ FOO }}` results in `FOO` instead of `self.FOO`.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -1721,8 +1721,11 @@ impl MapChain<'_, &str, LocalMeta> {
     }
 
     fn resolve_or_self(&self, name: &str) -> String {
-        self.resolve(name)
-            .unwrap_or_else(|| format!("self.{}", name))
+        match self.resolve(name) {
+            Some(name) => name,
+            None if name.chars().any(char::is_uppercase) => name.to_string(),
+            None => format!("self.{}", name),
+        }
     }
 }
 
--- a/testing/tests/simple.rs
+++ b/testing/tests/simple.rs
@@ -73,6 +73,36 @@ fn test_variables_no_escape() {
 }
 
 #[derive(Template)]
+#[template(
+    source = "{{ foo }} {{ foo_bar }} {{ FOO }} {{ FOO_BAR }} {{ self::FOO }} {{ self::FOO_BAR }} {{ Self::BAR }} {{ Self::BAR_BAZ }}",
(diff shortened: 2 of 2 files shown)


testing/tests/simple.rs
```diff
@@ -72,6 +72,36 @@ fn test_variables_no_escape() {
     );
 }
 
+const FOO: &str = "FOO";
+const FOO_BAR: &str = "FOO BAR";
```
vallentin, line 76: Decided to add the constants here, so they are near the test case they're used.

vallentin:
This indirectly also fixes so that `{{ None }}` results in `None` instead of `self.None` (that's why I only check for a single uppercase char).

However, `{{ Some("foo") }}` is also interpreted as a variable it seems. It at least incorrectly generates `(self.Some)("foo")`. I'll fix that in a future PR.

askama_shared/src/generator.rs
```diff
@@ -1722,7 +1722,10 @@ impl MapChain<'_, &str, LocalMeta> {
 
     fn resolve_or_self(&self, name: &str) -> String {
         self.resolve(name)
-            .unwrap_or_else(|| format!("self.{}", name))
```
djc, line 1725: I think this would be easier to follow as a three-arm match expression.

testing/tests/simple.rs
djc: It's fine to keep them close, but let's keep them below the other code related to this test.

askama_shared/src/generator.rs
vallentin: You mean like this?

```rust
match self.resolve(name) {
    Some(name) => name,
    None if name.chars().any(char::is_uppercase) => name.to_string(),
    None => format!("self.{}", name),
}
```

djc: Yup!

testing/tests/simple.rs
djc: Let's go with between `fn test_constants()` and `struct IfTemplate`.

vallentin: Unsure why that comment was first posted 3 hours after the other. I wrote them at the same time. However, both issues are now resolved 😅

djc (approved):

PR #434, opened 2021-01-13
I18n

jhoobergs:
This (draft) PR gives an initial onset for localization (#202) by using [fluent-templates](https://github.com/XAMPPRocky/fluent-templates) to statically load all fluent translations.
It is based on code of @kazimuth 

@djc Could you give some quick remarks / questions?

**TODO**
- [ ] Add `with-i18n` features
- [ ] Fix the test for invalid `text_id`'s (the test fails, as it should, but we should assert that it fails instead of failing)
- [ ] Look into a better way to check if translations exist (currently catches a panic of fluent-templates, maybe they should use a Result?)
- [ ] Documentation
- [ ] Improve `init_translations` macro
- [ ] Go through all remaining TODO comments

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -135,3 +135,61 @@ pub mod mime {
     note = "file-level dependency tracking is handled automatically without build script"
 )]
 pub fn rerun_if_templates_changed() {}
+
+pub trait Localizer {
+    fn get_fallback_language(&self) -> unic_langid::LanguageIdentifier;
+    fn get_language(&self) -> unic_langid::LanguageIdentifier;
+    fn get_loader(&self) -> &'static fluent_templates::StaticLoader;
+}
+
+#[macro_export]
+macro_rules! init_translation {
+    (
+        $v: vis $n: ident {
+            static_loader_name: $static_loader_name: ident,
+            locales: $locales: expr,
+            fallback_language: $fallback_language: expr,
+            customise: $customise: expr
+        }
+    ) => {
+        fluent_templates::static_loader! {
+            // Declare our `StaticLoader` named `LOCALES`.
+            static $static_loader_name = {
+                // The directory of localisations and fluent resources.
+                locales: $locales,
+                // The language to falback on if something is not present.
+                fallback_language: $fallback_language,
+                // Optional: A fluent resource that is shared with every locale.
+                //core_locales: "/core.ftl",
+                // Removes unicode isolating marks around arguments, you typically
+                // should only set to false when testing.
+                customise: $customise,
+            };
+        }
+        $v struct $n {
+            language: unic_langid::LanguageIdentifier,
+            loader: &'static fluent_templates::StaticLoader
+        }
+        impl $n {
+            pub fn new(language: unic_langid::LanguageIdentifier, loader: &'static fluent_templates::StaticLoader) -> $n {
+                $n {
+                    language,
+                    loader
+                }
+            }
+        }
+        impl askama::Localizer for $n {
+            fn get_fallback_language(&self) -> unic_langid::LanguageIdentifier {
+                unic_langid::langid!($fallback_language)
+            }
+
+            fn get_language(&self) -> unic_langid::LanguageIdentifier {
+                self.language.clone()
+            }
+
+            fn get_loader(&self) -> &'static fluent_templates::StaticLoader {
+                self.loader
+            }
+        }
+    }
+}
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -136,12 +136,6 @@ pub mod mime {
 )]
 pub fn rerun_if_templates_changed() {}
 
-pub trait Localizer {
-    fn get_fallback_language(&self) -> unic_langid::LanguageIdentifier;
-    fn get_language(&self) -> unic_langid::LanguageIdentifier;
-    fn get_loader(&self) -> &'static fluent_templates::StaticLoader;
-}
-
 #[macro_export]
 macro_rules! init_translation {
     (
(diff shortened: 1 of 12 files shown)


djc:
Looks like a good start! I wonder if we can simplify a bit, though. Can we make a template more like this?

```rust
#[derive(Template)]
#[template(path = "i18n.html")]
struct UsesI18n<'a> {
    #[locale]
    loc: unic_langid::LanguageIdentifier,
    name: &'a str,
    hours: f64,
}
```

Where the type of the field marked as locale must implement a `Localize` trait. Let's hard-code a directory name for now (what's the usual directory name that projects using Fluent use?), that can be customized through askama.toml. I'm also curious what other template solutions use to call Fluent-like messages. `localize(greetingsss, name: name)` looks okay to me, though it seems a little different from what we have so far (since we don't do named arguments in general). Does Fluent always require named arguments, or are positional arguments also used?

djc:
Also, thanks for working on this! I'm very excited to get this in!

jhoobergs:
> Looks like a good start! I wonder if we can simplify a bit, though. Can we make a template more like this?
> 
> ```rust
> #[derive(Template)]
> #[template(path = "i18n.html")]
> struct UsesI18n<'a> {
>     #[locale]
>     loc: unic_langid::LanguageIdentifier,
>     name: &'a str,
>     hours: f64,
> }
> ```
> 
> Where the type of the field marked as locale must implement a `Localize` trait. Let's hard-code a directory name for now (what's the usual directory name that projects using Fluent use?), that can be customized through askama.toml. I'm also curious what other template solutions use to call Fluent-like messages. `localize(greetingsss, name: name)` looks okay to me, though it seems a little different from what we have so far (since we don't do named arguments in general). Does Fluent always require named arguments, or are positional arguments also used?
> 
> (Since a language identifier will seemingly always just be parsed from a `&str`, we may at some point also want to allow `&str` as `#[locale]` attribute, taking care to bubble up errors from parsing the language identifier.)

How would we know to use `MyLocalizer` if it is not passed? Do you want to create a struct with a certain name that we can use? This would not allow having different Localizer's for different parts of the code. But this  is probably not really a problem / not a valid use case. Allowing only once Localizer, does also allow using the toml file for the fluent-folder path. As far as I can see, there is not really a default folder name. `i18n` seems good to me.

Fluent does not allow positional arguments.

Allowing a `&str` for the localizer, is indeed useful.

djc:
I guess my question is, what bits of state (other than the locale) does a `Localizer` need? And will most of these bits of state be per template or globally across a crate/project?

askama/src/lib.rs
```diff
@@ -135,3 +135,86 @@ pub mod mime {
...
+                // Optional: A fluent resource that is shared with every locale.
+                //core_locales: "/core.ftl",
+                // Removes unicode isolating marks around arguments, you typically
+                // should only set to false when testing.
+                customise: $customise,
+            };
+        }
+        $v struct $n {
+            language: unic_langid::LanguageIdentifier,
+            loader: &'static fluent_templates::once_cell::sync::Lazy<fluent_templates::StaticLoader>
```
jhoobergs, line 166: We need to be able to access this loader somehow. That's the reason for this struct.

askama/src/lib.rs
```diff
@@ -135,3 +135,86 @@ pub mod mime {
...
+        }
+    ) => {
+        use fluent_templates::Loader;
+        fluent_templates::static_loader! {
+            // Declare our `StaticLoader` named `LOCALES`.
+            static $static_loader_name = {
+                // The directory of localisations and fluent resources.
+                locales: $locales,
+                // The language to falback on if something is not present.
+                fallback_language: $fallback_language,
```
jhoobergs, line 156: The fallback language can be added to the toml file.

askama_shared/src/generator.rs
```diff
@@ -211,6 +215,53 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+
+
+                #[test]
+                fn test_i18n_default_coverage() {
+                    let messages = &[
+                        #(#messages),*
+                    ][..];
+    
+                    // create default localizer
+                    let localizer = super::#loc_ty::default();
```
jhoobergs, line 246: Here we need to be able to somehow access the static loader. Currently an object of the generated struct type is used for that.

askama_shared/src/generator.rs
```diff
@@ -1367,6 +1421,60 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+
+        assert!(
+            message.chars().find(|c| *c == '"').is_none(),
+            "message ids with quotes in them break the generator, please remove"
+        );
+
+        self.localized_messages.insert(message.clone());
+
+        buf.write(&format!(
+            "self.{}.translate(\"{}\", &std::iter::FromIterator::from_iter(vec![",
```
jhoobergs, line 1459: Here we need to be able to call the translate method. Currently it uses the Localizer object that has to be added to all Template structs.

jhoobergs (commented):
Basically: We need to initialize a static loader (with fallback language) and somehow have access to that loader in different parts of the code.

djc:
Huh, sorry I missed your responses on this -- are you still interested in working on this?

Having reviewed your code a bit more, how about having an Askama-provided type `Locale`, which is instantiated with a locale identifier and a reference to the `StaticLoader`. We then add a filter `i18n(&str)` which will expand to referencing the fields of the `Locale`?

jhoobergs:
@djc I would certainly be interested in working on this. I'm not completely sure if I understand your solution with the Askama-provided `Locale`-type. Could you write a simple example of the code that a user would write to use the localization? I guess this would consist of a rust struct, a template string and maybe some extra setup code?

djc:
Something like this:

```rust
#[derive(Template)]
#[template(source = "{{ localize(\"hello\", age: 37) }}")]
struct UsesI18n<'a> {
    #[locale]
    loc: askama::Locale::new(lang_id, &TEMPLATES),
}
```

with this 

```rust
struct Locale {
   ...
}

impl Locale {
     fn new(lang: unic_langid::LanguageIdentifier, templates: &fluent_templates::StaticLoader) { ... }
}
```

PR #450, opened 2021-02-17
Update askama_warp/README.md

freddyb:
The file calls itself the readme for askama_actix, but clearly it's warp.


djc (approved):
Oops! Thanks for fixing it.

Issue #451, opened 2021-02-18
Next release planning


djc:
@vallentin I forget what things I still wanted to resolve before releasing... Do you have anything else you would like to land in the short term?

vallentin:
Here's a few from my TODO.md:

- [ ] `{{ None }}` produces `self.None`
- [ ] `{{ Some("foo") }}` produces `(self.Some)("foo")`
- [ ] `{% let (x) = (2) %}` produces `let (x) = (2);`, which is semantically wrong and should be `let (x,) = (2);`
   - or `let x = (2);` of course
- [ ] `.zip(0..5)` produces `.zip(&0..5)` instead of `.zip(0..5)`
   - _Might have been fixed_

Then of course also doing the ones from #427, that we agreed on. Though overall I just wanna iron out the bugs.

Been a bit busy for the past month, but I think I'll be able to fix those bugs in the upcoming weekend.

djc:
Sounds good -- let's see where we are on Monday.

vallentin:
One of them was already fixed, but I've fixed the 3 other bugs.

I also discovered an additional bug (I've edited the previous comment).

djc:
I think it's past time we got this out. @vallentin when would you have time to investigate #494?

vallentin:
If it can wait until Thursday / the weekend, then yes :)

djc:
Yeah, that should be fine.

djc:
I'll see if I can spend some time on this tomorrow (mostly writing up a change log).

vallentin:
Was #494 the "last" important bug, before the new release? Just glancing through this issue and #427, then it seems the remaining stuff is new features.

When you start writing the change log, feel free to make a draft PR, then I'll comment any missing stuff I recall. I guess one of the most important things, is highlighting the change to the implicit borrowing. Mentioning that literal primitives and the result of functions are no longer borrowed.

djc:
At this point I'll probably wait for the nom 7 release, which should help with the silly dependencies.

https://github.com/Geal/nom/issues/1323#issuecomment-890846534

kristoff3r:
Since nom 7 has been released, is there anything else blocking a release?

djc:
I need to review/decide on #517 and #518 and then I need to write some release notes.

djc:
[...]

PR #452, opened 2021-02-22
Removed needless borrow of range

vallentin:
Before `.zip(10..20)` would result in `.zip(&(10..20))` which is not valid.

- Removed needless borrow of range
- Added test case for `(0..10).zip(10..20).zip(20..30)`

When adding the test case, I realized that Askama cannot parse `{% for ((i, j), k) in (0..10).zip(10..20).zip(20..30) %}`, specifically the for loop pattern `((i, j), k)`.

Reference #451

### DIFF
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -66,6 +66,7 @@ impl Expr<'_> {
             BinOp(_, lhs, rhs) => {
                 lhs.is_copyable_within_op(true) && rhs.is_copyable_within_op(true)
             }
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Thanks! I think there are likely cases where this is wrong (given that `Range` is unfortunately not `Copy`), but it likely makes sense for the common cases.

vallentin:
Yeah true. But I really hope people aren't doing _that_ kind of logic inside templates 😅

PR #453, opened 2021-02-22
Fixed path parser to account for single identifier type names

vallentin:
- Fixed `parser::path()` to account for `None` and `Some`. In short, a single identifier is considered a path, if it contains at least 1 uppercase and lowercase character
- Added parser tests for variables and paths
- Added generator test for `{{ Self::foo(None) }} {{ Self::foo(Some(2)) }}`

Reference #451

### DIFF
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -1226,6 +1226,60 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_var() {
+        let s = Syntax::default();
+
+        assert_eq!(
+            super::parse("{{ foo }}", &s).unwrap(),
+            vec![Node::Expr(Ws(false, false), Expr::Var("foo"))],
+        );
+        assert_eq!(
+            super::parse("{{ FOO }}", &s).unwrap(),
+            vec![Node::Expr(Ws(false, false), Expr::Var("FOO"))],
+        );
+
+        assert_eq!(
+            super::parse("{{ none }}", &s).unwrap(),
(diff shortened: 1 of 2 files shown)


vallentin:
While fixing this I was thinking about #433. Instead of having the following check, then we could e.g. change `Expr::Var(&'a str)` into `Expr::Var(&'a str, bool)`. The `bool` would then represent whether if it's a constant or not, and in short enforce that `self.` is not prepended.

https://github.com/djc/askama/blob/7ee368e57a419bf05826522df695ddd62d17684d/askama_shared/src/generator.rs#L1726

Note I didn't really look into it much, but I feel like using `bool` would be the smallest change. We could also refactor `Expr::Var` into `Expr::Field` and `Expr::Const`. Again, I didn't look into it, but the initial thought is that it would be a bigger change than adding a `bool`. Though I prefer the latter, as it is less ambiguous.

What do you think? Honestly, whether it's none, the former or the latter. It doesn't really matter, I'll refactor if you like the idea.

djc:
I like the idea of refactoring into `Field` and `Const` if you're up for it!

vallentin:
I assume you just want me to continue in this PR?

But yeah, I'll do it. I'll have time tonight or tomorrow.

djc:
I don't mind whether you want to do it here or start a new PR for it -- I haven't actually looked at this one yet, so feel free to do what is easiest for you.

vallentin:
Ah, in that case, just review this when you have time. Then I'll make a new PR after refactoring `Var` into `Field` and `Const`.

djc (approved):

Issue #395, opened 2020-12-07
Feature: handlebars frontend


djc:
In order to make Askama more accessible to folks used to handlebars templates (https://crates.io/crates/handlebars is likely the most Rust popular template engine), it would be awesome if Askama could parse (a sensible subset of) handlebars templates and compile/transpile it to the Askama AST, enabling people to use handlebars templates with Askama codegen.

harishperfect:
@djc Handlebars templates are more frontend based than lets say Jinja2. I'd say the functional features that Jinja offers outweighs what Handlebars does and offers. In my opinion, it seems like it would be best to continue more support on the functional aspects of Askama and Jinja. What do you think?

Also developers that use systems programming languages such as Rust might prefer to pre-render than to depend and offload to VanillaJS. With many markdown based static generators already available in Rust, do you still see Handlebars being a choice for Rust devs?

PR #454, opened 2021-02-23
Reworked constants to be parsed as paths

vallentin:
As discussed on #453.

- Refactored `Expr::Var` into `Expr::Var` and `Expr::Const`
- Updated parser test cases

I didn't do `Expr::Field` (yet?), because it would require moving a significant amount of the logic from the generator into the parser. Which first of all, I don't think that logic belongs in the parser. However, it might be possible to have an intermediate step between the parser and generator. To possibly have some simple form of IR. Essentially to be able to simplify what the generator receives and operates on, instead of AST.

That might even be useful if we wanted to add support for handlebars (#395)

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -1723,11 +1723,8 @@ impl MapChain<'_, &str, LocalMeta> {
     }
 
     fn resolve_or_self(&self, name: &str) -> String {
-        match self.resolve(name) {
-            Some(name) => name,
-            None if name.chars().any(char::is_uppercase) => name.to_string(),
-            None => format!("self.{}", name),
-        }
+        self.resolve(name)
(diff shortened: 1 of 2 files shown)


vallentin:
The back of my mind is telling me that it would probably be better to remove `Expr::Const` and just have all uppercase translate into `Expr::Path`.

djc:
> The back of my mind is telling me that it would probably be better to remove `Expr::Const` and just have all uppercase translate into `Expr::Path`.

I think that makes sense, yes.

djc (approved):
Seems like a nice simplification!

djc:
Maybe in the future wait for all the CI jobs to finish? 😛

vallentin:
Wait it wasn't 😳

I'm clearly seeing things then 😅

PR #458, opened 2021-03-09
Fixed extension to ignore common jinja extensions

### LINKED ISSUE #457: Proper file extension for templates for plugin integration?
First of all, let me congratulate you for your work. I'm developing an application with Askama and Rocket, and I'm very happy with the results. However, I recently hit an interesting bump. This was using VS Code as my development environment:

- As the templates are all HTML, I'm just using `.html` extension.
- The HTML formatting plugins don't recognize Askama templates. That means no syntax highlighting, no error detection, and no proper formatting. In fact one plugin completely mangled a few templates and caused my website to crash.
- As Askama is based on Jinja2, I tried to rename my files as `.html.j2` and use a Jinja2 plugin (no Askama plugin seems to exist), and lo and behold, it worked perfectly. Syntax highlighting and proper formatting. The code compiles and runs flawlessly, but...
- Askama's Rocket responder implementation guesses the content type from the file extension, but `.html.j2` is not recognized, so it just returns an empty response! No error message, no nothing, just a totally blank page.

So my question is, how can we fix this? I have a couple ideas:

- Write a proper Askama plugin with its own file extension, or reuse an existing Jinja2 plugin, even if the two syntaxes don't match perfectly?
- Fix the responders implementations to strip the last part of the extension from the file name, in case it matches a predefined list.

vallentin:
Fixes #457

- Fixed so `extension()` for `foo.html.{jinja,jinja2,j2}` resolves to `html` instead of `{jinja,jinja2,j2}`.
- Changed to allow both `path` and `ext`, i.e. so it would allow `#[template(path = "template.foo.bar.baz", ext = "html")]`

_This was just a quick fix, so I hope I didn't overlook anything_

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -156,10 +156,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         buf.writeln("}")?;
 
         buf.writeln("fn extension(&self) -> Option<&'static str> {")?;
-        buf.writeln(&format!(
-            "{:?}",
-            self.input.path.extension().map(|s| s.to_str().unwrap())
-        ))?;
+        buf.writeln(&format!("{:?}", self.input.extension()))?;
         buf.writeln("}")?;
 
         buf.writeln("fn size_hint(&self) -> usize {")?;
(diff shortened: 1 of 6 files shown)


vallentin:
Just to be clear, so first commit is just `fn extension()` where it just returns `self.path.extension()`. The second commit is then changing it, to account for `.jinja`, etc.

Good idea with the tests. I can add a private function `fn extension(&Path)` in `input.rs`, and then test that function. Is that fine?

vallentin:
Unsure if I'm misinterpreting or miscommunicated. But what I meant was, the first commit would just be:

```rust
pub fn extension(&self) -> Option<&str> {
        self.path.extension().map(|s| s.to_str().unwrap())
}
```

Then the second commit adds the `.jinja` check. That's how you wanted it split, right? :)

vallentin:
@djc fixed it and also added some tests for `#[template(path = "...", ext = "...")]`

PR #463, opened 2021-03-20
Update version in README.md

technic:
Hi,

I think you need to refer to the latest stable version in the example.
However, I am not  sure whether the other parts of the example shall be updated.


djc (approved):
I don't think any other changes are needed, thanks!

PR #468, opened 2021-03-29
Clippy 1.51

djc:


askama_shared/src/generator.rs
```diff
@@ -344,6 +344,8 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         buf.writeln(&format!("::askama_tide::try_into_body(&self, {:?})", &ext))?;
         buf.writeln("}")?;
         buf.writeln("}")?;
+
+        buf.writeln("#[allow(clippy::from_over_into)]")?;
```
vallentin, line 348: Did you check if it could be rewritten and `impl From` instead? or is it a false positive?

vallentin (approved):
Weird how I'm watching/subscribed to all activity, but I don't get a mail if the CI fails

askama_shared/src/generator.rs
djc: I think it probably couldn't be done (`impl<T: Template> From<T> for tide::Response` would likely violate coherence?). In any case, seems to me that the conversion the other way around is pointless and doesn't make sense.

vallentin: Yeah, fair, was just curious :)

PR #472, opened 2021-04-15
Allow for searching templates in OUT_DIR (fixes #471)

### LINKED ISSUE #471: Allow to easier access templates from OUT_DIR.
Hello, I would like to have two rendering stages of templates. One occurring at compile time (with a custom syntax) and the second occurring at run time (with default syntax). For that, I'm using a build script (https://doc.rust-lang.org/cargo/reference/build-script-examples.html). Unfortunately, askama.toml allows only relative paths from the crate root, whereas the default OUR_DIR (https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) location changes between build types (debug/release).

djc:


askama_shared/src/lib.rs
```diff
@@ -216,6 +223,7 @@ struct General<'a> {
     #[cfg_attr(feature = "serde", serde(borrow))]
     dirs: Option<Vec<&'a str>>,
     default_syntax: Option<&'a str>,
+    generated: Option<bool>,
```
vallentin, line 226:
```suggestion
    #[serde(default)]
    generated: bool,
```

Any particular reason for the `Option`? :)

askama_shared/src/lib.rs
```diff
@@ -45,17 +43,17 @@ impl<'a> Config<'a> {
...
                 dirs,
                 default_syntax,
+                generated,
             }) => (
-                dirs.map_or(default_dirs, |v| {
-                    v.into_iter().map(|dir| root.join(dir)).collect()
-                }),
+                dirs.as_deref().unwrap_or(DEFAULT_DIRS),
                 default_syntax.unwrap_or(DEFAULT_SYNTAX_NAME),
+                generated.unwrap_or(false),
```
vallentin, line 54:
```suggestion
                generated,
```

(if the `Option` is removed)

vallentin (approved):

djc:
Since we haven't had any feedback that this actually solves a problem for anyone, let's just close this.

If anyone is interested in this, feel free to reopen.

PR #474, opened 2021-04-17
Add section to README for IDE plugins

SamJakob:


vallentin (changes requested):
Casing mismatch :)

That being said, I'll let djc decide whether this should be added to the readme

SamJakob:
> Casing mismatch :)
> 
> That being said, I'll let djc decide whether this should be added to the readme

Oops, fixed - thanks!

djc:
I don't think I want a badge for it, but I think it would be good to add a bullet in Features list that says something like:

* IDE support available in [JetBrains products](https://plugins.jetbrains.com/plugin/16591-askama-template-support)

SamJakob:
> I don't think I want a badge for it, but I think it would be good to add a bullet in Features list that says something like:
> 
> * IDE support available in [JetBrains products](https://plugins.jetbrains.com/plugin/16591-askama-template-support)
> 
> (Somewhere near the bottom of the list.)

Okay will take out the badge.
I added an IDE support section further down, do you want me to change that?

djc:
> I added an IDE support section further down, do you want me to change that?

Yes, please change it to what I proposed. For now, a whole section for just the single link seems overkill to me.

SamJakob:
> > I added an IDE support section further down, do you want me to change that?
> 
> 
> 
> Yes, please change it to what I proposed. For now, a whole section for just the single link seems overkill to me.

Sure!

djc (approved):

djc:
Thanks!

PR #476, opened 2021-04-19
 Use raw identifiers for identifiers which collide with Rust keywords

SciStarter:
Rust keywords are valid identifier names in templates, so allow the template expansions to use those identifiers safely, by turning them into raw identifiers during the parse stage.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -14,6 +14,51 @@ use std::collections::HashMap;
 use std::path::PathBuf;
 use std::{cmp, hash, mem, str};
 
+// Identifiers to be replaced with raw identifiers, so as to avoid
+// collisions between template syntax and Rust's syntax. In particular
+// [Rust keywords](https://doc.rust-lang.org/reference/keywords.html)
+// should be replaced, since they're not reserved words in Askama
+// syntax but have a high probability of causing problems in the
+// generated code.
+//
+// This list excludes the Rust keywords *self*, *Self*, and *super*
+// because they are not allowed to be raw identifiers, and *loop*
+// because it's used something like a keyword in the template
+// language.
+#[rustfmt::skip]
+static USE_RAW: [&str; 47] = [
+    "as", "break", "const", "continue", "crate", "else", "enum", "extern", "false", "fn", "for",
+    "if", "impl", "in", "let", "match", "mod", "move", "mut", "pub", "ref", "return",
+    "static", "struct", "trait", "true", "type", "unsafe", "use", "where",
+    "while", "async", "await", "dyn", "abstract", "become", "box", "do", "final", "macro",
+    "override", "priv", "typeof", "unsized", "virtual", "yield", "try",
+];
+
+// The raw identifier versions of the USE_RAW identifiers. The indices
+// of the identifier and raw identifier must be the same, so any
+// modification made to one of the arrays must be mirrored in the
+// other array. We're doing it that way, rather than using a more
+// advanced data structure, because statically initializing more
+// advanced data structures requires extra dependencies like once_cell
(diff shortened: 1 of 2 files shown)


vallentin:
Unrelated, but how come you're posting as @SciStarter, but committing as @djarb?

djc:
Why do this in the parser? It would make much more sense to me in the code generator.

djc:
I'm sorry, but that doesn't make much sense to me. The parser is supposed to parse text into an AST that represents the template as parsed in a way that is easier to handle. The exact name of the variables only becomes a problem when they're used in the generated Rust code -- so that's where they should be renamed. I'm pretty sure there's only a few places where the code generator generates code for variable references; you could start by looking [here](https://github.com/djc/askama/blob/main/askama_shared/src/generator.rs#L1439).

SciStarter:
Okay, I've moved the identifier normalization into the code generator. Is this more to you liking?

djc:
This is much better, thanks! As I understand it, though, we don't actually need to store the raw form in the generator's data structures -- we only need to put the raw form in the generated code, right? That should simplify this some more and also make it easier to follow through on my suggestion from the last comment that we don't need the two separate arrays.

Also, please put the constants and helpers near the bottom of the module as per the convention in this project.

SciStarter:
Unfortunately, that doesn't work due to the way expressions are handled (specifically, because they're rendered into a returned string instead of directly into the output buffer). I'd have to make extensive changes to the expression processing.

SciStarter:
However, due to the return value actually being a String, I can actually do something reasonable along those lines after all.

vallentin:
Ah, I see what the issue is. It's because the key in `locals: MapChain<...>` is a `&str` and not a `String`. A workaround could then be to change the key to a `String` or `Cow<'a, str>`. Then it would be possible for `normalize_identifier()` to return `Cow<'a, str>` instead.

The question now ofc is, whether the price of dynamically allocated keys is better or worse, than the increased binary size, by having the "duplicate" arrays.

SciStarter:
As a rule of thumb, I prefer increased size over slower execution, but that's just a general opinion. Do you have a preference in this instance?

djc:
I think I prefer the constants, but let's make it one array of tuples rather than parallel arrays (and let rustfmt do its thing on it).

askama_shared/src/generator.rs
```diff
@@ -1732,3 +1743,72 @@ enum Writable<'a> {
...
+    ("priv", "r#priv"),
+    ("typeof", "r#typeof"),
+    ("unsized", "r#unsized"),
+    ("virtual", "r#virtual"),
+    ("yield", "r#yield"),
+    ("try", "r#try"),
+];
+
+fn normalize_identifier(ident: &str) -> &str {
+    if let Some(word) = USE_RAW.iter().find(|x| x.0 == ident) {
```
vallentin, line 1809: If the few instances that are out of order in `USE_RAW` were fixed, so the array was ordered. Then it would likely be more efficient to use [`binary_search_by`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by) :)

SciStarter: Maybe it would. However, then the entries would not be in the same order as in the Rust keyword documentation, and a linear search through less than fifty items is not exactly slow. Yeah, it's O(N) instead of O(log(N)) but the constant factors dominate either way.

I'm open to changing that, but I'm not convinced it's worth it.

djc: Agreed, linear search seems fine at this size.

djc (approved):

djc:
Thanks for doing the work on this!

SciStarter:
Thanks for being open to the idea :)

PR #485, opened 2021-05-06
Actix-web 4.0.0-beta.6

MTRNord:
This PR adds the changes requires for version  `4.0.0-beta.6` of actix-web.

I am not sure about the changes in `askama_shared/src/generator.rs`. This was more a quick fix I made than fully researched.

PR is tested and working.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -208,15 +208,15 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     // Implement Actix-web's `Responder`.
     fn impl_actix_web_responder(&mut self, buf: &mut Buffer) -> Result<(), CompileError> {
         self.write_header(buf, "::actix_web::Responder", None)?;
-        buf.writeln("type Future = ::askama_actix::futures::Ready<::std::result::Result<::actix_web::HttpResponse, Self::Error>>;")?;
-        buf.writeln("type Error = ::actix_web::Error;")?;
+        //buf.writeln("type Future = ::askama_actix::futures::Ready<::std::result::Result<::actix_web::BaseHttpResponse, ::actix_web::Error>>;")?;
+        //buf.writeln("type Error = ::actix_web::Error;")?;
         buf.writeln(
             "fn respond_to(self, _req: &::actix_web::HttpRequest) \
-             -> Self::Future {",
+             -> ::actix_web::HttpResponse {",
         )?;
 
         buf.writeln("use ::askama_actix::TemplateToResponse;")?;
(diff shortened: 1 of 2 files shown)


askama_shared/src/generator.rs
```diff
@@ -208,15 +208,15 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+        //buf.writeln("type Error = ::actix_web::Error;")?;
         buf.writeln(
             "fn respond_to(self, _req: &::actix_web::HttpRequest) \
-             -> Self::Future {",
+             -> ::actix_web::HttpResponse {",
         )?;
 
         buf.writeln("use ::askama_actix::TemplateToResponse;")?;
-        buf.writeln("::askama_actix::futures::ready(self.to_response())")?;
+        buf.writeln("self.to_response().expect(\"Failed to generate response\")")?;
```
djc, line 219: I don't think we should have this `expect()`. Can we propagate the error?

askama_shared/src/generator.rs
```diff
@@ -208,15 +208,15 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     // Implement Actix-web's `Responder`.
```
djc, line 208: Not sure what you mean by "`Future`s not being found"?

djc (commented):
Hey, sorry for the slow feedback! I'm on holiday for two weeks so not spending quite as much time at the keyboard as usual. :)

askama_shared/src/generator.rs
MTRNord: The previous version of actix-web used `type Future` while this does just return a `HttpResponse` without any Result or future. the hacky fix is to use expect instead of returning the error. As I think we should return the error but I didnt find any hint in the actix docs on how to return an error from the respond_to method :/ I am probably missing something I think but also didnt research hard enough.

Also enjoy your holiday :)

askama_shared/src/generator.rs
MTRNord: See the other comment :)

askama_shared/src/generator.rs
djc: I'm not sure how to make progress on this. I definitely don't want to call `expect()` here. Maybe you can ask the Actix folks how we're supposed to return an error in this case?

dvtkrlbs:
For tests actix now maintains some test functionality in `actix-test` crate which is still in beta. And since new beta uses tokio 1.0 we also need to bump the actix-rt dependency to 2.0.

djc:
Closing in favor of #553 which has better error handling (doesn't `expect()`).

PR #486, opened 2021-05-16
Remove forward-slash escape

### LINKED ISSUE #245: askama_escape Why escape slashes?
Sorry, I just don't seem to understand what harm slashes can do in html.

[Why does it escape slashes?](https://github.com/djc/askama/blob/467f4ade19fa34983de7e6f6d81c6b4d5ff140fe/askama_escape/src/lib.rs#L134)

alexwennerberg:
**Reopens and resolves #245 based on new recommendation from OWASP**

This was based off of the OWASP XSS prevention cheat sheet --
https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#output-encoding-rules-summary

However, there isn't really any attack vector based on forward slash alone, and
it's being removed in the next version of that document.

> There is no proof that escaping forward slash will improve
> defense against XSS, if all other special characters are escaped
> properly, but it forces developers to use non-standard implementation of
> the HTML escaping, what increases the risk of the mistake and makes the
> implementation harder.

https://github.com/OWASP/CheatSheetSeries/pull/516

### DIFF
--- a/askama_escape/src/lib.rs
+++ b/askama_escape/src/lib.rs
@@ -129,7 +129,6 @@ impl Escaper for Html {
                     b'&' => escaping_body!(start, i, fmt, bytes, "&amp;"),
                     b'"' => escaping_body!(start, i, fmt, bytes, "&quot;"),
                     b'\'' => escaping_body!(start, i, fmt, bytes, "&#x27;"),
-                    b'/' => escaping_body!(start, i, fmt, bytes, "&#x2f;"),
                     _ => (),
                 }
             }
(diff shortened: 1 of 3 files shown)


djc (approved):

vallentin:
This test needs to have `&#x2f;&#x2f;` changed to `//` :)

https://github.com/djc/askama/blob/9232cafb6678160c3fe5154a7619a5e65c037a2b/testing/tests/filters.rs#L18-L27

alexwennerberg:
> This test needs to have `&#x2f;&#x2f;` changed to `//` :)
> 
> https://github.com/djc/askama/blob/9232cafb6678160c3fe5154a7619a5e65c037a2b/testing/tests/filters.rs#L18-L27

Fixed! There was another test that broke too

djc (approved):

djc:
Thanks, this is nice!

PR #489, opened 2021-05-25
actix 4.0 migration

dvtkrlbs:
This should be left as draft until actix 4.0 hits stable.


djc:
Is this substantially different from #485?

dvtkrlbs:
oh i completely missed that i think not this just does not uses bytes and writes to a preallocated string buffer. so no bytes dependency. but i did not see your or so we are good to close this one

PR #495, opened 2021-06-13
Support rocket 0.5.0-rc.1 based on @flo-l's PR

shritesh:
Builds off #412

> FWIW, I've pointed Rocket to git and updated the test to use the new API on top of this PR at https://github.com/shritesh/askama/commit/b9f7a58cee2be9beb52449bedab06ee685287b06. 
Thank you for this.

_Originally posted by @shritesh in https://github.com/djc/askama/issues/412#issuecomment-777797623_

> If someone is more invested than this PR's author, maybe just start a new PR?

_Originally posted by @djc in https://github.com/djc/askama/issues/412#issuecomment-860224048_

### DIFF
--- a/askama_rocket/src/lib.rs
+++ b/askama_rocket/src/lib.rs
@@ -6,11 +6,11 @@ pub use rocket::request::Request;
 use rocket::response::Response;
 pub use rocket::response::{Responder, Result};
 
-pub fn respond<T: Template>(t: &T, ext: &str) -> Result<'static> {
+pub fn respond<'r, 'o, 'r, T: Template>(t: &'r T, ext: &'r str) -> Result<'o> {
     let rsp = t.render().map_err(|_| Status::InternalServerError)?;
     let ctype = ContentType::from_extension(ext).ok_or(Status::InternalServerError)?;
     Response::build()
         .header(ctype)
-        .sized_body(Cursor::new(rsp))
+        .sized_body(rsp.len(), Cursor::new(rsp))
         .ok()
 }
--- a/askama_rocket/src/lib.rs
+++ b/askama_rocket/src/lib.rs
@@ -6,7 +6,7 @@ pub use rocket::request::Request;
 use rocket::response::Response;
 pub use rocket::response::{Responder, Result};
 
-pub fn respond<'r, 'o, 'r, T: Template>(t: &'r T, ext: &'r str) -> Result<'o> {
+pub fn respond<'r, 'o, T: Template>(t: &'r T, ext: &'r str) -> Result<'o> {
     let rsp = t.render().map_err(|_| Status::InternalServerError)?;
     let ctype = ContentType::from_extension(ext).ok_or(Status::InternalServerError)?;
     Response::build()
--- a/askama_rocket/src/lib.rs
+++ b/askama_rocket/src/lib.rs
@@ -6,7 +6,7 @@ pub use rocket::request::Request;
 use rocket::response::Response;
 pub use rocket::response::{Responder, Result};
 
-pub fn respond<'r, 'o, T: Template>(t: &'r T, ext: &'r str) -> Result<'o> {
+pub fn respond<T: Template>(t: &T, ext: &str) -> Result<'static> {
     let rsp = t.render().map_err(|_| Status::InternalServerError)?;
     let ctype = ContentType::from_extension(ext).ok_or(Status::InternalServerError)?;
     Response::build()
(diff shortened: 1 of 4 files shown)


askama_rocket/src/lib.rs
```diff
@@ -6,11 +6,11 @@ pub use rocket::request::Request;
 use rocket::response::Response;
 pub use rocket::response::{Responder, Result};
 
-pub fn respond<T: Template>(t: &T, ext: &str) -> Result<'static> {
+pub fn respond<'r, 'o, 'r, T: Template>(t: &'r T, ext: &'r str) -> Result<'o> {
```
djc, line 9: Uh, is it actually useful/necessary to have two lifetimes with the same name? That seems potentially confusing.

djc (commented):
Thank you for working on this!

If you add a separate commit that bumps the askama_rocket version to an rc as well, I'd be happy to push that out to crates.io if that helps anyone.

askama_rocket/src/lib.rs
shritesh: Ah! It was from upstream. I have removed the subsequent lifetime.

shritesh:
I have bumped the version to 0.11.0-rc.1.

askama_rocket/src/lib.rs
```diff
@@ -6,11 +6,11 @@ pub use rocket::request::Request;
 use rocket::response::Response;
 pub use rocket::response::{Responder, Result};
 
-pub fn respond<T: Template>(t: &T, ext: &str) -> Result<'static> {
+pub fn respond<'r, 'o, T: Template>(t: &'r T, ext: &'r str) -> Result<'o> {
```
djc, line 9: Sorry, I'm still confused: why do we need `'o` here? Since there is no input parameter that uses `'o`, couldn't we keep it as is, with the return type using `'static` directly?

shritesh: We don't. Thank you for pointing that out. The lifetimes are part of Rocket's Responder signature. In our case we should be able to get away with elided param lifetimes and a static out lifetime.

askama_shared/src/generator.rs
```diff
@@ -309,16 +309,22 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+        let out_lifetime = syn::Lifetime::new("'out", Span::call_site());
+        let mut out_def = syn::LifetimeDef::new(out_lifetime);
+        out_def.bounds.push(req_lifetime);
+        let param1 = syn::GenericParam::Lifetime(out_def);
+
         self.write_header(
             buf,
-            "::askama_rocket::Responder<'askama>",
-            Some(vec![param]),
+            "::askama_rocket::Responder<'askama, 'out>",
```
djc, line 322: It seems like here, too, we can simplify to just using `'static` for the output lifetime?

shritesh: Yep. Fixed. Thank you.

djc (approved):

danbruder:
thanks @shritesh 🙂 I was just looking to try askama and rocket 5

jplatte:
I just tried this, and am getting errors about the generated `Responder` impls only having one lifetime parameter supplied when two are required (AFAICT). Am I doing sth. wrong?

Relevant dependencies:

```toml
askama = { version = "0.10.5", features = ["with-rocket"] }
askama_rocket = "0.11.0-rc.1"
rocket = "0.5.0-rc.1"
```

First template derive that fails:

```rust
#[derive(Template)]
#[template(path = "index.html")]
struct IndexTemplate;
```

Error message:

```
error[E0107]: this trait takes 2 lifetime arguments but 1 lifetime argument was supplied
   --> src/routes.rs:18:10
    |
18  | #[derive(Template)]
    |          ^^^^^^^^
    |          |
    |          expected 2 lifetime arguments
    |          supplied 1 lifetime argument
    |
note: trait defined here, with 2 lifetime parameters: `'r`, `'o`
   --> /home/jplatte/.cargo/registry/src/github.com-1ecc6299db9ec823/rocket-0.5.0-rc.1/src/response/responder.rs:181:11
    |
181 | pub trait Responder<'r, 'o: 'r> {
    |           ^^^^^^^^^ --  --
help: add missing lifetime argument
    |
18  | #[derive(Template, 'askama)]
    |                  ^^^^^^^^^
```

Full code: https://github.com/jplatte/turbo.fish/compare/askama

djc:
Doesn't look like you're doing anything. Want to file a new issue (or even better, a PR)?

PR #496, opened 2021-06-14
Fix needless borrowing

djc:


vallentin (approved):
Was this caught by a scheduled CI run? Still wish I would receive an automated email if it fails :)

djc:
Yes: https://github.com/djc/askama/actions/runs/927384208. I'd be happy to have you sent email for scheduled runs, but not sure how to make that happen.

vallentin:
Yeah, it wasn't possible a few months ago. So let's hope GitHub adds it at some point

PR #499, opened 2021-06-21
Fix code generation for macro calls that store args in variables.

### LINKED ISSUE #498: Error "proc-macro derive produced unparseable tokens" with nested macro call
I was working on updating [UniFFI](https://github.com/mozilla/uniffi-rs/) to the latest `main` branch of Askama, and ran into a case where the `#[derive(Template)]` seems to generate invalid code. Below is a reduced example:

```
use askama::Template;

#[derive(Template)]
#[template(source="

use askama::Template;

#[derive(Template)]
#[template(source="

    {%- macro outer(ctx) -%}
    {%- call inner(ctx, \"name\") -%}
    {%- endmacro -%}

    {%- macro inner(ctx, name) -%}
    {{ name }} = value;
    {%- endmacro -%}

    {%- call outer(ctx) -%}

", ext="txt", print="code")]
struct MacroBug {
    ctx: (),
}

impl MacroBug {
    #[cfg(test)]
    fn testme() -> String {
        MacroBug { ctx: () }.render().unwrap()
    }
}

#[cfg(test)]
mod test {
    use super::*;
    #[test]
    fn test_it() {
        assert_eq!(MacroBug::testme(), "name = value;")
    }
}
```

This compiles and runs correctly on the latest Askama release, but when compiled against current `main` ([b318d7cb](https://github.com/djc/askama/commit/b318d7cbcded2c6dfc66bbe19687f1246a9a9eab)) it gives an error:

```
error: expected pattern, found `,`
 --> src/lib.rs:4:10
  |
4 | #[derive(Template)]
  |          ^^^^^^^^ expected pattern
  |
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: proc-macro derive produced unparseable tokens
 --> src/lib.rs:4:10
  |
4 | #[derive(Template)]
  |          ^^^^^^^^
```

The issue seems to be in the code for calling the inner macro. The generated code front `print="code"` looks like:

```
impl ::askama::Template for MacroBug {
    fn render_into(&self, writer: &mut dyn ::std::fmt::Write) -> ::askama::Result<()> {
        {
            {
                let (, name) = (, ("name"));
```

rfk:
This fixes #498 for me locally. I don't know enough about the codebase to know how to write a good test for this or whether there's a cleaner way to fix it. But I'll be happy to polish this up a bit if you can give me a few pointers.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -717,6 +717,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
 
         let mut names = Buffer::new(0);
         let mut values = Buffer::new(0);
+        let mut is_first_variable = true;
         for (i, arg) in def.args.iter().enumerate() {
             let expr = args.get(i).ok_or_else(|| {
                 CompileError::String(format!("macro '{}' takes more than {} arguments", name, i))
@@ -742,7 +743,9 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                 // multiple times, e.g. in the case of macro
                 // parameters being used multiple times.
                 _ => {
-                    if i > 0 {
+                    if is_first_variable {
+                        is_first_variable = false
+                    } else {
                         names.write(", ");
                         values.write(", ");
                     }
--- a/testing/templates/nested-macro-args.html
+++ b/testing/templates/nested-macro-args.html
@@ -1,0 +1,9 @@
+{%- macro outer(first) -%}
+{%- call inner(first, "second") -%}
+{%- endmacro -%}
+
+{%- macro inner(first, second) -%}
+{{ first }} {{ second }}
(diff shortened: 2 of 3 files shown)


djc:
Thanks for fixing this!

There are integration tests in the testing crate, should be easy to copy your reproducer in there. Probably put this in the `macro` module.

(Maybe also shorten the commit message little bit so it fits in one line?)

rfk:
> There are integration tests in the testing crate, should be easy to copy your reproducer in there.
> Probably put this in the macro module.

Sounds good, I should be able to make time to revisit this in the next couple of days.

rfk:
Updated with a test (which I confirmed fails on current `main`) and a briefer commit message.

djc (approved):

djc:
Thanks!

PR #500, opened 2021-06-23
Fixes loop generation when accessing field

### LINKED ISSUE #492: Calling a function with a literal and a variable requires manually writing .borrow()
Using the latest release, we could write code like this (untested, from memory):

```jinja2
{{ foo.bar(value) }}
{{ foo.bar(1) }}
```

```rust
#[derive(Template)]
struct MyTemplate {
    foo: Foo,
    value: u32
}

struct Foo;
impl Foo {
    fn bar(value: &u32) -> String { format!("{}", value) }
}
```

Using the code currently on the main branch, `foo.bar(1)` gets translated to `self.foo.bar(1)`, whereas `foo.bar(value)` gets translated to `self.foo.bar(&self.value)`. This makes it impossible to write such a function without introducing a trait that is implemented for both `u32` and `&u32`, as the function now needs to accept both values and references.

The only other workaround I could find was to use `{{ foo.bar(1.borrow()) }}` in the template code, which requires an explicit `use std::borrow::Borrow` in the file that invokes the derive macro. Still less code than defining my own trait.

vallentin:
Fixes #492

Straightforward to fix, added a case when `iter` is a `Expr::Attr`, so it borrows the field.

Added a few tests, to assert it works regardless if the inner type is `Vec<_>` or `&[_]`. I also added a case for `Range`, though since `Range` is not `Copy` then it requires `.clone()`, so it was added more to ensure that continues to work.

_Not a big fan of the naming of the tests and structs._

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -658,6 +658,12 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                 ", _loop_item) in ::askama::helpers::TemplateLoop::new(((&{}).into_iter())) {{",
                 expr_code
             )),
+            // If accessing a field then it most likely needs to be
+            // borrowed, to prevent an attempt of moving.
+            Expr::Attr(..) => buf.writeln(&format!(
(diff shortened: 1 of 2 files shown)


djc (approved):
Thanks, this is great!

PR #503, opened 2021-06-25
Implement "if let" statement

Kijewski:
This PR implements ``{% if let variant = value %}`` conditions.

The variant can be ``None``, ``Some(foo)``, ``MyStruct { x, y }``,
etc., but nested matching is not implemented. Names are matched by
reference.

The syntax is mostly sugar for

```rs
{% match val %}
    {% when variant %}
        ...
    {% when _ -%}
{% endmatch %}
```

but it can be used in ordinary if-else blocks.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -3,7 +3,8 @@ use crate::filters;
 use crate::heritage::{Context, Heritage};
 use crate::input::{Source, TemplateInput};
 use crate::parser::{
-    parse, Cond, Expr, MatchParameter, MatchParameters, MatchVariant, Node, Target, When, Ws,
+    parse, Cond, CondTest, Expr, MatchParameter, MatchParameters, MatchVariant, Node, Target, When,
+    Ws,
 };
 
 use proc_macro2::Span;
@@ -483,23 +484,38 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                 self.locals.pop();
             }
 
+            self.locals.push();
             let mut arm_size = 0;
-            if let Some(expr) = cond {
+            if let Some(CondTest { target, expr }) = cond {
                 if i == 0 {
                     buf.write("if ");
                 } else {
                     buf.dedent()?;
                     buf.write("} else if ");
                 }
-                // The following syntax `*(&(...) as &bool)` is used to
-                // trigger Rust's automatic dereferencing, to coerce
-                // e.g. `&&&&&bool` to `bool`. First `&(...) as &bool`
-                // coerces e.g. `&&&bool` to `&bool`. Then `*(&bool)`
-                // finally dereferences it to `bool`.
-                buf.write("*(&(");
-                let expr_code = self.visit_expr_root(expr)?;
-                buf.write(&expr_code);
-                buf.write(") as &bool)");
+
+                if let Some((variant, params)) = target {
+                    let mut expr_buf = Buffer::new(0);
+                    self.visit_expr(&mut expr_buf, expr)?;
+                    buf.write("let ");
+                    self.visit_match_variant(buf, variant);
+                    if let Some(params) = params {
+                        self.visit_match_params(buf, params);
+                    }
+                    buf.write(" = &(");
+                    buf.write(&expr_buf.buf);
+                    buf.write(")");
+                } else {
+                    // The following syntax `*(&(...) as &bool)` is used to
+                    // trigger Rust's automatic dereferencing, to coerce
+                    // e.g. `&&&&&bool` to `bool`. First `&(...) as &bool`
+                    // coerces e.g. `&&&bool` to `&bool`. Then `*(&bool)`
+                    // finally dereferences it to `bool`.
+                    buf.write("*(&(");
+                    let expr_code = self.visit_expr_root(expr)?;
+                    buf.write(&expr_code);
+                    buf.write(") as &bool)");
+                }
             } else {
                 buf.dedent()?;
                 buf.write("} else");
@@ -507,7 +523,6 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             }
 
             buf.writeln(" {")?;
-            self.locals.push();
 
             arm_size += self.handle(ctx, nodes, buf, AstLevel::Nested)?;
             arm_sizes.push(arm_size);
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -564,44 +564,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                 None => buf.write("_"),
             };
 
-            match params {
-                MatchParameters::Simple(params) => {
-                    if !params.is_empty() {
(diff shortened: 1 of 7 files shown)


askama_shared/src/generator.rs
```diff
@@ -1106,9 +1129,64 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+                        }
+                        buf.write(param.0);
+                        if let Some(param) = &param.1 {
+                            buf.write(":");
+                            self.visit_match_param(buf, param);
+                        }
+                    }
+                    buf.write("}");
+                }
+            }
```
vallentin, line 1182: Just giving it a quick glance, so I might have overlooked something. But can't all this be replaced with `self.visit_match_params(buf, params);`? :)

askama_shared/src/generator.rs
```diff
@@ -1106,9 +1129,64 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
             Expr::RustMacro(name, args) => self.visit_rust_macro(buf, name, args),
+            Expr::Let(ref variant, ref params, ref val) => self.visit_let(buf, variant, params, val)?,
         })
     }
 
+    fn visit_let(
+        &mut self,
+        buf: &mut Buffer,
+        variant: &MatchVariant<'a>,
+        params: &Option<MatchParameters<'a>>,
```
vallentin, line 1140:
```suggestion
        params: Option<&MatchParameters<'a>>,
```

Please use `Option<&T>` instead of `&Option<T>` :)

askama_shared/src/generator.rs
```diff
@@ -1106,9 +1129,64 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                 self.visit_method_call(buf, obj, method, args)?
             }
             Expr::RustMacro(name, args) => self.visit_rust_macro(buf, name, args),
+            Expr::Let(ref variant, ref params, ref val) => self.visit_let(buf, variant, params, val)?,
```
vallentin, line 1132:
```suggestion
            Expr::Let(ref variant, params, ref val) => self.visit_let(buf, variant, params.as_ref(), val)?,
```

(in extension to the last comment)

vallentin (commented):
A test case with `{% else %}` would also be nice :)

Other than that, I personally would love some sugar for simplifying "if some then unwrap"

askama_shared/src/generator.rs
Kijewski: Thank you for your review!

Oops, yes, you're right.

askama_shared/src/generator.rs
Kijewski: Yes, that's cleaner. One problem is, that I had to add explicit lifetimes to every &Expr.

Kijewski:
> A test case with `{% else %}` would also be nice :)

I added more tests. 

> Other than that, I personally would love some sugar for simplifying "if some then unwrap"

Thank you! Since it's such a prevalent pattern in rust, I thought such a syntactic sugar might be useful to others, too. :)

askama_shared/src/generator.rs
vallentin: I think you can have it be `Option<&MatchParameters<'a>>`, if you remove `'a` from `params` in `visit_match_params()`. At least I don't see why it's required there, though I'm only giving it a quick glance. Overall I think most of the `&'a T` can be changed to `&T` :)

vallentin:
Nice, it didn't register to me, that you could do `if let` on a whole struct with fields. That's pretty cool.

I'll wait until @djc has given it a review :)

djc:
I think the goal here is good to have. In order to be able to review this well, however, some work needs to be done. At the very least, mechanic changes to add explicit lifetimes all over the place should be in a separate commit. I'd also prefer to have any refactoring in a separate commit before the new stuff is added.

Kijewski:
I wanted to wait for the general consensus, if that feature is wanted before I clean up my commits. I'll do that now. :)

vallentin:
I only just gave it a quick scroll through. Besides being explicit, are the added lifetimes required for anything? :)

Kijewski:
> I only just gave it a quick scroll through. Besides being explicit, are the added lifetimes required for anything? :)

In the first iteration I only added the explicit lifetimes when the compiler refused to compile the code without it. Well, that was just about everywhere.

Now, while refactoring (actually re-doing) my changes, I did not look where I needed to add the lifetime. Since the changes are internal to Generator they can't break anything. `cargo clean && cargo clippy -- -D clippy::needless_lifetimes` did not report anything.

Kijewski:
The lifetimes are not needed anymore. I mixed up expressions and nodes. That's fixed now.

askama_shared/src/parser.rs
```diff
@@ -144,7 +144,15 @@ pub enum Target<'a> {
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct Ws(pub bool, pub bool);
 
-pub type Cond<'a> = (Ws, Option<Expr<'a>>, Vec<Node<'a>>);
+pub type Cond<'a> = (Ws, Option<CondTest<'a>>, Vec<Node<'a>>);
+
+type CondTest<'a> = (
```
djc, line 149: Let's make this a `struct` with named fields, I guess the first field could be named `target` or `pattern`?

askama_shared/src/generator.rs
```diff
@@ -1118,7 +1129,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         wrapped
     }
 
-    fn visit_match_params(&mut self, buf: &mut Buffer, params: &'a MatchParameters<'a>) {
+    fn visit_match_params(&mut self, buf: &mut Buffer, params: &MatchParameters<'a>) {
```
djc, line 1132: Can you move this change into the previous commit?

askama_shared/src/generator.rs
```diff
@@ -483,15 +483,28 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                 self.locals.pop();
             }
 
+            self.locals.push();
             let mut arm_size = 0;
-            match *cond {
```
djc, line 487: Can you isolate the `match` -> `if let` change in a separate commit, please?

djc (commented):
Okay, this is much better! I think it's pretty close. Please follow up with these suggestions then I'll take a closer look at the codegen changes for the condition handling.

Kijewski:
Thanks for the review! I made changes.

djc (approved):
This is looking great, thanks for your contribution!

djc:
We can ignore the test failure (which is due to #504), can you fix up the clippy thing?

Kijewski:
cargo fmt is happy now.

djc:
But the clippy issue is still there...

Kijewski:
Ah, sorry. Fixed: https://github.com/djc/askama/compare/4e9493a..12a6783

djc (approved):

Kijewski:
Yay, thank you!

PR #505, opened 2021-07-01
Fix a clippy::needless-borrow warning in askama_shared

Kijewski:


djc (approved):
Thanks!

Kijewski:
You're welcome. :) I wonder why the CI did not report the warning?

djc:
Good question. I think it's because this is in an optional feature, and maybe we don't run clippy with `--all-features`? I think this lint has gotten more strict on nightly, and we probably don't run the Rocket job clippy with `--all-features`, in particular.

vallentin:
One note, if we were to use `--all-features`, then we might want to run Clippy twice, as in with and without `--all-features`. Otherwise some lints won't be triggered.

Simple example:

```rust
fn foo() {}

#[cfg(feature = bar)]
fn bar() {
    foo()
}
```

Assuming `foo()` is only called in `bar()`, then doing `cargo clippy` will trigger a dead code lint, saying `foo()` is unused. However `cargo clippy --all-features` won't, as now it's used. Thus if you do `cargo build` without any features enabled, then you might get lints triggered (locally).

PR #506, opened 2021-07-01
Implement destructoring nested tuples on "let" and "for" statement target

Kijewski:
This PR implements the parsing of nested tuples for the target of `{% let %}` statements. E.g.

```rs
{% let (a, (b, c)) = … %}
```


Kijewski:
If this PR is accepted, my next two PRs will

1. introduce destructoring structs on the lhs, and 
2. the use of the same function for "match" / "if let" targets.

vallentin:
Nice 🎉 

Didn't look at the diffs yet, but from the title I assume it doesn't apply for e.g. `for (a, (b, c)) in ...`? Can the same parsing and generation be applied to that? Would be nice, as then the _last_ point in #451 can be closed :)

Kijewski:
Yes, "for … in" uses the new parsing function: https://github.com/djc/askama/pull/506/files#diff-48297b9b3a3f70c8521bedbaa0601f9e2b405eaa7a99fbffe0d9131e3207e478L904-L907

I'll add a test for nested loops in for-loops.

vallentin:
So, that's a yes. Nice 🎉 

It's much appreciated, that you've arrived and implemented all these nice to have features :)

Kijewski:
Hehe, your welcome. Well, the library is well written, you guys answer fast, so it's quite easy to contribute.

Please have a look at the added test if I understood #451 correctly.

vallentin:
> you guys answer fast

That's probably mostly due to luck + all of us being in the same timezone.

> Please have a look at the added test if I understood #451 correctly.

Yes, that looks great. Optionally, a simple `.iter().enumerate()` might also be nice, as a simple base case test. :)

askama_shared/src/parser.rs
```diff
@@ -404,17 +404,34 @@ fn variant_path(i: &[u8]) -> IResult<&[u8], MatchVariant> {
...
+        let (i, first_target) = target(i)?;
+        let (i, is_unused_paren) = opt_closing_paren(i)?;
+        if is_unused_paren {
+            return Ok((i, first_target));
+        }
+
+        let mut targets = vec![first_target];
+        let push_target = |_, target| targets.push(target);
+        let (i, _) = tuple((
+            fold_many0(preceded(ws(tag(",")), target), (), push_target),
```
vallentin, line 427:
```suggestion
        let (i, _) = tuple((
            fold_many0(preceded(ws(tag(",")), target), (), |_, target| targets.push(target)),
```

Any particular reason this isn't inlined? :)

askama_shared/src/parser.rs
```diff
@@ -404,17 +404,34 @@ fn variant_path(i: &[u8]) -> IResult<&[u8], MatchVariant> {
...
+            opt(ws(tag(","))),
+            ws(tag(")")),
+        ))(i)?;
+        return Ok((i, Target::Tuple(targets)));
```
vallentin, line 431:
```suggestion
        Ok((i, Target::Tuple(targets)))
```

askama_shared/src/parser.rs
```diff
@@ -404,17 +404,34 @@ fn variant_path(i: &[u8]) -> IResult<&[u8], MatchVariant> {
...
+    }
 
-    let (i, (elems, _)) = full(i)?;
-    Ok((i, Target::Tuple(elems)))
+    map(identifier, Target::Name)(i)
```
vallentin, line 434:
```suggestion
    else {
        map(identifier, Target::Name)(i)
    }
```

testing/tests/loops.rs
```diff
@@ -214,3 +217,46 @@ fn test_for_vec_attr_range() {
     };
     assert_eq!(t.render().unwrap(), "1 2 3 4 5 6 ");
 }
+
+struct NotClonable<T: fmt::Display>(T);
+
+impl<T: fmt::Display> fmt::Display for NotClonable<T> {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
```
vallentin, line 224:
```suggestion
    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
```

I know every other `impl fmt::Display` also have `<'_>`, but it's not needed :)

askama_shared/src/generator.rs
```diff
@@ -1513,19 +1492,35 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         DisplayWrap::Unwrapped
     }
 
-    fn visit_target(&mut self, buf: &mut Buffer, target: &'a Target) {
-        match *target {
+    fn visit_target(&mut self, buf: &mut Buffer, var: &Target<'a>) {
+        match *var {
```
vallentin, line 1496:
```suggestion
        match var {
```

Pretty sure you can remove the `*`, it's not needed anymore :)

askama_shared/src/generator.rs
```diff
@@ -1513,19 +1492,35 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+                    self.visit_target(buf, target);
+                    buf.write(",");
+                }
+                buf.write(")");
+            }
+        }
+    }
+
+    fn visit_target_initialized(&mut self, buf: &mut Buffer, var: &Target<'a>) {
+        match *var {
```
vallentin, line 1514:
```suggestion
        match var {
```

askama_shared/src/generator.rs
```diff
@@ -803,24 +803,31 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+                        true
+                    }
+                    Some(meta) if meta.initialized => {
+                        // an initialized variable gets shadowed
+                        true
+                    }
+                    _ => {
+                        // initializes a variable that was introduced in a LetDecl before
+                        false
+                    }
```
vallentin, line 826: Maybe this could/should be a method on `MapChain`, to be able to say `.is_*()`?

vallentin (commented):
Here's a few comments :)

testing/tests/loops.rs
djc: Implicit elided lifetimes are going away, so I think we should move in the other direction.

askama_shared/src/generator.rs
djc: Only if we can use it in multiple places. I would suggest moving the comments out of the match arms, which should make this a lot more compact.

testing/tests/loops.rs
vallentin: Wait, really? That's going to introduce a lot of `<'_>` in general then. Do you have a reference? :)

askama_shared/src/generator.rs
vallentin: Fair, I guess it's a matter of preference. But yes, I agree with the moving the comments :)

Updated suggestion:

```suggestion
                match self.locals.get(&name) {
                    // declares a new variable
                    None => true,
                    // an initialized variable gets shadowed
                    Some(meta) if meta.initialized => true,
                    // initializes a variable that was introduced in a LetDecl before
                    _ => false,
```

testing/tests/loops.rs
djc: See some discussion here: https://github.com/rust-lang/rust/issues/80165. I guess there isn't really a concrete timeline, but I am personally convinced that the eliding the lifetimes does more harm (in terms of being able to understand code) than good.

vallentin: Huh, interesting. I always removed `<'_>`, as I was expecting `clippy::needless-lifetimes` around the corner to lint that in a new version. Even Rust's own `impl fmt::Display` examples use just `fmt::Formatter`.

Though I do agree, that eliding lifetimes in paths do more harm. Well, guess I have a lot of refactoring to do, in a lot of projects :)

djc: I think `cargo fix --edition-idioms` should be able to fix most of it, but it didn't seem to work all that well in this case (maybe `cargo fix` works in general after denying the lint?).

Kijewski:
I applied your suggestions.

askama_shared/src/parser.rs
```diff
@@ -404,17 +404,35 @@ fn variant_path(i: &[u8]) -> IResult<&[u8], MatchVariant<'_>> {
...
 
-fn target_single(i: &[u8]) -> IResult<&[u8], Target<'_>> {
-    map(identifier, |s| Target::Name(s))(i)
-}
+fn target(i: &[u8]) -> IResult<&[u8], Target<'_>> {
+    let mut opt_opening_paren = map(opt(ws(tag("("))), |o| o.is_some());
+    let mut opt_closing_paren = map(opt(ws(tag(")"))), |o| o.is_some());
+
+    let (i, target_is_tuple) = opt_opening_paren(i)?;
+    if target_is_tuple {
```
djc, line 412: Please invert the condition here so that we can return early for the simple case.

askama_shared/src/generator.rs
```diff
@@ -807,24 +807,25 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         self.handle_ws(ws);
         self.write_buf_writable(buf)?;
         buf.write("let ");
+        self.visit_target(buf, var);
+        buf.writeln(";")
+    }
+
+    fn write_let_needs_let(&self, var: &'a Target<'_>) -> bool {
```
djc, line 814: This name is not so clear to me. Maybe revisit it or add some documentation about what cases this is intended to handle? In particular, this is quite dissimilar from the other methods that have a name starting with `write_`, so that should be stripped off.

askama_shared/src/generator.rs
```diff
@@ -1533,19 +1506,35 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         DisplayWrap::Unwrapped
     }
 
-    fn visit_target(&mut self, buf: &mut Buffer, target: &'a Target<'_>) {
-        match *target {
+    fn visit_target(&mut self, buf: &mut Buffer, var: &Target<'a>) {
```
djc, line 1509: In general, you have a lot going on in this commit. It would be great to split it into smaller pieces. For example, renaming this variable should and getting rid of the explicit lifetime could go in a separate commit so that it doesn't clutter up commits with more meaty changes.

If possible, it would be good to extract `visit_target_initialized()` and `write_let_needs_let` in different commits. Maybe the commit messages could also explain more clearly what you're trying to achieve here, since it's not that clear to me.

testing/templates/for-destructoring-tuple.html
```diff
@@ -0,0 +1 @@
+{% for (((a,), b), c) in v %}{{a}}{{b}}{{c}}-{% endfor %}
```
djc, line 1: For short templates like this, maybe put them inline with `#[template(source)]`?

djc (commented):
The scond commit looks great, but I can't quite make sense of what's going on with the first one. Left a bunch more specific questions/suggestions, below.

askama_shared/src/parser.rs
Kijewski: I would need to invert the inversion  in my next PR #509: https://github.com/djc/askama/pull/509/commits/465f69cdad36656bac57761ba87573af8fb45522#diff-48297b9b3a3f70c8521bedbaa0601f9e2b405eaa7a99fbffe0d9131e3207e478R443-R468

Shall I do in anyway?

askama_shared/src/generator.rs
Kijewski: Yeah … I need to clean that up. Will do.

Kijewski:
I split the 1st commit into 6 commits. :-/ → #510

Kijewski:
Rebased on #510.

Kijewski:
Rebased on main. I could not reproduce the error in askama_rocket at home.

djc:
Yeah, I think the askama_rocket issue was a nightly regression that has already been fixed.

askama_shared/src/parser.rs
djc: Nope, no need.

djc (approved):
Thanks!

PR #507, opened 2021-07-01
Stop eliding lifetimes in paths

djc:


askama_shared/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![cfg_attr(feature = "cargo-clippy", allow(unused_parens))]
+#![deny(elided_lifetimes_in_paths)]
```
vallentin, line 2: Maybe this should be added to the other crates as well?

vallentin (approved):

vallentin (approved):

PR #509, opened 2021-07-01
Implement destructoring structs on "let" and "for" statement target 

Kijewski:
Depends on #506.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -824,9 +824,12 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                     _ => false,
                 }
             }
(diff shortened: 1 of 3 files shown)


Kijewski:
Rebased on #506.

djc (approved):

djc:
Thanks again! Do you have any more short-term plans for adding stuff?

Kijewski:
You're welcome! All my short-term plans are fulfilled.

In a week (or most likely later) I'd like to unify parser.rs' match_variant() and target(), so that `match … with …` can parse (almost) every valid pattern.

I think #513 (Better error messages using cuts) would be a nice-to-have, too, but right now I don't have the time to test it properly.

PR #510, opened 2021-07-01
Fix and optimize visit_target usage

Kijewski:
There was a bit of code duplication in generator.rs with respect to the output of targets.

Also there was a bug which caused broken code for

```rs
{% for val in range %}{% set val = "anything" %}{% endfor %}
```

which tried to write to `val` instead of shadowing it.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -1533,7 +1533,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         DisplayWrap::Unwrapped
     }
 
-    fn visit_target(&mut self, buf: &mut Buffer, target: &'a Target<'_>) {
+    fn visit_target(&mut self, buf: &mut Buffer, target: &Target<'a>) {
         match *target {
             Target::Name(name) => {
                 let name = normalize_identifier(name);
@@ -1543,9 +1543,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             Target::Tuple(ref targets) => {
                 buf.write("(");
                 for name in targets {
-                    let name = normalize_identifier(name);
-                    self.locals.insert_with_default(name);
-                    buf.write(name);
+                    self.visit_target(buf, &Target::Name(name));
                     buf.write(",");
                 }
                 buf.write(")");
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -811,6 +811,25 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         buf.writeln(";")
     }
 
+    fn is_shadowing_variable(&self, var: &Target<'a>) -> bool {
+        match var {
+            Target::Name(name) => {
+                let name = normalize_identifier(name);
+                match self.locals.get(&name) {
+                    // declares a new variable
+                    None => false,
+                    // an initialized variable gets shadowed
+                    Some(meta) if meta.initialized => true,
+                    // initializes a variable that was introduced in a LetDecl before
+                    _ => false,
+                }
+            }
+            Target::Tuple(targets) => targets
+                .iter()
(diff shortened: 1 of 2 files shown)


askama_shared/src/generator.rs
```diff
@@ -811,6 +811,32 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+                    None => (false, true),
+                    // an initialized variable gets shadowed
+                    Some(meta) if meta.initialized => (true, true),
+                    // initializes a variable that was introduced in a LetDecl before
+                    _ => (false, false),
+                }
+            }
+            Target::Tuple(targets) => {
+                let shadowed = targets.iter().any(|name| {
+                    let (shadowed, _) = self.shadows_or_declares_variable(&Target::Name(name));
```
djc, line 828: `self.shadows_or_declares_variable(&Target::Name(name)).0`?

askama_shared/src/generator.rs
```diff
@@ -811,6 +811,32 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+                    shadowed
+                });
+                if shadowed {
+                    (true, true)
+                } else {
+                    (false, true)
+                }
```
djc: `(shadowed, true)`

askama_shared/src/generator.rs
```diff
@@ -822,36 +848,24 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
 
+        let (shadowed, declared) = self.shadows_or_declares_variable(var);
+        if shadowed {
+            // Need to flush the buffer if the variable is being shadowed,
+            // to ensure the old variable is used.
+            self.write_buf_writable(buf)?;
+        }
+        if declared {
+            buf.write("let ");
+        }
```
djc: Would it make sense to inline these effects into `shadows_or_declares_variable()` (I suppose it would need yet another name)? Splitting it this way doesn't make for a great abstraction.

askama_shared/src/generator.rs
```diff
@@ -1531,17 +1515,21 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+    fn visit_target(&mut self, buf: &mut Buffer, initialized: bool, target: &Target<'a>) {
         match target {
             Target::Name(name) => {
                 let name = normalize_identifier(name);
-                self.locals.insert_with_default(name);
+                if initialized {
+                    self.locals.insert(name, LocalMeta::initialized());
+                } else {
+                    self.locals.insert_with_default(name);
+                }
```
djc, line 1526: Here's an example where I'd probably write a `match` instead. What do you think?

djc (commented):
Wow, thanks for taking on this much-needed refactoring. Some fairly minor comments.

Kijewski:
I addressed you comments.

Actually I'm not sure if I like matching bools. I'm quite new to rust and as a Python / C++ programmer it just feels wrong. You'd get publicly stoned if you shared C++ code that contains `switch (cond) { case true: …`  :)

askama_shared/src/generator.rs
```diff
@@ -811,6 +811,25 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         buf.writeln(";")
     }
 
+    fn is_shadowing_variable(&self, var: &Target<'a>) -> bool {
```
djc, line 814: This is currently single use. Will you add more direct uses in the downstream PRs? If not, I'd prefer to just inline it.

Kijewski: Yes, the method calls itself recursively for nested tuples and structs.

djc: Duh...

djc:
> Actually I'm not sure if I like matching bools. I'm quite new to rust and as a Python / C++ programmer it just feels wrong.

Can you articulate what you dislike about them?

djc (approved):

Issue #504, opened 2021-06-29
Test failure in macros::args on beta Rust


djc:
@vallentin did you already have some attempt to minimize the reproduction for this? I really want to report it upstream soon so we can understand what's going on here.

vallentin:
Yes, but I'm not home until tomorrow :)

djc:
Ok.

djc:
@vallentin friendly ping?

Kijewski:
The diff between stable and testing is:

```diff
-"some ninety \"(\\\"()\"nine !"
+"some ninety \"(\\\"()\"nine!"
```

I.e. stable adds a spaces in front of the bang. Since `call_a_or_b_on_tail!()` operates on a TokenTree, I guess it is allowed to add extraneous spaces, or omit them if they are not needed. IMO the old behavior was wrong, but rust_macro.rs operates under the assumption that the space gets added implicitly.

djc:
Thanks for looking into this! Do you have any suggestions on how to make rust_macro.rs robust to these kinds of changes?

vallentin:
@Kijewski already identified the issue and posted it on Gitter, but thanks anyways :)

Having `stringify!(foo ! bar !)` snap exclamation marks to the left _does make sense_ to a degree, if and only if the goal of `stringify!` is to loosely mirror rustfmt. So I guess the question is more whether stringification should take everything literal, or if it should _format_ it, which I don't think it should, given it's called `stringify` and not `rustfmtify`.

I'm looking for the commit and PR where this was changed, to see if it was acknowledged or an accidental side-effect of another change.

-----

In regards to testing, then I noticed that the [`stringify!`](https://doc.rust-lang.org/std/macro.stringify.html) docs, includes the following:

> Note that the expanded results of the input tokens may change in the future. You should be careful if you rely on the output.

So I guess we should be wary of using `stringify!` in tests.

Kijewski:
No, not really. I guess the only option is to write a test that does not depend on the output of stringify.

djc:
Ah, the offending commit is https://github.com/djc/askama/commit/39e08325c0e0fba75384579427d1c6f7eca8fbc9, that's much older than I thought.

vallentin:
Yes, that the test here in askama that's the issue.

Though to be clear, I'm looking for the commit and PR in the rust repo ofc :)

vallentin:
[...]

PR #511, opened 2021-07-02
Replace rust_macro test to work on nightly

Kijewski:
The current rust_test uses `stringify!()`. The documentation gives us
the warning:

> Note that the expanded results of the input tokens may change in the
> future. You should be careful if you rely on the output.

In the current nightly rust the result was indeed changed, so the test
not fails.

This PR replaces the test with another macro, that does not depend on
`stringify!()`.

Closes issue #504.


vallentin:
I'm sick for the next few days. However, I just wanted to chime in, and say thanks once again :)

Sadly, an error message has changed, so now there's still another test failing :(

djc:
@vallentin hope you get well soon, get some rest!

Kijewski:
Get well soon!

Kijewski:
The next broken test is because of rust-lang/rust#82069. It would be possible to utilize [rustversion](https://crates.io/crates/rustversion) select the right failure?

djc:
Ahhh, green tests. Thanks!

djc (approved):

Kijewski:
You're welcome! :)

PR #512, opened 2021-07-02
Fix expected error message for missing file

Kijewski:
rust-lang/rust#82069 made error message that stem macro invocations more
verbose. Since Rust 1.54 (currently in beta) the message includes the
name of the offending macro.

This PR uses version_check to select the appropriate expected error
message.


djc (approved):

djc:
Thanks! In future work, for if/else expressions where the contained block is a single expression that fits on one line, please consider matching the boolean instead.

Kijewski:
Will do.

PR #513, opened 2021-07-03
Better error messages using cuts

Kijewski:
This PR is not well tested and is more akin to a prove of concept.

The idea it to turn recoverable parsing errors into failures, when there can be no other valid path.

The error location then is used to report the (approximate) row and column.

### DIFF
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -1,7 +1,7 @@
 use nom::branch::alt;
 use nom::bytes::complete::{escaped, is_not, tag, take_until};
 use nom::character::complete::{anychar, char, digit1};
-use nom::combinator::{complete, map, opt, recognize, value};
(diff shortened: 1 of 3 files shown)


djc:
I think this makes sense, thanks for looking into it! Would you like to rebase it and fix the formatting?

djc:
(Not sure about pulling in the extra dependencies just for more accurate error reporting, unless they're already part of the transitive dependency set.)

Kijewski:
Yeah, that's why I left it a separate commit. It's probably vastly overkill, and is probably only useful if you use emojis in your templates.

Kijewski:
Rebased and formatted.

(block_raw that parses `{% raw %} … {% endraw %}` looks broken to me, but I haven't actually tested it. I'd tackle that after this PR is merged.)

djc (approved):
Nice, thanks!

PR #516, opened 2021-07-17
 Use "target()" to parse "when" block 

Kijewski:
`target()` as used in parsing "let" and "if let" implements parsing nested tuples and structs. But it does not implement parsing literals.

The functions `match_variant()` and `with_parameters()` as used in parsing "when" blocks do not implement parsing nested structs, but it implements parsing literals.

This PR combines `match_variant()` and `with_parameters()` into `target()`, so that all `{%when%}` support nested structs, too.


djc (approved):

djc:
Thanks, and sorry for the slow review!

Kijewski:
You're welcome! No problem at all.

Issue #427, opened 2021-01-06
Tracking issue for unsupported Jinja features


vallentin:
This is a tracking issue of an excerpt of unsupported Jinja features, as specified in Jinja's [Template Designer Documentation].
The goal is not to list all unsupported features, but mainly focus on features that are useful or otherwise fit in Askama.

[Template Designer Documentation]: https://jinja.palletsprojects.com/en/2.11.x/templates/

_Missing filters are not included, as these are already tracked in issue #18._

The included features, are those that seem feasible to implement in Rust.

-----

- [ ] `loop.cycle("odd", "even")`
  - Which can currently be expressed with `{% if loop.index0 % 2 == 0 %}odd{% else %}even{% endif %}`
  - This could allow any amount of items to cycle, e.g. `loop.cycle("foo", "bar", "baz", "qux")`, which would make it less trivial to express with `{% if %}`

- [ ] `loop.length`, `loop.revindex`, `loop.revindex0`
  - Could be implemented using [`len()`] from [`ExactSizeIterator`], that way they are guaranteed to be exact
  - Support for any generic iterator, can remain, as long as [`ExactSizeIterator`] is only required, if `loop.length`, etc is used

[`len()`]: https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html#method.len
[`ExactSizeIterator`]: https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html

- [ ] `loop.previtem`, `loop.nextitem`
  - Could be implemented similarly to the previous, by instead requiring `Clone`
  - Jinja defines e.g. `loop.previtem` as undefined on the first iteration, so either it could become `Option<T>` or `T` if it implicitly `unwrap()`ed

- [ ] `loop.changed`, _"True if previously called with a different value (or not called at all)."_
  - Could be useful when splitting sections based on some change

- [ ] `{% for user in users if not user.hidden %}`
  - Support for `if` in `for` could simplify template code
  - Could potentially also improve performance, by using `filter()` on the iterator

- [ ] `{% for ... %}{% else %}{% endfor %}`
  - Support for `else` in `for` would simplify handling for loops with no iterations

-----

- [ ] `{% if foo is defined %}`
  - This is already mention in #257. However, that proposes an [`is_some()`] check, along with an `unwrap()`, i.e. `if let Some`
  - Instead, to be more on par with Tera and Jinja, I instead propose that `foo is defined` is checked in the generator, and checks if the `foo` field exists in the template struct.
[...]

PR #517, opened 2021-07-18
Implement {{loop.cycle(…)}} similar to Jinja

Kijewski:


djc:
When this was discussed before, I proposed we should make this `loop.cycle([1, 2, 3])` to make this more like Rust (where we don't have arbitrary numbers of arguments. What do you think?

Kijewski:
Sorry, I didn't see the prior discussions, I only saw that is was mentioned in #427. The braces it does look more like rust. The only problem I foresee is that someone might try to use the function like `loop.cycle(self.array)`. What code would you generate in this case?

```rs
{
    // This moves the variables out of self:
    let _cycle = self.array;
    _cycle[_cycle.len() % index]
}

{
    // If the variable is used for printing, then this should be fine,
    // but it's used in e.g. an addition, it fails:
    let _cycle = &self.array;
    _cycle[_cycle.len() % index]
}

{
    // This must not be used for e.g. {{ loop.cycle([a(), b(), c()]) }},
    // but I guess you shouldn't use the function like that in any case:
    self.sequence[self.array.len() % index]
}
```

djc:
We take a reference to the array and index into it?

vallentin:
I was just wondering, what would happen if you moved `loop.cycle` into a macro, and called it multiple times in the loop.

Jinja:

```jinja
{% macro foo() %}
{{ loop.cycle("A", "B", "C") }}
{% endmacro %}

{% for v in values %}
{{ foo() }} {{ foo() }} {{ foo() }}
{% endfor %}
```

I was wondering if it would output `AAABBBCCC...` or `ABCABCABC...`. I did a quick test using 2 online "Jinja live parsers". The first one froze, the second one said "'loop' is undefined".

Should Askama support this? If yes, should the output be the former or the latter?

vallentin:
I only spent a couple of minutes searching, but I wasn't able to find a specific source stating the behavior, thus why I did a quick test. Both behaviors are possible using Jinja's `cycler()`, so if `loop.cycle()` is a shorthand for creating a `cycler(...)` before the loop, and then in the loop doing `.next()`. Then the output would be `ABCABCABC...` :)

i.e., if the previous Jinja code (using `loop.cycle("A", "B", "C")`) is a shorthand of this:

```jinja
{% set abc = cycler("A", "B", "C") %}

{% macro foo() %}
    {{ abc.next() }}
{% endmacro %}

{% for v in values %}
    {{ foo() }} {{ foo() }} {{ foo() }}
{% endfor %}
```

djc:
Oh, I didn't know that Jinja had that feature. Still, since we're specifically referring to the loop context here, I think we should stick with the semantics mentioned in my previous comment.

vallentin:
> Still, since we're specifically referring to the loop context here

I agree, with "given loop context". I was just posing the question, so we're all on the same page :)

Kijewski:
I'm not a 100% sure if I like the variant with brackets better, so I didn't squash the commit immediately.

djc:
I think the variant with brackets is nice, especially if that means it can accept arbitrary slices.

Would you mind rebasing this (and ideally squashing it)?

djc (approved):

djc:
Thanks! If anyone wants to follow up with some documentation in the book, that would be great.

PR #518, opened 2021-07-18
 Implement for-else 

Kijewski:
This PR implements for-else statements like in Jinja. They make it easy
to print an alternative message if the loop iterator was empty. E.g.

```rs
{% for result in result %}
  <li>{{ result }}</li>
{% else %}
  <li><em>no results</em></li>
{% endfor %}
```

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -2,7 +2,7 @@ use super::{get_template_source, CompileError, Integrations};
 use crate::filters;
 use crate::heritage::{Context, Heritage};
 use crate::input::{Source, TemplateInput};
-use crate::parser::{parse, Cond, CondTest, Expr, Node, Target, When, Ws};
+use crate::parser::{parse, Cond, CondTest, Expr, Loop, Node, Target, When, Ws};
 
 use proc_macro2::Span;
 
@@ -420,8 +420,8 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                 Node::Match(ws1, ref expr, ref arms, ws2) => {
                     self.write_match(ctx, buf, ws1, expr, arms, ws2)?;
                 }
-                Node::Loop(ws1, ref var, ref iter, ref body, ws2) => {
-                    self.write_loop(ctx, buf, ws1, var, iter, body, ws2)?;
+                Node::Loop(ref loop_block) => {
+                    self.write_loop(ctx, buf, loop_block)?;
                 }
                 Node::BlockDef(ws1, name, _, ws2) => {
                     self.write_block(buf, Some(name), Ws(ws1.0, ws2.1))?;
@@ -596,21 +596,19 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         &mut self,
         ctx: &'a Context<'_>,
         buf: &mut Buffer,
-        ws1: Ws,
-        var: &'a Target<'_>,
-        iter: &Expr<'_>,
-        body: &'a [Node<'_>],
-        ws2: Ws,
+        loop_block: &'a Loop<'_>,
     ) -> Result<usize, CompileError> {
-        self.handle_ws(ws1);
+        self.handle_ws(loop_block.ws1);
         self.locals.push();
 
-        let expr_code = self.visit_expr_root(iter)?;
+        let expr_code = self.visit_expr_root(&loop_block.iter)?;
 
         let flushed = self.write_buf_writable(buf)?;
+        buf.writeln("{")?;
+        buf.writeln("let mut _did_not_loop = true;")?;
         buf.write("for (");
-        self.visit_target(buf, true, true, var);
-        match iter {
+        self.visit_target(buf, true, true, &loop_block.var);
+        match loop_block.iter {
             Expr::Range(_, _, _) => buf.writeln(&format!(
                 ", _loop_item) in ::askama::helpers::TemplateLoop::new({}) {{",
                 expr_code
@@ -645,13 +643,24 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             )),
         }?;
 
-        let mut size_hint = self.handle(ctx, body, buf, AstLevel::Nested)?;
-        self.handle_ws(ws2);
-
-        size_hint += self.write_buf_writable(buf)?;
-        buf.writeln("}")?;
+        buf.writeln("_did_not_loop = false;")?;
+        let mut size_hint1 = self.handle(ctx, &loop_block.body, buf, AstLevel::Nested)?;
+        self.handle_ws(loop_block.ws2);
+        size_hint1 += self.write_buf_writable(buf)?;
         self.locals.pop();
-        Ok(flushed + (size_hint * 3))
+        buf.writeln("}")?;
+
+        buf.writeln("if _did_not_loop {")?;
+        self.locals.push();
+        let mut size_hint2 = self.handle(ctx, &loop_block.else_block, buf, AstLevel::Nested)?;
+        self.handle_ws(loop_block.ws3);
+        size_hint2 += self.write_buf_writable(buf)?;
+        self.locals.pop();
+        buf.writeln("}")?;
+
+        buf.writeln("}")?;
+
+        Ok(flushed + ((size_hint1 * 3) + size_hint2) / 2)
     }
 
     fn write_call(
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -605,52 +605,50 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
 
         let flushed = self.write_buf_writable(buf)?;
         buf.writeln("{")?;
-        buf.writeln("let mut _did_not_loop = true;")?;
-        buf.write("for (");
-        self.visit_target(buf, true, true, &loop_block.var);
+        buf.writeln("let mut _did_loop = false;")?;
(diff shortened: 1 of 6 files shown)


djc:
I'm not sure I'm a fan of this, since Rust does not have for-else itself. In most cases, you could probably handle this by calling `is_empty()` on the source of the iterator (for example, for slice or `Vec`). Any thoughts?

Kijewski:
I implemented for-else like Jinja does, which is not equivalent to Python's semantics.

In Python the else block is executed unless a `break` statement was hit. That's trivially true for empty ranges, but might also be the case for non-empty ranges.

In Jinja the else block is only executed if the inner block was never entered. I guess I should implement `for … in … if …`, too, then the feature is vastly more useful, too.

Kijewski:
I added `for … in … if …`. I don't quite like how the source for that particular feature looks like, but I don't know to to make it "prettier" right now.

djc:
Do you actually have use cases for this syntax, or are you just aiming for feature parity with Jinja?

vallentin:
For what it's worth, I technically have a use case for both :)

On one hand, I always think less code is nice. On the other hand, it only requires 1 extra set of `{% if ... %}` + `{% endif %}` (and `{% else %}` for the first one.) So it isn't specifically "needed", but it would be nice, but again, that's just my opinion.

askama_shared/src/generator.rs
```diff
@@ -593,13 +593,17 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+        ws3: Ws,
     ) -> Result<usize, CompileError> {
         self.handle_ws(ws1);
         self.locals.push();
 
         let expr_code = self.visit_expr_root(iter)?;
 
         let flushed = self.write_buf_writable(buf)?;
+        buf.writeln("{")?;
+        buf.writeln("let mut _did_not_loop = true;")?;
```
djc, line 626: Let's call this `did_loop` (defaulting to false), which is easier to understand. I don't think it will be unused, since we always read it?

askama_shared/src/generator.rs
```diff
@@ -637,13 +641,24 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             )),
         }?;
 
-        let mut size_hint = self.handle(ctx, body, buf, AstLevel::Nested)?;
+        buf.writeln("_did_not_loop = false;")?;
```
djc, line 662: Please keep changes to the code formatting in a different commit, or just leave them out, as this makes it harder to review the actual changes here.

askama_shared/src/parser.rs
```diff
@@ -864,6 +872,16 @@ fn block_let(i: &[u8]) -> IResult<&[u8], Node<'_>> {
...
+    let else_block = preceded(
+        ws(tag("else")),
+        cut(tuple((
+            opt(tag("-")),
+            |i| tag_block_end(i, s),
+            |i| parse_template(i, s),
+            |i| tag_block_start(i, s),
+            opt(tag("-")),
+        ))),
+    );
```
djc, line 886: I think I'd prefer to have this in a separate function.

askama_shared/src/parser.rs
```diff
@@ -879,24 +897,29 @@ fn block_for<'a>(i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], Node<'a>>
...
-            var,
-            iter,
-            block,
-            Ws(pws2.is_some(), nws2.is_some()),
-        ),
-    ))
+    let (i, (pws1, _, (var, _, (iter, nws1, _, (block, (_, pws2, else_block, _, nws2)))))) = p(i)?;
+    match else_block {
+        None => {
+            let ws1 = Ws(pws1.is_some(), nws1.is_some());
```
djc, line 915: Let's hoist the definition of `ws1` out of the `match` expression.

testing/tests/loop_else.rs
```diff
@@ -0,0 +1,19 @@
+use askama::Template;
+
+#[derive(Template)]
+#[template(
+    source = "{% for v in values %}{{v}}{% else %}empty{% endfor %}",
```
djc, line 5: Common style in this project is to keep spaces around expression in `{{ v }}`.

askama_shared/src/generator.rs
djc: Ah, never mind about the _ prefix... we're doing this to prevent clashes with user-defined variables.

askama_shared/src/generator.rs
Kijewski: What are you referring to, there is no formatting change? Before you had one one branch, now you have two, so the size_hint needs to be calculated differently.

askama_shared/src/parser.rs
Kijewski: Done.

askama_shared/src/parser.rs
Kijewski: Done.

testing/tests/loop_else.rs
Kijewski: Done.

askama_shared/src/generator.rs
Kijewski: Renamed to `_did_loop`.

askama_shared/src/parser.rs
```diff
@@ -879,24 +902,28 @@ fn block_for<'a>(i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], Node<'a>>
...
-            Ws(pws1.is_some(), nws1.is_some()),
-            var,
-            iter,
-            block,
-            Ws(pws2.is_some(), nws2.is_some()),
-        ),
-    ))
+    let (i, (pws1, _, (var, _, (iter, nws1, _, (block, (_, pws2, else_block, _, nws2)))))) = p(i)?;
+    let ws1 = Ws(pws1.is_some(), nws1.is_some());
+    match else_block {
```
djc, line 919: Here, it would be nice to keep the `match` block more tightly scoped, having it return the right part of `ws2`, the left part of `ws3`, and the else block. Makes it a bit clearer what's going and reduces code duplication.

askama_shared/src/parser.rs
```diff
@@ -863,6 +871,21 @@ fn block_let(i: &[u8]) -> IResult<&[u8], Node<'_>> {
     ))
 }
 
+fn else_block<'a>(i: &'a [u8], s: &'a Syntax<'a>) -> IResult<&'a [u8], (Ws, Vec<Node<'a>>)> {
```
djc, line 875: What is the rationale for extracting this from `block_for`? If we're going to extract it, can we share it with `if`?

askama_shared/src/generator.rs
djc: No clue... Never mind this.

djc (commented):
Ugh, sorry for the extremely slow review cycle. I've been conflicted about whether to accept this, but in the end I think it makes sense. Some minor style nits to discuss and then hopefully we can get this merged.

askama_shared/src/parser.rs
Kijewski: Would it make sense to add helper function like this?

```r
fn parse_ws(inner: impl Fn(Input) -> Output) -> impl Fn(Input) -> IResult<Input, (Ws, Output)>
```

Matching the optional hyphens and returning  `Ws(pws.is_some(), nws.is_some())` comes up _a lot_.

askama_shared/src/parser.rs
Kijewski: There's no good reason other than readability. I'll turn it into a closure.

askama_shared/src/parser.rs
Kijewski: I just noticed that I can use unwrap_or_default() to that no match at all is needed.

askama_shared/src/generator.rs
```diff
@@ -598,62 +598,69 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
-                expr_code
-            )),
+            Expr::Attr(..) => buf.writeln(&format!("let _iter = (&{}).into_iter();", expr_code)),
             // Otherwise, we borrow `iter` assuming that it implements `IntoIterator`.
-            _ => buf.writeln(&format!(
-                ", _loop_item) in ::askama::helpers::TemplateLoop::new(({}).into_iter()) {{",
-                expr_code
-            )),
+            _ => buf.writeln(&format!("let _iter = ({}).into_iter();", expr_code)),
         }?;
```
Kijewski, line 630: It's possibly somewhat difficult to see what's going on in this change. Essentially it's this for every case:

```diff
-", _loop_item) in ::askama::helpers::TemplateLoop::new(…) {{",
+"let _iter = (…).into_iter();"
```

askama_shared/src/parser.rs
```diff
@@ -33,6 +33,17 @@ pub enum Node<'a> {
     Continue(Ws),
 }
 
+#[derive(Debug, PartialEq)]
+pub struct Loop<'a> {
```
djc, line 37: Thanks for extracting the type here!

askama_shared/src/generator.rs
djc: Yup, did figure that out, but thanks for calling it out!

djc:
Ah, this has a conflict now. Otherwise looking good; for bonus points, also add some docs?

djc (approved):

Kijewski:
Thank you! I'll have a look at the book. I guess the changes of some of the older PRs need to be mentioned, too.

PR #519, opened 2021-07-18
Add {% break %} and {% continue %}

Kijewski:
This PR adds `{% break %}` and `{% continue %}` statements to break out
of a loop, or continue with the next element of the iterator.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -459,6 +459,16 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                     // No whitespace handling: child template top-level is not used,
                     // except for the blocks defined in it.
                 }
+                Node::Break(ws) => {
+                    self.handle_ws(ws);
+                    self.write_buf_writable(buf)?;
+                    buf.writeln("break;")?;
+                }
+                Node::Continue(ws) => {
+                    self.handle_ws(ws);
+                    self.write_buf_writable(buf)?;
+                    buf.writeln("continue;")?;
+                }
             }
         }
 
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -1,3 +1,6 @@
+use std::cell::Cell;
+use std::str;
+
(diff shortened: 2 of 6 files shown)


djc:
This looks like a nice straightforward implementation, but it looks like this would not throw any error if `{% break %}` or `{% continue %}` are added outside of a loop context -- I think it probably should. Any ideas? (I think it might be easier to do it at the codegen stage than in the parser.)

Kijewski:
I implemented the check inside of the parsing step. It was easier to implement, and it has the added bonus that the user gets informed about the exact position of the offending {%break%}.

This was the first time I had to use a Cell. Does this mean I can call myself a Rust programmer now? ;)

djc:
Why do we need to use `Cell` here? Could we just pass `&mut State<'a>` around here? (Also I think we could pass `State<'_>` rather than `State<'a>`, would you mind checking that?)

djc:
Note that Jinja does not allow this:

> Unlike in Python, it’s not possible to break or continue in a loop.

Kijewski:
Replaced `State<'a>` with `State<'_>`. I didn't think much about it because all the functions already used lifetime argument, and replaced `Syntax<'a>` did so, too.

State cannot be passed mutable because it's captured by multiple closures. 

> Here, too, I wonder: do you have use cases for these constructs?

I used it before in Flask apps. You don't have to use break or continue in templates, since you can always put more logic into the Rust code and less in the template code, but I guess that's a matter of taste.

> Note that Jinja does not allow this:
>
>>  Unlike in Python, it’s not possible to break or continue in a loop.

After [13 years](https://github.com/pallets/jinja/blame/55848d36c0a589139e5003a5ad263fce0d95aa2f/src/jinja2/ext.py#L588), can we call an extension part of the standard? :)

djc:
If you rebase this, I'm happy to merge this. Sorry for the long review cycle!

Kijewski:
No problem. Rebased.

djc (approved):

djc:
Thanks!

Issue #522, opened 2021-07-31
Build error thanks to bitvec vs funty


leehambley:
I have a project using `askama@0.10.5` and `mysql@21.0.0` which leaves me with the following 

```
├── askama v0.10.5
│   ├── askama_derive v0.10.5 (proc-macro)
│   │   ├── askama_shared v0.11.1
│   │   │   ├── askama_escape v0.10.1
│   │   │   ├── humansize v1.1.1
│   │   │   ├── nom v6.1.1
│   │   │   │   ├── bitvec v0.19.4
│   │   │   │   │   ├── funty v1.2.0
│   │   │   │   │   ├── radium v0.5.3
│   │   │   │   │   ├── tap v1.0.1
│   │   │   │   │   └── wyz v0.2.0
│   │   │   │   ├── lexical-core v0.7.5
│   │   │   │   │   ├── arrayvec v0.5.2
│   │   │   │   │   ├── bitflags v1.2.1
│   │   │   │   │   ├── cfg-if v1.0.0
│   │   │   │   │   ├── ryu v1.0.5
│   │   │   │   │   └── static_assertions v1.1.0
│   │   │   │   └── memchr v2.3.4
│   │   │   │   [build-dependencies]
│   │   │   │   └── version_check v0.9.2
│   │   │   ├── num-traits v0.2.14
│   │   │   │   [build-dependencies]
│   │   │   │   └── autocfg v1.0.1
│   │   │   ├── percent-encoding v2.1.0
│   │   │   ├── proc-macro2 v1.0.24 (*)
│   │   │   ├── quote v1.0.9 (*)
│   │   │   ├── serde v1.0.124 (*)
│   │   │   ├── syn v1.0.61 (*)
│   │   │   └── toml v0.5.8
│   │   │       └── serde v1.0.124 (*)
│   │   ├── proc-macro2 v1.0.24 (*)
│   │   └── syn v1.0.61 (*)
│   ├── askama_escape v0.10.1
│   └── askama_shared v0.11.1 (*)
├── askama_derive v0.10.5 (proc-macro) (*)
├── askama_shared v0.11.1 (*)
```

The key thing (if I understood this) is that here we have `bitvec@0.19.4`.

Elsewhere in my dependency tree I have:

```
│   ├── mysql_common v0.27.4
│   │   ├── base64 v0.13.0
│   │   ├── bigdecimal v0.2.0
│   │   │   ├── num-bigint v0.3.2
[...]

PR #523, opened 2021-07-31
Upgrade nom to 6.2.1 (bitvec & funty issue)

leehambley:
More information here https://github.com/djc/askama/issues/522.

I haven't checked the contributing guidelines here if contributors are expected to also send changes to versions, changelogs, etc - but I will follow-up in this PR over the next hours/days.

### DIFF
--- a/askama_shared/Cargo.toml
+++ b/askama_shared/Cargo.toml
@@ -19,8 +19,7 @@ yaml = ["serde", "serde_yaml"]
 [dependencies]
 askama_escape = { version = "0.10", path = "../askama_escape" }
 humansize = { version = "1.1.0", optional = true }
-funty = "=1.1.0" # Temporary fix for bitvec, remove once fixed. (https://github.com/bitvecto-rs/bitvec/issues/105)
-nom = { version = "6", features = ["std"], default-features = false }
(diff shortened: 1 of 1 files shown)


djc:
I don't quite want to pull in the alpha. Let's stick to 6.2.1 if that fixes things.

leehambley:
6.2.1 would allow you to pin `nom`, rather than `funty`, however it cannot coexist with the MySQL crate which forces funty at 1.2.0.

Unfortunately, it seems I need the alpha specifically, although askama might benefit from the (slightly) improved Cargo.toml iny first commit Herr.

djc:
I don't really want to depend on an alpha there, given the [issues](https://internals.rust-lang.org/t/changing-cargo-semver-compatibility-for-pre-releases/14820) with depending on pre-release versions. I've tried to [lobby](https://github.com/Geal/nom/issues/1323#issuecomment-890843025) upstream nom a bit more.

I'd be happy to take your first commit though.

leehambley:
I suspect it's easier for you to simply cherry-pick it, or I'm happy to close this PR and send one for just the first commit. 

Just let me know what you prefer :+1:

djc:
If you just `git reset --hard HEAD^ && git push -f`, that should take care of things.

leehambley:
Done, once I see that you are merged here, I'll go back to my `af7b1ab ` version where I have the alpha, so I can use my fork until this upstream nonsesne resolves itself. Thanks @djc

djc (approved):

PR #527, opened 2021-08-05
Remove the askama_iron integration

djc:
This web framework seems not to have been updated for quite a while,
and its current version appears to depend on vulnerable crates.
Remove it for now. If someone wants to fix up Iron upstream and
reinstate this, I'd be happy to review your PR.


vallentin (approved):
Makes sense. Should the release be yanked?

djc:
I don't think so. Anyone using Iron is already vulnerable, and it would probably be useful if askama_iron stays working with Iron 0.6.2 -- we just don't have to maintain it going forward.

vallentin:
Yeah, understandable. I was just checking :)

PR #531, opened 2021-08-24
Add test case for matching on Option<bool>

Restioson:
This adds a (currently failing) test for `{% match %}`ing on `Option<bool>` with a `Some(true)` pattern. Since https://github.com/djc/askama/commit/268d8250fb0a9cdcbbd760bdf39424ed02fd1920, this emits `Some(r#true)` (matching any `Some` and storing the boolean value into a variable named `true`) which is a bug, as before it emitted `Some(true)` as expected (matching only `Some` where the value is `true`). CC @Kijewski, author of the commit which I think introduced this bug after bisecting.

### DIFF
--- a/testing/tests/matches.rs
+++ b/testing/tests/matches.rs
@@ -26,6 +26,24 @@ fn test_match_option() {
     assert_eq!(s.render().unwrap(), "\nNot Found\n");
 }
 
+#[derive(Template)]
+#[template(path = "match-opt-bool.html")]
+struct MatchOptBoolTemplate {
+    item: Option<bool>
+}
+
+#[test]
+fn test_match_option_bool() {
+    let s = MatchOptBoolTemplate { item: Some(true) };
+    assert_eq!(s.render().unwrap(), "\nFound Some(true)\n");
+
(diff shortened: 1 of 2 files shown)


Kijewski:
Thank you for bisecting the error! In `target` only string, char and number literals are matched. I'll open a PR fixing the problem.

testing/tests/matches.rs
```diff
@@ -26,6 +26,24 @@ fn test_match_option() {
     assert_eq!(s.render().unwrap(), "\nNot Found\n");
 }
 
+#[derive(Template)]
+#[template(path = "match-opt-bool.html")]
+struct MatchOptBoolTemplate {
+    item: Option<bool>
```
Kijewski, line 32: Could you please add a comma to the end of the line to make "cargo fmt" happy? You can test it the linter wit `cargo fmt --all -- --check`.

Kijewski (changes requested):

testing/tests/matches.rs
Restioson: Done

PR #532, opened 2021-08-25
 Parse boolean literals in assignment targets 

Kijewski:
268d825 introduced a regression that made matching against boolean
literals impossible. E.g. "true" was interpreted as the variable
"r#true". This PR fixes the problem.

The bug was reported by @Restioson in issue #531.


vallentin:
LGTM. However, you should squash the "Add test" and "Fixup" commits to just a single "Add test" commit. Along with flipping the order, so the "Add test" commit comes after. Thus ensuring all commits build and pass CI :)

djc (dismissed):
Thanks for the quick response! @vallentin is saying exactly what I was wondering about (though I do get some reluctance to amend someone else's commit, the triviality of the fix up change seems to make this fine).

Kijewski:
>  Along with flipping the order, so the "Add test" commit comes after.

Ah, I didn't even think about that. Flipped.

I combined your two commit, @Restioson, I hope that's okay.

PR #537, opened 2021-10-05
Offload #[template(...)] attribute parsing to `darling` crate

TedDriggs:
`darling` is "serde for the AST." Its main advantages are:

* Errors are now more-precisely attached to the input AST
* Remove boilerplate for attribute parsing

This PR is a fairly minimal change; if this is directionally acceptable, there are a couple additions that a future PR could make:

- Using `#[darling(supports(struct_named))]` to only allow derivation on named structs
- Retaining span information in `TemplateOptions` to make the lookup failures (e.g. syntax, escaper) point to the correct input span

Disclaimer: I'm the author and maintainer of `darling`.

### DIFF
--- a/askama_derive/src/lib.rs
+++ b/askama_derive/src/lib.rs
@@ -17,6 +17,9 @@ pub fn derive_template(input: TokenStream) -> TokenStream {
     let ast: syn::DeriveInput = syn::parse(input).unwrap();
     match build_template(&ast) {
         Ok(source) => source.parse().unwrap(),
+        // Use precise span information if available, otherwise fall back to placing
+        // the error at the macro call site.
+        Err(CompileError::Darling(e)) => e.write_errors().into(),
         Err(e) => syn::Error::new(Span::call_site(), e)
             .to_compile_error()
             .into(),
--- a/askama_shared/Cargo.toml
+++ b/askama_shared/Cargo.toml
@@ -18,6 +18,7 @@ yaml = ["serde", "serde_yaml"]
 
 [dependencies]
 askama_escape = { version = "0.10", path = "../askama_escape" }
+darling = "0.13.0"
 humansize = { version = "1.1.0", optional = true }
 nom = "7"
 num-traits = { version = "0.2.6", optional = true }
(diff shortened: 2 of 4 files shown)


djc:
Hey, thanks for this, and thanks for building darling!

I've adopted darling recently for one of my other projects (the Mendes web framework), however I have been wondering about the compile-time impact of adopting it. I usually try to be intentional about not relying on dependencies that don't add much value for the end user of my projects. Have you done any benchmarking on the compile-time impact of adding darling as a dependency?

TedDriggs:
Proc macros in general increase compile times, so adding `darling` as a project's only dependency will result in a jump in compile times. However, `darling_core` is very judicious about the dependencies it brings in, so for a project already using `syn`, `quote`, and `proc-macro2`, the only new indirect dependencies added are `ident_case`, `fnv`, and `str_sim`. None of the latter set of 3 have any further dependencies, and all are quick to compile.

I have not had time to do a meaningful clean comparison of a proc-macro crate with `syn`, `quote`, and `proc-macro2` vs. a crate with `syn`, `quote`, `proc-macro2` and `darling`.

TedDriggs:
That's totally reasonable. This particular conversion was quick to implement, and I posted it in part to get feedback before exploring the more involved work of plumbing span information down to the extension and syntax lookups later on.

FYI, under the implementation on `main`, splitting the meta-items across multiple attribute tags wouldn't work, e.g.

```rust
#[derive(Template)]
#[template(source = "...", ext = "...")]
#[template(escape = "...")] // SILENTLY IGNORED
pub struct Sample {}
```

It also looks like non-meta attributes, even with a path other than `template`, will break the macro; note that `parse_meta` is called _before_ checking `if attr.path.is_ident("template")`, rather than being called after that check. The following code would therefore fail:

```rust
#[derive(Template, Bad)]
#[bad(x = foo::bar)]
#[template(path = "if-coerce.html")]
struct IfCoerceTemplate {}
```

I can file those as separate issues if they're something you want to address.

djc:
Yes, if you can file separate issues about those that would be great!

PR #538, opened 2021-10-07
Handle build hangs due to unintended template recursion.

grv07:
Created a vector to keep track of source and extends dependency.
If the same dependency comes again throw the Err as there is some cyclic dependency between source and extends file.
 
Case 1: When there is a self-loop in dependency
self_loop

Case 2: When there is a cyclic dependency
cyclic_deps


grv07:
I have done some force-pushed that leads to disturb the file changes and add unwanted changes.
Fix that in the same PR will make it difficult to read.
So, closing it for now and will open again tomorrow.

Thanks,

PR #541, opened 2021-10-11
Parse `&str` instead of `&[u8]`

Kijewski:
Askama's takes valid UTF-8 files as input. So why operate on byte slices
instead of strings? This makes writing some functions a lot simpler.

### DIFF
--- a/askama_shared/src/parser.rs
+++ b/askama_shared/src/parser.rs
@@ -2,13 +2,13 @@ use std::cell::Cell;
 use std::str;
 
 use nom::branch::alt;
-use nom::bytes::complete::{escaped, is_not, tag, take_until};
+use nom::bytes::complete::{escaped, is_not, tag, take_till, take_until};
 use nom::character::complete::{anychar, char, digit1};
 use nom::combinator::{complete, cut, map, opt, recognize, value};
-use nom::error::{Error, ErrorKind, ParseError};
+use nom::error::{Error, ErrorKind};
 use nom::multi::{fold_many0, many0, many1, separated_list0, separated_list1};
 use nom::sequence::{delimited, pair, preceded, terminated, tuple};
-use nom::{self, error_position, Compare, IResult, InputLength, InputTake};
+use nom::{self, IResult, error_position};
 
 use crate::{CompileError, Syntax};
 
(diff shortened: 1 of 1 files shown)


Kijewski:
I refactored the commits and opened PR #545. This PR is now based on #545.

askama_shared/src/parser.rs
```diff
@@ -128,6 +128,37 @@ pub struct CondTest<'a> {
...
+                let skipped = &start[..start.offset(i)];
+                return Ok((j, (skipped, o)));
+            } else {
+                i = inner(i)?.0;
+            }
+        }
+    }
+}
+
+/// Succeeds if parser matches, but the input is not consumed.
```
djc, line 151: Is this different from https://docs.rs/nom/7.1.0/nom/combinator/fn.peek.html?

askama_shared/src/parser.rs
```diff
@@ -128,6 +128,37 @@ pub struct CondTest<'a> {
     pub expr: Expr<'a>,
 }
 
+/// Skips over the any amout of `inner` until `end` was found.
```
djc, line 131: "the any amout" -> "any amount"

askama_shared/src/parser.rs
```diff
@@ -128,6 +128,37 @@ pub struct CondTest<'a> {
...
+/// Returns a tuple of the skipped string and the found end marker.
+fn skip_till<'a, I, O>(
+    mut inner: impl FnMut(&'a [u8]) -> IResult<&'a [u8], I>,
```
djc: I think `take_content()` is now the only caller for `skip_till`, which means `inner` is always `anychar`. Can we simplify? Maybe even inline `skip_till()` into `take_content()`?

djc (commented):
Again sorry for the slow review. Your contributions are highly valued, I've just found it hard to find the time to spend enough attention on these nuanced changes.

askama_shared/src/parser.rs
Kijewski: Huh, I swear I looked for this function but I didn't find it.

djc:
@Kijewski do you have a moment to work on this soon? Otherwise I'll probably take it over, I think it's the last blocker to 0.11.

Kijewski:
I rebased #541 and #546.

askama_shared/src/parser.rs
Kijewski: The function is used in #546, too.
I removed the "inner" parameter, because it's "anychar" for both its uses.
I cleaned up the implementation a bit. I think it's much easier on the eyes now. :)

djc (approved):
Looks great, thanks for all the work on this!

PR #549, opened 2021-10-13
Don't parse non-template attributes

### LINKED ISSUE #543: Non-meta attributes (e.g. #[bad(x = foo::bar)]) cause parsing to fail unnecessarily
Non-meta attributes, even with a path other than template, will break the macro; note that parse_meta is called before checking if attr.path.is_ident("template"), rather than being called after that check. The following code would therefore fail:

```rust
#[derive(Template, Bad)]
#[bad(x = foo::bar)]
#[template(path = "if-coerce.html")]
struct IfCoerceTemplate {}
```

Filed based on discussion in #537

djc:
Fixes #543.

### DIFF
--- a/askama_shared/src/input.rs
+++ b/askama_shared/src/input.rs
@@ -31,15 +31,11 @@ impl<'a> TemplateInput<'a> {
         let meta = ast
             .attrs
             .iter()
-            .find_map(|attr| match attr.parse_meta() {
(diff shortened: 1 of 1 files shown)


djc:
@vallentin would you mind approving this again?

vallentin (approved):

PR #558, opened 2021-11-15
Use strict matching for prereleases

djc:


vallentin (approved):
I assume the same failing CI is to be ignored ofc :)

djc:
No, these changes are supposed to fix the CI failure...

PR #560, opened 2021-11-18
 Added support for optional escaper for escape filter

### LINKED ISSUE #556: escape filter does not escape when escape attribute is set to none
Perhaps this is just a documentation issue but following the recursive macro workaround described in https://github.com/djc/askama/issues/105#issuecomment-483937352 via #308 I annotated my struct as follows:

```rust
#[derive(Debug, Template)]
#[template(path = "topic/node.html", escape = "none")]
pub struct TopicNode {
    pub topic: Topic,
    pub children: TopicTree,
}
```

With a template that looks like this:

```jinja
<li>{{ topic.name }}</li>
{% if !children.is_empty() %}
<li>
    <ul class="tree-list">
        {% for child in children %}
            {{ child.render().unwrap() }}
        {% endfor %}
    </ul>
</li>
{% endif %}
```

I noticed that the name in the first line was no longer being escaped... not unexpected given `escape = "none"` so I tried to manually escape by doing:

```jinja
<li>{{ topic.name|escape }}</li>
```

which still did not escape the text. It seems the `E` argument of the `escape` filter is filled in with the current escaper, which in this case in `none`. I was able to work around this by defining a custom filter as follows:

```rust
pub fn html_escape<T>(v: T) -> askama::Result<MarkupDisplay<Html, T>>
    where
        T: fmt::Display,
{
    Ok(MarkupDisplay::new_unsafe(v, Html))
}
```

Which, works as expected. Anyway, [the docs](https://djc.github.io/askama/filters.html#escape--e) suggest that HTML will always be escaped so I thought it was worth opening an issue. Thanks.

vallentin:
Resolves #556

- Added support for optional escaper arg for `escape filter`, e.g. `{{ foo|escape("html") }}` and `{{ foo|e("html") }}`
- Added tests
- Updated book

Additionally, I removed `fn e()`, which is technically a breaking change, so I added a `pub use escape as e;`. Maybe it's _safe_ to assume _nobody_ depends on `askama_shared` directly. Only to `use askama_shared::filters::e;`, but I'll let you decide on that, then I'll remove it if needed.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -1091,7 +1091,10 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         mut name: &str,
         args: &[Expr<'_>],
     ) -> Result<DisplayWrap, CompileError> {
-        if name == "format" {
+        if matches!(name, "escape" | "e") {
+            self._visit_escape_filter(buf, args)?;
+            return Ok(DisplayWrap::Wrapped);
+        } else if name == "format" {
             self._visit_format_filter(buf, args)?;
             return Ok(DisplayWrap::Unwrapped);
         } else if name == "fmt" {
@@ -1115,7 +1118,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             return Err("the `yaml` filter requires the `serde-yaml` feature to be enabled".into());
         }
 
-        const FILTERS: [&str; 5] = ["safe", "escape", "e", "json", "yaml"];
+        const FILTERS: [&str; 3] = ["safe", "json", "yaml"];
         if FILTERS.contains(&name) {
             buf.write(&format!(
                 "::askama::filters::{}({}, ",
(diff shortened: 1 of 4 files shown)


askama_shared/src/generator.rs
```diff
@@ -1135,6 +1138,39 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+            Some(_) => return Err("invalid escaper type for escape filter".into()),
+            None => None,
+        };
+        let escaper = match opt_escaper {
+            Some(name) => self
+                .input
+                .config
+                .escapers
+                .iter()
+                .filter(|(escapers, _)| escapers.contains(name))
```
djc, line 1160: Instead of `.filter().map().next()` can we just use `find()`?

djc (commented):
Good stuff! One style nit:

askama_shared/src/generator.rs
vallentin: Yeah sure, using `.find()` reduces it by a line :)

```rust
let escaper = match opt_escaper {
    Some(name) => self
        .input
        .config
        .escapers
        .iter()
        .find(|(escapers, _)| escapers.contains(name))
        .map(|(_, escaper)| escaper)
        .ok_or(CompileError::Static("invalid escaper for escape filter"))?,
    None => self.input.escaper,
};
```

Actually, what version are we targetting? If we allow Rust 1.50.0, then we can use [`.then()`](https://doc.rust-lang.org/std/primitive.bool.html#method.then) and reduce it with an additional line more.

```rust
let escaper = match opt_escaper {
    Some(name) => self
        .input
        .config
        .escapers
        .iter()
        .find_map(|(escapers, escaper)| escapers.contains(name).then(|| escaper))
        .ok_or(CompileError::Static("invalid escaper for escape filter"))?,
    None => self.input.escaper,
};
```

vallentin:
Also, just to make sure. Does that mean you want to keep the re-export? (`pub use escape as e;`)

askama_shared/src/generator.rs
djc: Yeah, requiring 1.50.0 for `bool::then()` sounds good.

djc:
I think it'd be better to just ditch the public `e` function if we no longer need it to generate code.

djc (approved):

djc:
Thanks for getting this done!

PR #561, opened 2021-11-25
Move askama_mendes integration into Askama repo

djc:
This previously lived in the mendes repo, which was weird since there's also a part that lives in the code generator.

### DIFF
--- a/askama_mendes/Cargo.toml
+++ b/askama_mendes/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "askama_mendes"
+version = "0.1.0"
+description = "Mendes integration for Askama templates"
+documentation = "https://docs.rs/askama"
+keywords = ["markup", "template", "jinja2", "html"]
+categories = ["template-engine"]
+homepage = "https://github.com/djc/askama"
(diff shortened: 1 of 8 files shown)


askama_mendes/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "askama_mendes"
+version = "0.1.0"
+authors = ["Dirkjan Ochtman <dirkjan@ochtman.nl>"]
```
Kijewski, line 4: In #562 you removed the author fields.

djc: Yeah, but this PR is independent of that one. Depending on which one lands first, I'll fix up the other one.

Kijewski: Hehe, now this is the second PR to land. :) Approved after you cut this one line.

Kijewski (changes requested):

Kijewski:
Hmm, the debug message for errors was changed. It now includes leading spaces for no obvious reason.

Kijewski (approved):

PR #562, opened 2021-11-25
Prepare 0.12

djc:
Since the Actix and Rocket integrations currently depend on pre-release versions of their respective frameworks, I'm inclined to just defer releasing new version of these integrations until they get their releases out.


Kijewski (approved):
LGTM

PR #563, opened 2021-11-28
Add Axum integration

malyn:
This integration is based on [a sample in the Axum repo](https://github.com/tokio-rs/axum/blob/main/examples/templates/src/main.rs) which I modified to fit the structure and layout of the other Askama integrations. I tested this in a small project using a variety of templates and everything works the same as with the Tide integration (which I have used previously).

I see that an 0.12.0 version of Askama is imminent; I based this PR on the version in main, but am happy to rebase onto the 0.12 branch if you would prefer.


djc (commented):
Thanks, this looks great!

djc (approved):

PR #564, opened 2021-11-30
Fix tests for new error messages in Rust nightly

Kijewski:


Kijewski:
Ah, beta was changed, too, not only nightly.

djc (approved):

djc:
Thanks for fixing that up!

PR #579, opened 2021-12-15
Use a separate trait for object safety

djc:
This is a relatively major change to the main trait's API. For context,
I always started from the concept of monomorphized traits, but later
several contributors asked about object safety. At that point I made
`Template` object-safe, and then even later added a `SizedTemplate`
to make some things easier for people who don't need object safety.

However, having object-safety in the primary trait is bad for
performance (a substantial number of calls into the virtual `Write`
trait is relatively slow), and I don't think those who don't need
object safety should pay for the cost of having it.

Additionally, I feel using associated consts for the extension and
size hint is more idiomatic than having accessor methods. I don't
know why I didn't use these from the start -- maybe associated
consts didn't exist yet, or I didn't yet know how/when to use them.
Askama is pretty old at this point...

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -73,26 +73,59 @@ use std::path::Path;
 pub use askama_escape::{Html, Text};
 
 /// Main `Template` trait; implementations are generally derived
+///
+/// If you need an object-safe template, use [`DynTemplate`].
 pub trait Template {
     /// Helper method which allocates a new `String` and renders into it
     fn render(&self) -> Result<String> {
-        let mut buf = String::with_capacity(self.size_hint());
+        let mut buf = String::with_capacity(Self::SIZE_HINT);
         self.render_into(&mut buf)?;
         Ok(buf)
     }
+
     /// Renders the template to the given `writer` buffer
-    fn render_into(&self, writer: &mut dyn std::fmt::Write) -> Result<()>;
+    fn render_into(&self, writer: &mut (impl std::fmt::Write + ?Sized)) -> Result<()>;
+
+    /// The template's extension, if provided
+    const EXTENSION: Option<&'static str>;
+
+    /// Provides a conservative estimate of the expanded length of the rendered template
+    const SIZE_HINT: usize;
+}
+
+/// Object-safe wrapper trait around [`Template`] implementers
+///
+/// This trades reduced performance (mostly due to writing into `dyn Write`) for object safety.
+pub trait DynTemplate {
+    /// Helper method which allocates a new `String` and renders into it
+    fn dyn_render(&self) -> Result<String>;
+
+    /// Renders the template to the given `writer` buffer
+    fn dyn_render_into(&self, writer: &mut dyn std::fmt::Write) -> Result<()>;
+
     /// Helper function to inspect the template's extension
     fn extension(&self) -> Option<&'static str>;
-    /// Provides an conservative estimate of the expanded length of the rendered template
+
+    /// Provides a conservative estimate of the expanded length of the rendered template
     fn size_hint(&self) -> usize;
 }
 
-pub trait SizedTemplate {
-    /// Helper function to inspect the template's extension
-    fn extension() -> Option<&'static str>;
-    /// Provides an conservative estimate of the expanded length of the rendered template
-    fn size_hint() -> usize;
+impl<T: Template> DynTemplate for T {
+    fn dyn_render(&self) -> Result<String> {
+        <Self as Template>::render(self)
+    }
+
+    fn dyn_render_into(&self, writer: &mut dyn std::fmt::Write) -> Result<()> {
+        <Self as Template>::render_into(self, writer)
+    }
+
+    fn extension(&self) -> Option<&'static str> {
+        Self::EXTENSION
+    }
+
+    fn size_hint(&self) -> usize {
+        Self::SIZE_HINT
+    }
 }
 
 pub use crate::shared::filters;
@@ -137,3 +170,31 @@ pub mod mime {
     note = "file-level dependency tracking is handled automatically without build script"
 )]
 pub fn rerun_if_templates_changed() {}
+
+#[cfg(test)]
+mod tests {
+    use super::{DynTemplate, Template};
+
+    #[test]
+    fn dyn_template() {
+        struct Test;
+        impl Template for Test {
+            fn render_into(
+                &self,
+                writer: &mut (impl std::fmt::Write + ?Sized),
+            ) -> askama_shared::Result<()> {
+                Ok(writer.write_str("test")?)
+            }
+
+            const EXTENSION: Option<&'static str> = Some("txt");
+
+            const SIZE_HINT: usize = 4;
+        }
+
+        fn render(t: &dyn DynTemplate) -> String {
+            t.dyn_render().unwrap()
+        }
+
+        assert_eq!(render(&Test), "test");
+    }
+}
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -128,7 +128,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     ) -> Result<(), CompileError> {
         self.write_header(buf, "::askama::Template", None)?;
         buf.writeln(
-            "fn render_into(&self, writer: &mut dyn ::std::fmt::Write) -> \
+            "fn render_into(&self, writer: &mut (impl ::std::fmt::Write + ?Sized)) -> \
              ::askama::Result<()> {",
         )?;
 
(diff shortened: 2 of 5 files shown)


Kijewski:
Maybe lose the `Template` requirement for `DynTemplate`? E.g.

```rs
pub trait DynTemplate {
    /// Helper method which allocates a new `String` and renders into it
    fn dyn_render(&self) -> Result<String>;

    /// Renders the template to the given `writer` buffer
    fn dyn_render_into(&self, writer: &mut dyn std::fmt::Write) -> Result<()>;

    /// Helper function to inspect the template's extension
    fn extension(&self) -> Option<&'static str>;

    /// Provides an conservative estimate of the expanded length of the rendered template
    fn size_hint(&self) -> usize;
}

impl<T: Template> DynTemplate for T {
    fn dyn_render(&self) -> Result<String> {
        <Self as Template>::render(self)
    }

    fn dyn_render_into(&self, writer: &mut dyn std::fmt::Write) -> Result<()> {
        <Self as Template>::render_into(self, writer)
    }

    fn extension(&self) -> Option<&'static str> {
        Self::EXTENSION
    }

    fn size_hint(&self) -> usize {
        Self::SIZE_HINT
    }
}
```

Otherwise DynTemplate isn't object-safe because all super-traits need to be object-safe, or am I misunderstanding the requirement in https://doc.rust-lang.org/reference/items/traits.html#object-safety?

djc:
Ohh, good call, thanks! I thought we had some tests for this...

askama/src/lib.rs
```diff
@@ -76,23 +76,50 @@ pub use askama_escape::{Html, Text};
 pub trait Template {
     /// Helper method which allocates a new `String` and renders into it
     fn render(&self) -> Result<String> {
-        let mut buf = String::with_capacity(self.size_hint());
+        let mut buf = String::with_capacity(Self::SIZE_HINT);
         self.render_into(&mut buf)?;
         Ok(buf)
     }
     /// Renders the template to the given `writer` buffer
```
Kijewski, line 83: Could to add an empty line before the doc string?

askama/src/lib.rs
```diff
@@ -76,23 +76,50 @@ pub use askama_escape::{Html, Text};
...
-    fn render_into(&self, writer: &mut dyn std::fmt::Write) -> Result<()>;
+    fn render_into<W: std::fmt::Write + ?Sized>(&self, writer: &mut W) -> Result<()>;
+
+    /// The template's extension, if provided
+    const EXTENSION: Option<&'static str>;
```
vallentin, line 87: Do we have a preference for item order? For instance, in `impl`s, I usually do `type`, then `const`, then all `fn`s. The `type` then `fn` seems to be _standard_ in Rust. For consistency, should be place `const` first, last, anywhere?

askama/src/lib.rs
```diff
@@ -76,23 +76,50 @@ pub use askama_escape::{Html, Text};
...
+
+    /// Provides an conservative estimate of the expanded length of the rendered template
+    const SIZE_HINT: usize;
+}
+
+pub trait DynTemplate {
+    fn dyn_render(&self) -> Result<String>;
```
Kijewski, line 94: Maybe copy the Template's doc strings verbatim for the dyn_render() and dyn_render_into()?

askama/src/lib.rs
```diff
@@ -76,23 +76,50 @@ pub use askama_escape::{Html, Text};
...
+impl<T: Template> DynTemplate for T {
+    fn dyn_render(&self) -> Result<String> {
+        <Self as Template>::render(self)
+    }
+
+    fn dyn_render_into(&self, writer: &mut dyn std::fmt::Write) -> Result<()> {
+        <Self as Template>::render_into(self, writer)
+    }
+
     /// Helper function to inspect the template's extension
```
Kijewski, line 114: No need for doc strings in the implementation.

askama/src/lib.rs
```diff
@@ -76,23 +76,50 @@ pub use askama_escape::{Html, Text};
...
-    fn extension() -> Option<&'static str>;
+    fn extension(&self) -> Option<&'static str> {
```
vallentin, line 115: Should we mark these as `#[deprecated]`, and point to using `::EXTENSION` and `::SIZE_HINT`?

askama/src/lib.rs
```diff
@@ -76,23 +76,50 @@ pub use askama_escape::{Html, Text};
...
         Ok(buf)
     }
     /// Renders the template to the given `writer` buffer
-    fn render_into(&self, writer: &mut dyn std::fmt::Write) -> Result<()>;
+    fn render_into<W: std::fmt::Write + ?Sized>(&self, writer: &mut W) -> Result<()>;
+
+    /// The template's extension, if provided
+    const EXTENSION: Option<&'static str>;
+
+    /// Provides an conservative estimate of the expanded length of the rendered template
```
Kijewski, line 89: s/an/a/

Kijewski (changes requested):
The change looks sensible to me. :+1:

vallentin:
> I feel using associated consts for the extension and
size hint is more idiomatic than having accessor methods. I don't
know why I didn't use these from the start -- maybe associated
consts didn't exist yet

Your assumption is right, they weren't stable yet. But yes, it seems to be more idiomatic to use associated constants, a bunch of `::consts` modules, have also moved into `f32`, `i32`, etc types, in the standard library

I'm still waiting for `associated_type_defaults`, I always randomly need it :)

askama/src/lib.rs
djc: They shouldn't be deprecated, because you can't access the associated consts from a `dyn DynTemplate`. Since `DynTemplate` is new, people using the no-argument form will get an error anyway and will hopefully be able to figure out they should use the associated consts.

askama/src/lib.rs
djc: Good point, done.

askama/src/lib.rs
djc: Yup.

askama/src/lib.rs
djc: Indeed!

askama/src/lib.rs
djc: Done.

Kijewski:
I don't know if this diff is too ugly, but it could prevent accidental shadowing of the identifier `W` if it's used in the user's code:

```diff
diff --git a/askama/src/lib.rs b/askama/src/lib.rs
index c116f55..185e8cc 100644
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -84,7 +84,7 @@ pub trait Template {
     }
 
     /// Renders the template to the given `writer` buffer
-    fn render_into<W: std::fmt::Write + ?Sized>(&self, writer: &mut W) -> Result<()>;
+    fn render_into(&self, writer: &mut (impl std::fmt::Write + ?Sized)) -> Result<()>;
 
     /// The template's extension, if provided
     const EXTENSION: Option<&'static str>;
@@ -179,9 +179,9 @@ mod tests {
     fn dyn_template() {
         struct Test;
         impl Template for Test {
-            fn render_into<W: std::fmt::Write + ?Sized>(
+            fn render_into(
                 &self,
-                writer: &mut W,
+                writer: &mut (impl std::fmt::Write + ?Sized),
             ) -> askama_shared::Result<()> {
                 Ok(writer.write_str("test")?)
             }
diff --git a/askama_shared/src/generator.rs b/askama_shared/src/generator.rs
index bf553fb..ad7e1fe 100644
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -128,7 +128,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     ) -> Result<(), CompileError> {
         self.write_header(buf, "::askama::Template", None)?;
         buf.writeln(
-            "fn render_into<W: ::std::fmt::Write + ?Sized>(&self, writer: &mut W) -> \
+            "fn render_into(&self, writer: &mut (impl ::std::fmt::Write + ?Sized)) -> \
              ::askama::Result<()> {",
         )?;
```

djc:
A manual implementation of `Template` of a type with a type parameter named `W` seems pretty unlikely, but it doesn't hurt. 👍

askama/src/lib.rs
djc: I added some more docs to both trait definitions, hopefully that will help.

askama/src/lib.rs
vallentin: Oh, interesting. Personally, I order stuff in groups, and then within each group I order by visibility, so `pub` comes first. I recall [this Reddit comment](https://www.reddit.com/r/rust/comments/453ipw/comment/czuwj4z/?context=3), it's 90% of how I order my code, in my own projects.

That said, I don't really care what the order is. It was more a question of, if there was an order, so that I know it going forward. If you prefer `const` at the bottom, then that's fine with me :)

askama/src/lib.rs
vallentin: True

djc:
Thanks for all the feedback! I think this is ready to go, and then I'll likely do a release soon.

askama_shared/src/generator.rs
```diff
@@ -160,25 +160,13 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         buf.writeln("Ok(())")?;
         buf.writeln("}")?;
 
-        buf.writeln("fn extension(&self) -> Option<&'static str> {")?;
+        buf.writeln("const EXTENSION: Option<&'static str> = ")?;
```
Kijewski, line 163: `::std::option::Option`

askama_shared/src/generator.rs
```diff
@@ -160,25 +160,13 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
         buf.writeln(&format!("{:?}", self.input.extension()))?;
-        buf.writeln("}")?;
+        buf.writeln(";")?;
 
-        buf.writeln("fn size_hint(&self) -> usize {")?;
+        buf.writeln("const SIZE_HINT: usize = ")?;
```
Kijewski, line 167: `::std::primitive::usize`? (but this might be overzealous)

Kijewski (changes requested):
LGTM except for the two last comments.

Kijewski (approved):

djc:
I was waiting for @vallentin's okay, but I guess it's fine in this case. :)

vallentin:
I didn't have any additional "deny comments", which is why I just added regular comments. So it was fine with me :)

PR #583, opened 2021-12-21
askama_axum: prepare version 0.1.0

djc:


askama_axum/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "askama_axum"
-version = "0.10.0"
+version = "0.1.0"
```
Kijewski, line 3: Um, I guess you meant to hit the "1" key one more often

Kijewski (changes requested):

askama_axum/Cargo.toml
djc: I actually didn't... This is a new package that hasn't been published before, so we can just start from 0.1.0.

djc:
The failure here is unrelated to the changes; it will be fixed by #585.

djc:
Ping? Would be nice to get this out.

Kijewski (approved):
Looks good to me!

PR #584, opened 2021-12-21
Revert Actix and Rocket integrations to stable versions for release

djc:


Kijewski:
I only looked at the CI messages, so I guess some `?`  is missing somewhere. :-/

Kijewski:
But yes, it should be reverted to their respective stable versions.

djc:
The failure here is unrelated to the changes; it will be fixed by #585.

djc:
Ping? Would be nice to get these out.

askama_actix/src/lib.rs
```diff
@@ -3,23 +3,23 @@
...
+use actix_web::{error::ErrorInternalServerError, Error, HttpResponse};
 
 pub trait TemplateToResponse {
-    fn to_response(&self) -> HttpResponse;
+    fn to_response(&self) -> std::result::Result<HttpResponse, Error>;
 }
 
 impl<T: askama::Template> TemplateToResponse for T {
-    fn to_response(&self) -> HttpResponse {
+    fn to_response(&self) -> std::result::Result<HttpResponse, Error> {
```
Kijewski, line 13: Is `std::result::` needed in here?

Kijewski:
Regardless of the question if the `std::result::` path is needed, the PR looks good to me. Even if it is not needed, it does not hurt the legibility of the return type.

Kijewski (approved):

askama_actix/src/lib.rs
djc: These are actually needed, because we import everything from `askama` and it exports a `Result` alias which we don't want to use here.

PR #590, opened 2022-01-04
 Implement error propagation expression: `?`

Kijewski:
This PR consists of ~three~ ~two~ a change~s~:

* ~Right now almost every expression needs to be parsed twice: `expr_any()` first parses the left-hand side of a range expression, and if no `..` or `..=` was found the left-hand expression is parsed again, this time as the result of the function. This diff removes the second parsing step by first looking for `.. (opt rhs)`, then for `lhs .. (opt rhs)`.~
* ~Instead of having `Expr::VarCall`, `Expr::PathCall` and `Expr::MethodCall`, this diff unifies the handling of calls by removed the former three variants, and introducing `Expr::Call`. This makes parsing postfix operators easier.~
* This change allows using the operator `?` in askama expressions. It works like the same operator in Rust: if a `Result` is `Ok`, it is unwrapped. If it is an error, then the `render()` method fails with this error value.

### DIFF
--- a/askama_shared/src/error.rs
+++ b/askama_shared/src/error.rs
@@ -1,6 +1,7 @@
 use std::fmt::{self, Display};
+use std::marker::PhantomData;
 
-pub type Result<I> = ::std::result::Result<I, Error>;
+pub type Result<I, E = Error> = ::std::result::Result<I, E>;
 
 /// askama error type
 ///
@@ -28,6 +29,8 @@ pub enum Error {
     /// formatting error
     Fmt(fmt::Error),
 
+    Custom(Box<dyn std::error::Error + Send + Sync>),
+
     /// json conversion error
     #[cfg(feature = "serde_json")]
     Json(::serde_json::Error),
@@ -41,6 +44,7 @@ impl std::error::Error for Error {
     fn cause(&self) -> Option<&dyn std::error::Error> {
         match *self {
             Error::Fmt(ref err) => err.source(),
+            Error::Custom(ref err) => Some(err.as_ref()),
             #[cfg(feature = "serde_json")]
             Error::Json(ref err) => err.source(),
             #[cfg(feature = "serde_yaml")]
@@ -51,12 +55,13 @@ impl std::error::Error for Error {
 
 impl Display for Error {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match *self {
-            Error::Fmt(ref err) => write!(formatter, "formatting error: {}", err),
+        match self {
+            Error::Fmt(err) => write!(formatter, "formatting error: {}", err),
+            Error::Custom(err) => write!(formatter, "{}", err),
             #[cfg(feature = "serde_json")]
-            Error::Json(ref err) => write!(formatter, "json conversion error: {}", err),
+            Error::Json(err) => write!(formatter, "json conversion error: {}", err),
             #[cfg(feature = "serde_yaml")]
-            Error::Yaml(ref err) => write!(formatter, "yaml conversion error: {}", err),
+            Error::Yaml(err) => write!(formatter, "yaml conversion error: {}", err),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


askama_shared/src/parser.rs
```diff
@@ -63,7 +61,7 @@ pub enum Expr<'a> {
     BinOp(&'a str, Box<Expr<'a>>, Box<Expr<'a>>),
     Range(&'a str, Option<Box<Expr<'a>>>, Option<Box<Expr<'a>>>),
     Group(Box<Expr<'a>>),
-    MethodCall(Box<Expr<'a>>, &'a str, Vec<Expr<'a>>),
+    Call(Box<Expr<'a>>, Vec<Expr<'a>>),
```
djc, line 64: Hmm, what would happen if we use a smaller enum as the first element here, that is, one that can only be `Var`, `Method`, or `Path`?

Also it seems this commit does a bunch of refactoring of parser bits that would be nice to split out into (a) separate commit(s) if at all possible -- currently it feels like it has a lot going on.

askama_shared/src/error.rs
```diff
@@ -51,12 +55,13 @@ impl std::error::Error for Error {
 
 impl Display for Error {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match *self {
-            Error::Fmt(ref err) => write!(formatter, "formatting error: {}", err),
+        match self {
+            Error::Fmt(err) => write!(formatter, "formatting error: {}", err),
+            Error::Custom(err) => write!(formatter, "formatting error: {}", err),
```
djc, line 60: This should probably have a different prefix?

djc (commented):
Haven't done an in-depth review of the final commit yet, but here's some feedback on the prerequisite refactoring steps.

askama_shared/src/error.rs
Kijewski: Do you think the prefix is needed at all? Probably I could just omit it. The contained err should have a useful message by itself.

askama_shared/src/parser.rs
Kijewski: I don't know how I could split the second commit "Unify handling of calls" into smaller parts that still compile on its own.

Kijewski: I'm it's difficult to restrict what may be on the lhs of a call: e.g. [an array](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0213078dca6afb17cd75a240932247f6), [an addition](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1e9f6d5267ce5c76276aa7b55ff2137d).

djc: Fair enough!

askama_shared/src/error.rs
djc: Seems fair.

Kijewski: Removed.

djc:
Having thought about this some more, and I'm not sure I'm comfortable with the addition of the try operator. It seems like a bunch of complexity with a goal that I'm not sure meshes well with the design of Askama -- in my opinion, templates should generally be written not to yield errors. Did you have some concrete use cases where you wanted this behavior? I don't think fully matching the Rust grammar should be a goal unto itself.

(The changes to introduce a unified `Call` AST type seems nice, though.)

Kijewski:
Right now the use of serde_json is expected at it's errors are caught in a nice way, but I guess [serde_qs](https://docs.rs/serde_qs/0.8.5/serde_qs/fn.to_string.html) is equally as useful. Well, at least I am using serde_qs to generate URLs.

Right now I am formatting the query string outside or render(), and supply it in a field. But I would like it if I could use serde_qs::to_string() directly in the template without resorting to using unwrap().

askama_shared/src/error.rs
```diff
@@ -81,6 +86,70 @@ impl From<::serde_yaml::Error> for Error {
     }
 }
 
+#[doc(hidden)]
```
djc, line 89: All this stuff seems really complex. Could we only add a `Custom(Box<dyn Error + Send + Sync + 'static>)` variant to `Error` and be done with it?

Kijewski: Of course. :)

djc (approved):

PR #591, opened 2022-01-05
 Update for actix-web beta

Kijewski:
Does this PR resolve #558?


djc (approved):
I don't think this solves the problem that #558 was looking to solve; that is, I expect our CI to break again once the next Actix beta comes out. Still, this otherwise looks good.

PR #593, opened 2022-01-05
Add more warnings

Kijewski:
Might be overkill, but it can aid preventing programming errors.

I was not able to add `#![forbid(unsafe_code)]` to askama_escape without
a 15% performance drop, so I omitted it.

### DIFF
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -59,16 +59,9 @@
 //!   in the configuration file. The default syntax , "default",  is the one
 //!   provided by Askama.
 
-#![allow(unused_imports)]
 #![deny(elided_lifetimes_in_paths)]
 
-#[macro_use]
-extern crate askama_derive;
 pub use askama_shared as shared;
-
-use std::fs::{self, DirEntry};
-use std::io;
-use std::path::Path;
 
 pub use askama_escape::{Html, Text};
 
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -59,6 +59,7 @@
 //!   in the configuration file. The default syntax , "default",  is the one
 //!   provided by Askama.
 
+#![forbid(unsafe_code)]
 #![deny(elided_lifetimes_in_paths)]
 #![deny(unreachable_pub)]
 
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -60,6 +60,7 @@
 //!   provided by Askama.
 
 #![deny(elided_lifetimes_in_paths)]
+#![deny(unreachable_pub)]
 
 pub use askama_shared as shared;
 
--- a/askama_actix/src/lib.rs
+++ b/askama_actix/src/lib.rs
@@ -2,11 +2,9 @@
 
 use actix_web::body::BoxBody;
 use actix_web::http::StatusCode;
-use actix_web::HttpResponseBuilder;
+use actix_web::{HttpResponse, HttpResponseBuilder, ResponseError};
 use askama::mime::extension_to_mime_type;
 pub use askama::*;
-
-use actix_web::HttpResponse;
 
 pub trait TemplateToResponse {
     fn to_response(&self) -> HttpResponse<BoxBody>;
--- a/askama_actix/src/lib.rs
+++ b/askama_actix/src/lib.rs
@@ -1,3 +1,4 @@
+#![forbid(unsafe_code)]
 #![deny(elided_lifetimes_in_paths)]
 #![deny(unreachable_pub)]
 
--- a/askama_actix/src/lib.rs
+++ b/askama_actix/src/lib.rs
@@ -1,4 +1,5 @@
 #![deny(elided_lifetimes_in_paths)]
+#![deny(unreachable_pub)]
 
 use std::fmt;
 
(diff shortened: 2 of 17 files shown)


djc:
Hmm, not really a fan. I feel like it will generally not prevent programming errors (as most such lints are enabled by default anyway), and instead we have to keep these lists in sync and have bikeshed discussions about which lints should get included. But maybe we can keep maybe keep most of the actual code changes here?

askama/src/lib.rs
```diff
@@ -59,17 +59,23 @@
 //!   in the configuration file. The default syntax , "default",  is the one
 //!   provided by Askama.
 
-#![allow(unused_imports)]
```
djc, line 62: I forget why we have `unused_imports` here, if we can get rid of it that certainly seems like a win.

askama/src/lib.rs
```diff
@@ -59,17 +59,23 @@
...
+#![warn(non_ascii_idents)]
+#![warn(noop_method_call)]
+#![warn(single_use_lifetimes)]
+#![warn(trivial_casts)]
+#![warn(unreachable_pub)]
+#![warn(unused_extern_crates)]
+#![warn(unused_lifetimes)]
 
-#[macro_use]
-extern crate askama_derive;
```
djc, line 66: I'm happy to get rid of this.

askama_actix/src/lib.rs
```diff
@@ -12,6 +27,24 @@ pub trait TemplateToResponse {
     fn to_response(&self) -> HttpResponse<BoxBody>;
 }
 
+struct WrappedError(Error);
```
djc, line 30: What problem is this solving?

askama_shared/src/error.rs
```diff
@@ -37,6 +37,19 @@ pub enum Error {
     Yaml(::serde_yaml::Error),
 }
 
+#[cfg(not(any(feature = "serde_json", feature = "serde_yaml")))]
```
djc, line 40: Features are supposed to be additive, so I don't think this is a good idea.

askama_shared/src/helpers/mod.rs
```diff
@@ -1,6 +1,6 @@
-use std::iter::Enumerate;
-use std::iter::Peekable;
+use std::iter::{Enumerate, Peekable};
 
+#[allow(missing_debug_implementations)]
```
djc, line 3: I don't consider the `askama_shared` interface public even though it technically is, so I don't think following the advice about applying `Debug` ~everywhere is useful here.

askama_escape/src/lib.rs
```diff
@@ -65,12 +77,13 @@ where
...
 }
 
+#[derive(Debug)]
 pub struct EscapeWriter<'a, E, W> {
     fmt: W,
     escaper: &'a E,
 }
 
-impl<'a, E, W> Write for EscapeWriter<'a, E, W>
+impl<E, W> Write for EscapeWriter<'_, E, W>
```
djc, line 86: I think these changes to elicit lifetimes are useful.

askama_shared/src/input.rs
```diff
@@ -211,12 +212,13 @@ fn extension(path: &Path) -> Option<&str> {
...
 }
 
+#[derive(Debug)]
 pub enum Source {
     Path(String),
     Source(String),
 }
 
-#[derive(PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq)]
```
djc, line 221: I don't see a good reason to derive these here.

djc (commented):
Okay, I guess the exception is `unreachable_pub`, which I find very useful.

djc:
(And thanks for working on this, I do appreciate the time put into keeping the code quality up to par!)

askama_actix/src/lib.rs
Kijewski:
```rs
HttpResponse::from_error(Box::new(err) as Box<dyn std::error::Error + 'static>)
```
This line gives a "trivial_casts" warning, most likely a false positive, because `as Box<…>` cannot be omitted or  replaced by `as _`.

But actually WrappedError is a better implementation, because it removes the need for `dyn` and `Box`.

djc: Well, let's discuss that change in a separate PR.

Kijewski:
It's a lot of commits, now. :)

askama_shared/src/heritage.rs
```diff
@@ -4,6 +4,7 @@ use std::path::{Path, PathBuf};
 use crate::parser::{Expr, Loop, Macro, Node};
 use crate::{CompileError, Config};
 
+#[derive(Debug)]
```
djc, line 7: I don't consider `askama_shared` as a truly public API, so let's skip these.

askama_actix/src/lib.rs
djc: Let's rename it to `ActixError` and add a comment that this is just a newtype wrapper that implements `actix_web::ResponseError`?

djc (commented):
This is much better, thanks! Some minor nits left.

askama_shared/src/heritage.rs
Kijewski: Removed.

djc (approved):

PR #594, opened 2022-01-05
Determine Content-Type during compilation

### LINKED ISSUE #592: Why is there a runtime dependency for mime-guess?
I guess the MIME type could be "guessed" at compile time, and be stored in a field Template::MIMETYPE. Or am I overlooking something?

Kijewski:
Closes #592.

### DIFF
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -31,12 +31,14 @@ with-rocket = ["askama_derive/rocket"]
 with-tide = ["askama_derive/tide"]
 with-warp = ["askama_derive/warp"]
 
+# deprecated
+mime = []
+mime_guess = []
+
 [dependencies]
 askama_derive = { version = "0.11.0", path = "../askama_derive" }
 askama_escape = { version = "0.10", path = "../askama_escape" }
 askama_shared = { version = "0.12.0", path = "../askama_shared", default-features = false }
-mime = { version = "0.3", optional = true }
-mime_guess = { version = "2", optional = true }
 
 [package.metadata.docs.rs]
 features = ["config", "humansize", "num-traits", "serde-json", "serde-yaml"]
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -128,33 +128,11 @@ pub use crate::shared::helpers;
 pub use crate::shared::{read_config_file, Error, MarkupDisplay, Result};
 pub use askama_derive::*;
 
+#[deprecated(since = "0.11.1", note = "The only function in this mod is deprecated")]
 pub mod mime {
     #[cfg(all(feature = "mime_guess", feature = "mime"))]
-    pub fn extension_to_mime_type(ext: &str) -> mime_guess::Mime {
-        let basic_type = mime_guess::from_ext(ext).first_or_octet_stream();
-        for (simple, utf_8) in &TEXT_TYPES {
-            if &basic_type == simple {
-                return utf_8.clone();
-            }
-        }
-        basic_type
-    }
-
-    #[cfg(all(feature = "mime_guess", feature = "mime"))]
-    const TEXT_TYPES: [(mime_guess::Mime, mime_guess::Mime); 6] = [
-        (mime::TEXT_PLAIN, mime::TEXT_PLAIN_UTF_8),
-        (mime::TEXT_HTML, mime::TEXT_HTML_UTF_8),
-        (mime::TEXT_CSS, mime::TEXT_CSS_UTF_8),
-        (mime::TEXT_CSV, mime::TEXT_CSV_UTF_8),
-        (
-            mime::TEXT_TAB_SEPARATED_VALUES,
-            mime::TEXT_TAB_SEPARATED_VALUES_UTF_8,
-        ),
-        (
-            mime::APPLICATION_JAVASCRIPT,
-            mime::APPLICATION_JAVASCRIPT_UTF_8,
-        ),
-    ];
+    #[deprecated(since = "0.11.1", note = "Use Template::MIME_TYPE instead")]
+    pub use crate::shared::extension_to_mime_type;
 }
 
 /// Old build script helper to rebuild crates if contained templates have changed
--- a/askama/src/lib.rs
+++ b/askama/src/lib.rs
@@ -86,6 +86,9 @@ pub trait Template {
 
     /// Provides a conservative estimate of the expanded length of the rendered template
     const SIZE_HINT: usize;
+
+    /// The MIME type (Content-Type) of the data that gets rendered by this Template
+    const MIME_TYPE: &'static str;
 }
 
 /// Object-safe wrapper trait around [`Template`] implementers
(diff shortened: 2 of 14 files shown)


askama/Cargo.toml
```diff
@@ -36,8 +36,6 @@ with-warp = ["askama_derive/warp"]
 askama_derive = { version = "0.11.0", path = "../askama_derive" }
 askama_escape = { version = "0.10", path = "../askama_escape" }
 askama_shared = { version = "0.12.0", path = "../askama_shared", default-features = false }
-mime = { version = "0.3", optional = true }
```
djc, line 39: I think we should release this in a semver-compatible way, so this stuff should stay.

askama/src/lib.rs
```diff
@@ -91,6 +91,9 @@ pub trait Template {
 
     /// Provides a conservative estimate of the expanded length of the rendered template
     const SIZE_HINT: usize;
+
+    /// The MIME type (Content-Type) of the data that gets rendered by this Template
+    const MIME_TYPE: &'static str;
```
djc, line 96: Should we deprecate `EXTENSION` and `extension()`? I feel like this is probably better.

askama/src/lib.rs
```diff
@@ -126,42 +132,17 @@ impl<T: Template> DynTemplate for T {
...
+        Self::MIME_TYPE
+    }
 }
 
 pub use crate::shared::filters;
 pub use crate::shared::helpers;
 pub use crate::shared::{read_config_file, Error, MarkupDisplay, Result};
 pub use askama_derive::*;
 
-pub mod mime {
```
djc, line 136: Should stay for semver compat.

askama_actix/Cargo.toml
```diff
@@ -14,7 +14,7 @@ edition = "2018"
 
 [dependencies]
 actix-web = { version = "4.0.0-beta.19", default-features = false }
-askama = { version = "0.11.0", path = "../askama", default-features = false, features = ["with-actix-web", "mime", "mime_guess"] }
+askama = { version = "0.11.0", path = "../askama", default-features = false, features = ["with-actix-web"] }
```
djc, line 17: So I guess semver-compat means we should keep it here, as well.

askama_actix/src/lib.rs
```diff
@@ -1,29 +1,42 @@
...
+use actix_web::{HttpResponse, HttpResponseBuilder, ResponseError};
 pub use askama::*;
 
-use actix_web::HttpResponse;
-
 pub trait TemplateToResponse {
     fn to_response(&self) -> HttpResponse<BoxBody>;
 }
 
+struct WrappedError(Error);
```
djc, line 15: This is unrelated and should at least be in a separate commit.

askama_shared/src/generator.rs
```diff
@@ -160,13 +160,18 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
         buf.writeln("}")?;
 
-        buf.writeln("const EXTENSION: ::std::option::Option<&'static ::std::primitive::str> = ")?;
-        buf.writeln(&format!("{:?}", self.input.extension()))?;
-        buf.writeln(";")?;
-
-        buf.writeln("const SIZE_HINT: ::std::primitive::usize = ")?;
-        buf.writeln(&format!("{}", size_hint))?;
-        buf.writeln(";")?;
+        buf.writeln(&format!(
```
djc, line 163: Would prefer not changing style in the same commit as making functional changes.

askama_gotham/src/lib.rs
```diff
@@ -1,24 +1,27 @@
...
 
 pub use askama::*;
-
 pub use gotham::handler::IntoResponse;
 pub use gotham::state::State;
-pub use hyper::{Body, Response, StatusCode};
+use hyper::header::{HeaderValue, CONTENT_TYPE};
+pub use hyper::{Body, Response};
+
+pub trait TemplateToResponse {
```
djc, line 9: This change is unrelated and should be in a separate commit (or probably in a separate PR), same for the other integrations.

askama_shared/src/input.rs
```diff
@@ -13,6 +14,7 @@ pub struct TemplateInput<'a> {
     pub print: Print,
     pub escaper: &'a str,
     pub ext: Option<String>,
+    pub mime_type: String,
```
djc, line 17: Let's keep this out for now and rely on guessing based on the extension first?

djc (commented):
I think the core idea here is great, looks like you brought a bunch of other changes along that I think need a separate discussion.

askama/src/lib.rs
vallentin: But the `#[cfg(all(feature = "mime_guess", feature = "mime"))]` could probably be moved to the `mod` instead of each item

djc: It's public, so we need to keep it  around for semver-compat reasons. However, we can deprecate all these!

vallentin: Makes sense. Still kinda funny that empty module is available if the feature is disabled, haha

Kijewski:
I refactored the PR.

The last commit "Remove ext argument in integrations" is probably a breaking change according to semver. I'm not sure if the functions in the integration crates (e.g.  askama_axum::into_response) can be considered to be part of the public interface, they have no documentation after all.

askama_shared/src/input.rs
```diff
@@ -175,6 +176,35 @@ impl TemplateInput<'_> {
...
+                    (mime::TEXT_CSV, mime::TEXT_CSV_UTF_8),
+                    (
+                        mime::TEXT_TAB_SEPARATED_VALUES,
+                        mime::TEXT_TAB_SEPARATED_VALUES_UTF_8,
+                    ),
+                    (
+                        mime::APPLICATION_JAVASCRIPT,
+                        mime::APPLICATION_JAVASCRIPT_UTF_8,
+                    ),
+                ];
```
djc, line 197: I'd prefer to outline the const here, just dump it near the bottom of the module. Also, could there be a separate commit that moves the logic from `askama::mime` into askama_shared, forwarding the methods from the old place such that we don't duplicate the logic, and also a separate commit that changes the logic to what you have here?

djc (commented):
Sorry for the somewhat iterative review, but got some more changes I'd like to the PR.

Also I think we should ditch the last commit that removes the `_ext` arguments in the integrations to preserve compatibility, even if it's undocumented API.

askama_shared/src/input.rs
Kijewski: Done

Kijewski:
> Sorry for the somewhat iterative review, but got some more changes I'd like to the PR.

No problem!

> Also I think we should ditch the last commit that removes the `_ext` arguments in the integrations to preserve compatibility, even if it's undocumented API.

I omitted the diff to remove the "ext" parameters. Maybe we can cherry-pick https://github.com/djc/askama/pull/594/commits/d796abdfbb35b9100d2cbced452a4e38109f310a before the next big release?

djc:
> I omitted the diff to remove the "ext" parameters. Maybe we can cherry-pick [d796abd](https://github.com/djc/askama/commit/d796abdfbb35b9100d2cbced452a4e38109f310a) before the next big release?

Sounds good. Or even file it as a draft PR including something about compatibility in the title?

djc (approved):

djc:
Awesome work, thanks!

PR #595, opened 2022-01-06
Remove the iron integration from generator

Kijewski:
Issue #527 removed the askama_iron package, but not the integration if
someone uses askama_derive's feature with "iron".

The old askama_iron crate uses askama v0.10, so it will still work.


djc (approved):
Nice! I noticed this recently as well but apparently forgot to act on it.

PR #597, opened 2022-01-06
Optimize parsing of ranges

Kijewski:
Right now almost every expression needs to be parsed twice: `expr_any()`
first parses the left-hand side of a range expression, and if no `..`
or `..=` was found the left-hand expression is parsed again, this time
as the result of the function.

This diff removes the second parsing step by first looking for
`.. (opt rhs)`, then for `lhs .. (opt rhs)`.


Kijewski:
Extracted from #590.

djc (approved):

Issue #581, opened 2021-12-15
global config


yatesco:
Hi - I imagine like most people, I don't care about whitespace being preserved and all of my templates should be escaped as if they are HTML.

Is there any way of saying:
 - "discard all whitespace" within a template without adding the "-" to each bracket
 - to askama itself "discard all whitespace for all templates`

And I assume the file extension `html` of the template is sufficient for it choose the right encoding?

Thanks!

vallentin:
Currently, it's not possible to configure that. Though, personally I would like that feature as well

djc:
I'm open to having a `preserve-whitespace` setting in `askama.toml` which defaults to true, if anyone wants to hack that in. I wonder if we should also have a `+` to then be able to locally opt in again.

vallentin:
I'll hack at it later or during the weekend. Since I need this feature anyways, and always end up piping my static pages through html-minifier.

But yes, we should definitely have something like `+`. Since there's cases, where removing all whitespaces, messes with HTML pages, e.g. with horizontal menus. In those cases, only 1 single space is needed. So should we be able to differentiate between `+` being "keep all spaces" vs "if any spaces, then keep 1 single space".

Overall for my use cases, I'd prefer the latter behavior, but the former behavior ofc makes more sense for a general `+`

yatesco:
> I'll hack at it later or during the weekend. Since I need this feature anyways, and always end up piping my static pages through html-minifier.
> 
> But yes, we should definitely have something like `+`. Since there's cases, where removing all whitespaces, messes with HTML pages, e.g. with horizontal menus. In those cases, only 1 single space is needed. So should we be able to differentiate between `+` being "keep all spaces" vs "if any spaces, then keep 1 single space".
> 
> Overall for my use cases, I'd prefer the latter behavior, but the former behavior ofc makes more sense for a general `+`

Excellent. I'm itching to have a go but I just don't have time :-(

vallentin:
I'm always pressed for time as well. But if you'd like a go, then I can wait. Just say whenever you think you have time, and keep us posted if anything changes or you need assistance or suggestions :)

yatesco:
Thanks @vallentin - unfortunately I won't have time. I am thinking of moving towards a client side library (e.g. Sauron) instead, but I will watch this with interest and may return (I have to generate some email templates further down the line)

vallentin:
Alright, fair enough :)

(I'll continue with my initial plan then, to hack at it later or during the weekend)

PR #598, opened 2022-01-06
Add whitespace stripping in the input

Kijewski:
In HTML sources indentation can aid readability, but it only bloats the
output. This PR adds the argument "strip" to templates, which lets you
automatically strip the template input:

* "none": No stripping of the input. The generator still strips spaces
  at the back of the input.
* ~~"tail": The current default, i.e. pop a single newline character at
  the end.~~
* ~~"trim": remove all spaces at the front and back of a line, and remove
  empty lines.~~
* "eager": like "trim", but replace multiple whitespaces with a single
  space.

The automatic stripping does not try to understand the input at all. It
does not know what `xml:space="preserve"`, `<pre>` or `<textarea>`
means.

### DIFF
--- a/askama_shared/src/lib.rs
+++ b/askama_shared/src/lib.rs
@@ -13,6 +13,7 @@ use std::{env, fmt, fs};
 use serde::Deserialize;
 
 pub use crate::input::extension_to_mime_type;
+pub use crate::strip::Strip;
 pub use askama_escape::MarkupDisplay;
 
 mod error;
@@ -27,6 +28,7 @@ pub mod heritage;
 pub mod input;
 #[doc(hidden)]
 pub mod parser;
+mod strip;
 
 #[derive(Debug)]
 pub struct Config<'a> {
@@ -34,6 +36,7 @@ pub struct Config<'a> {
     pub syntaxes: BTreeMap<String, Syntax<'a>>,
     pub default_syntax: &'a str,
     pub escapers: Vec<(HashSet<String>, String)>,
+    pub strip: Strip,
 }
 
 impl Config<'_> {
@@ -102,6 +105,7 @@ impl Config<'_> {
             syntaxes,
             default_syntax,
             escapers,
+            strip: raw.strip.unwrap_or_default(),
         })
     }
 
@@ -200,6 +204,7 @@ struct RawConfig<'d> {
     general: Option<General<'d>>,
     syntax: Option<Vec<RawSyntax<'d>>>,
     escaper: Option<Vec<RawEscaper<'d>>>,
+    strip: Option<Strip>,
 }
 
 impl RawConfig<'_> {
@@ -257,19 +262,17 @@ where
 }
 
 #[allow(clippy::match_wild_err_arm)]
-pub fn get_template_source(tpl_path: &Path) -> std::result::Result<String, CompileError> {
+pub fn get_template_source(
+    tpl_path: &Path,
+    strip: Strip,
+) -> std::result::Result<String, CompileError> {
     match fs::read_to_string(tpl_path) {
         Err(_) => Err(format!(
             "unable to open template file '{}'",
             tpl_path.to_str().unwrap()
         )
         .into()),
-        Ok(mut source) => {
-            if source.ends_with('\n') {
-                let _ = source.pop();
-            }
-            Ok(source)
-        }
+        Ok(source) => Ok(strip.apply(source)),
     }
 }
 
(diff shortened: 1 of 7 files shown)


vallentin:
I "don't" like the idea of `trim`. Because whitespace is or can be significant in HTML/CSS, i.e. with inline menus :)

Also, can this be used for #581? Because I looked a bit at that before Christmas, but then I got busy with Christmas and New Year's

Kijewski:
It's like trim() is called on every line, then empty lines are dropped, and for the eager case `s/\s+/ /g` applied.

Of course, white spaces are significant in `<pre>` and `<textarea>`. But how often do you see `<pre>` nowadays? When do you use a `<textarea>` that is populated by static data, not by `{{data}}`? For these two cases you cannot apply the automatic stripping.

askama_shared/src/lib.rs
```diff
@@ -257,19 +258,71 @@ where
...
 }
 
+/// Whitespace handling of the input source.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Strip {
+    /// Don't strip any spaces in the input.
+    None,
+    /// Remove a single single newline at the end of the input. This is the default.
+    Tail,
+    /// Remove all whitespaces at the front and back all lines, and remove empty lines.
```
djc, line 268: Missing `of`? Please remove periods at the end of the first line of docstrings.

askama_shared/src/lib.rs
```diff
@@ -257,19 +258,71 @@ where
...
+    Trim,
```
djc, line 269: Maybe call this `TrimLines` to avoid confusion as with @vallentin?

askama_shared/src/lib.rs
```diff
@@ -257,19 +258,71 @@ where
...
+
+impl Default for Strip {
+    fn default() -> Self {
+        Strip::Tail
+    }
+}
+
+impl Strip {
+    pub fn apply<'a, S: Into<Cow<'a, str>>>(self, src: S) -> String {
+        let src = S::into(src);
```
djc, line 282: I think I'd prefer spelling this `src.into()`.

djc (commented):
Awesome that you're working on this, did you guys already see https://github.com/rust-lang/rust/pull/92526? It's pretty amazing! This also came up there. I'll link to this PR in from there so that we can feedback from them.

I guess this also needs some good documentation in the book.

My main point of feedback is that I think we'd want a way to change this setting on a crate-wide basis, in the `askama.toml` configuration. There, we might make it part of an extension/type-based configuration (I think we already have some?) so that people can easily apply this to HTML but not plain text templates. I was going to suggest that we ditch any kind of per-template configuration but it just occurs to me that that might make it harder to test this well (end-to-end) the way you've done here.

camelid:
Am I correct in thinking that this is all done at compile-time? If so, that makes this extra-exciting :)

djc:
> Am I correct in thinking that this is all done at compile-time? If so, that makes this extra-exciting :)

Yes, that would all be done at compile time.

askama_shared/src/lib.rs
Kijewski: Fixed.

askama_shared/src/lib.rs
Kijewski: I renamed it into TrimLines / "trim-lines".

askama_shared/src/lib.rs
Kijewski: You're right. I don't know why I used the more complicated syntax before. :)

Kijewski:
I moved Strip from askama_shared's root into askama_shared::input. That's were the other arguments life.

djc:
Did you have any thoughts about the configuration approach?

Kijewski:
> Did you have any thoughts about the configuration approach?

Yes, that should be a project-wide configuration. I added the optional strip parameter to the config, too.

But I actually have not idea how to write a test for it. Can you give me some pointers?

Kijewski:
I'm not convinced anymore that this is the best implementation.

Most likely it would be simpler and better to use Strip::apply() in Generator::write_buf_writable() instead of preformatting the input.

Kijewski:
I simplified the PR. No only "none" and "eager" are allowed.

Kijewski:
My "simplified" version does not work. The "complicated" version https://github.com/djc/askama/commit/766d049536c17a089fd7031e639d9f4999c9e64c did. Reverted and rebased.

Kijewski:
Superseded by #673.

PR #599, opened 2022-01-11
Basic skeleton for a multilinual actix-askama site

gpmake:
I created a basic multi-language site with two pages (Home and About). For each page, I used two template structs with the same content. A possible next step is to create a newtype wrapped around an inner type with a Deref impl to share the same type definitions for the template structs. Here is the repository create on my GitHub account: git@github.com:gpmake/basic-actix-askama-multilingual-site.git

### DIFF
--- a/examples/basic-actix-askama-multilingual-site/Cargo.toml
+++ b/examples/basic-actix-askama-multilingual-site/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "basic-actix-askama-multilingual-page"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+actix-web = "4.0.0-beta.19"
+askama = "0.11.0"
+tokio = "1"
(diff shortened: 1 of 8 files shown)


Kijewski:
Please have a look at #600 if that kind of localization would work for you. It has the added feature that you don't need to copy the structure multiple times, which I guess works better if you want to support _a lot_ of languages.

djc:
So I don't think we actually want an example of this, but here's the pattern I was trying to explain:

```rust
#[derive(Template)]
#[template(path = "home/home-en.html")]
struct HomeEn<'a>(Home<'a>);

impl<'a> std::ops::Deref for HomeEn<'a> {
    type Target = Home<'a>;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

struct Home<'a> {
    lang: &'a str,
    title: &'a str,
    page: &'a str,
}
```

This way you can have multiple templates defined that all use the same underlying context type. You could potentially also have an `impl<'a> From<Home<'a>> for HomeEn<'a>` if that makes life easier for you.

djc:
Going to close this for now; let's discuss in #600.

PR #600, opened 2022-01-12
WIP: Multiple templates per struct

Kijewski:


djc:
Have you looked at #434? I think that's probably a better approach.

Kijewski:
I don't think it's an either-or, both approaches could be used together.

I called my implementation "l10n", but it is also usable e.g. if you have a different templates for desktop and mobile users:

```rs
#[template(path = "desktop/index.html", localizer = "is_mobile")]
#[l10n("true", "mobile/index.html")]
struct Index { … }
```

Or to present an "eat our cookies!" template

```rs
#[template(path = "desktop/index.html", localizer = "cookie")]
#[l10n("None", "accept_cookie_form.html")]
struct Index { … }
```

gpmake:
I refactored my test code with WIP: I18n #600.

`{https://github.com/gpmake/basic-actix-askama-multilingual-site.git, branch = "pr-multi-paths}`

It works well and I find the proposed solution very clean. 

I don't quite understand why you need to add the index.html file. What is the purpose?
In my code I had to insert these (empty) files otherwise it would not compile.

``` rust
#[derive(Template)]
#[template(path = "localization/index.html", localizer = "language")]
#[l10n(r#""de""#, "localization/index.de.html")]
#[l10n(r#""en""#, "localization/index.en.html")]
#[l10n(r#""es""#, "localization/index.es.html")]
#[l10n(r#""fr""#, "localization/index.fr.html")]
struct LetDestructoringTuple<'a> {
    language: &'a str,
    user: &'a str,
}
```

Kijewski:
> In my code I had to insert these (empty) files otherwise it would not compile.

The basic case is meant as a fallback. Probably I should make this optional if the other cases are exhaustive.

djc:
I don't think we should call this i18n, it's more like `multi`-template support or something (my use case would be to implement templates for plaintext and HTML email). So can you paste some generated code here that shows wat this ends up generating?

Kijewski:
```rs
impl<'a> ::askama::Template for LetDestructoringTuple<'a> {
        include_bytes!("…/askama/testing/templates/localization/index.es.html");
        include_bytes!("…/askama/testing/templates/localization/index.fr.html");
        include_bytes!("…/askama/testing/templates/localization/index.de.html");
        include_bytes!("…/askama/testing/templates/localization/index.html");
        include_bytes!("…/askama/testing/templates/localization/index.en.html");
        include_bytes!("…/askama/testing/templates/localization/base.html");
        match &self.language {
            &"de" => {
                write!(
                    writer,
                    "Hallo, {expr0}!",
                    expr0 = &::askama::MarkupDisplay::new_unsafe(&(self.user), ::askama::Html),
                )?;
            }
            &"en" => {
                write!(
                    writer,
                    "Hello, {expr1}!",
                    expr1 = &::askama::MarkupDisplay::new_unsafe(&(self.user), ::askama::Html),
                )?;
            }
            &"es" => {
                write!(
                    writer,
                    "¡Hola, {expr2}!",
                    expr2 = &::askama::MarkupDisplay::new_unsafe(&(self.user), ::askama::Html),
                )?;
            }
            &"fr" => {
                write!(
                    writer,
                    "Localization test:\nBonjour, {expr3} !",
                    expr3 = &::askama::MarkupDisplay::new_unsafe(&(self.user), ::askama::Html),
                )?;
            }
            _ => {
                write!(
                    writer,
                    "Not implemented: {expr4}",
                    expr4 = &::askama::MarkupDisplay::new_unsafe(&(self.language), ::askama::Html),
                )?;
            }
        }
        Ok(())
    }
    const EXTENSION: ::std::option::Option<&'static ::std::primitive::str> = Some("html");
    const SIZE_HINT: ::std::primitive::usize = 17;
    const MIME_TYPE: &'static ::std::primitive::str = "text/html; charset=utf-8";
}
impl<'a> ::std::fmt::Display for LetDestructoringTuple<'a> {
    #[inline]
    }
}
```

djc:
Okay, that seems pretty neat! I'll need to review this in more detail...

gpmake:
It is a neat improvement for my case, too! The PR title "Multiple templates per struct" also seems fitting fine.

djc:
So one thing that this approach does not solve is having templates with different MIME types. I feel that is a fairly important use case here, and we should not commit to this approach if we think we might need another approach in order to have templates with different MIME types (but the same context type).

Kijewski:
Hm, maybe something like:

    fn get_alternative_index(&self) -> usize {
        …the generated match block…
    }

    /// The template's extension, if provided
    const EXTENSION: &'static [Option<&'static str>];

    /// Provides a conservative estimate of the expanded length of the rendered template
    const SIZE_HINT: &'static [usize];

    /// The MIME type (Content-Type) of the data that gets rendered by this Template
    const MIME_TYPE: &'static [&'static str];

?

djc:
I think it ends up being a crappy fit for what the `Template` trait is supposed to mean. I think maybe this should be a different trait, like `MultiTemplate` or `TemplateGroup`?

Kijewski:
Closed in favor of https://github.com/djc/askama/issues/666#issuecomment-1104825932. Unless someone wants to adopt this PR, but I don't  think that it's much more useful than EnumTemplate.

PR #602, opened 2022-01-12
Replace `&PathBuf` with `&Path`

Kijewski:
PathBuf is to String like Path is to str, so `&PathBuf` is a pointer to a pointer. Clippy does not likes that.

In #600 clippy did notice the problem after some refactoring. This change is at least a little useful on its own.

### DIFF
--- a/askama_derive/src/lib.rs
+++ b/askama_derive/src/lib.rs
@@ -46,7 +46,7 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
 
     let mut parsed = HashMap::new();
     for (path, src) in &sources {
-        parsed.insert(path, parse(src, input.syntax)?);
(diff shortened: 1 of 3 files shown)


vallentin (approved):
Looks good 👍 

(But don't merge until @djc says it looks good too)

djc:
The `as_path()` calls seem unidiomatic to me. In order of preference, which of `&path`, `*path` and `path.as_ref()` works?

djc:
Ugh, that's still pretty ugly. I'll see if I can play with this tomorrow.

Kijewski:
I'd actually say I made it worse. :D

djc:
Yeah, okay, would you mind going back to using `as_ref()` instead of all the double dereferencing?

djc (approved):

PR #603, opened 2022-01-12
Make sure `#[template(…)]` is given exactly once

Kijewski:


askama_shared/src/input.rs
```diff
@@ -28,23 +28,25 @@ impl TemplateInput<'_> {
...
+            if attr.path.is_ident("template") {
+                if template_args.is_some() {
+                    return Err(CompileError::Static("duplicated 'template' attribute"));
+                }
+                let template = attr.parse_meta().map_err(|e| {
+                    CompileError::String(format!("unable to parse attribute: {}", e))
+                })?;
+                match template {
+                    syn::Meta::List(inner) => template_args = Some(inner),
+                    _ => return Err("attribute 'template' has incorrect type".into()),
```
vallentin, line 44: The rests of the code uses `CompileError::Static` explicitly. For consistency, should we do that here as well?

vallentin (commented):
Other than that, nice!

askama_shared/src/input.rs
Kijewski: Yes, it does look more consistent this way. The only bad think is that the line is now longer than 100 characters, and "cargo fmt" even added a block around the return statement, turning this one line into five. :)

vallentin: Ah... Damn :)

PR #606, opened 2022-01-13
Tweak attribute parsing some more

djc:
I think this a little easier to follow than the result from #603, what do you think?

### DIFF
--- a/askama_shared/src/input.rs
+++ b/askama_shared/src/input.rs
@@ -36,16 +36,13 @@ impl TemplateInput<'_> {
                 if template_args.is_some() {
                     return Err(CompileError::Static("duplicated 'template' attribute"));
                 }
-                let template = attr.parse_meta().map_err(|e| {
-                    CompileError::String(format!("unable to parse attribute: {}", e))
-                })?;
-                match template {
-                    syn::Meta::List(inner) => template_args = Some(inner),
-                    _ => {
-                        return Err(CompileError::Static(
-                            "attribute 'template' has incorrect type",
-                        ));
+
+                match attr.parse_meta() {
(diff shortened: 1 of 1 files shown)


askama_shared/src/input.rs
```diff
@@ -36,16 +36,15 @@ impl TemplateInput<'_> {
...
-                        return Err(CompileError::Static(
-                            "attribute 'template' has incorrect type",
-                        ));
+
+                match attr.parse_meta() {
+                    Ok(syn::Meta::List(syn::MetaList { nested, .. })) => {
+                        template_args = Some(nested);
+                    }
+                    Ok(_) => {
+                        return Err("'template' attribute must be a list".into());
```
vallentin, line 45:
```suggestion
                    Ok(_) => return Err("'template' attribute must be a list".into()),
```

To match the following one. Unless it was intentional? :)

Kijewski:
I like it! It's easier to follow than my variant.

askama_shared/src/input.rs
Kijewski: :+1:

askama_shared/src/input.rs
```diff
@@ -36,16 +36,14 @@ impl TemplateInput<'_> {
...
-                        return Err(CompileError::Static(
-                            "attribute 'template' has incorrect type",
-                        ));
+
+                match attr.parse_meta() {
+                    Ok(syn::Meta::List(syn::MetaList { nested, .. })) => {
+                        template_args = Some(nested);
+                    }
+                    Ok(_) => return Err("'template' attribute must be a list".into()),
                     }
```
vallentin, line 45: This one needs to be removed as well. But couldn't mark it, while making the initial suggestion

Kijewski (approved):

Issue #607, opened 2022-01-14
Request: Content check blocks


jaw-sh:
Some templating engines support a streamlined "if content" block.

For instance, in XenForo's templating engine, it looks like this.
![image](https://user-images.githubusercontent.com/9262991/149522538-e1e96afc-d31f-478a-a705-29286c57444b.png)

Take for instance the following scenario which is very common in a template.

```
<ul class="moderator-actions">
    ​{% if client.can_update_post(post) %}<li><a href="/posts/{{ post.id }}/edit">Edit</a></li>{% endif %}
    ​{% if client.can_delete_post(post) %}<li><a href="/posts/{{ post.id }}/delete">Delete</a></li>{% endif %}
</ul>
```

We can have many buttons in this list which should only display to the client if they have the ability to use them. The more buttons in this list, the more checks we are using. However, the list itself should only render at all if at least a single item in the list is available.

To achieve this effect, we'd need to add a check like this:

```
{% if client.can_update_post(post) || if client.can_delete_post(post)  %}
<ul class="moderator-actions">
    ​{% if client.can_update_post(post) %}<li><a href="/posts/{{ post.id }}/edit">Edit</a></li>{% endif %}
    ​{% if client.can_delete_post(post) %}<li><a href="/posts/{{ post.id }}/delete">Delete</a></li>{% endif %}
</ul>
{% endif %}
```

This is undesirable for two reasons. First, if this list had 10 options, we would need 10 checks, and that is very clumsy to write out.  Second, these checks add computational complexity, so needing to run them all twice just for a cosmetic effect is a very poor fix.

A content check could look like this.

```
{% contentcheck  %}
<ul class="moderator-actions">
    {% content %}
    ​{% if client.can_update_post(post) %}<li><a href="/posts/{{ post.id }}/edit">Edit</a></li>{% endif %}
    ​{% if client.can_delete_post(post) %}<li><a href="/posts/{{ post.id }}/delete">Delete</a></li>{% endif %}
    {% endcontent %}
</ul>
{% endcontentcheck %}
```

If only whitespace exists within a `{% content %}` block, then the entire `{% contentcheck %}` will not render.

This should be nestable. Example: A list may have two categories of options with a header. The category+header does not render unless it has at least one option. The whole list will not render unless at least one category is visible.

djc:
What Rust code would you expect your template example to compile to?

Kijewski:
[...]

PR #608, opened 2022-01-14
Parse tuple expressions

Kijewski:
Askama understands how to destructure tuples in let and match
statements, but it does not understand how to build a tuple.

This PR fixes this shortcoming.


Kijewski:
I think this would resolve #607 quite nicely. It would compile the example to ↓, which is reasonably similar to what you would write by hand:
```rs
match &(
    self.client.can_post(&(self.post)),
    self.client.can_update(&(self.post)),
) {
    (false, false) => {
        writer.write_str("No!")?;
    }
    (can_post, can_update) => {
        writer.write_str("<ul>")?;
        if *(&(can_post) as &bool) {
            writer.write_str("<li>post</li>")?;
        }
        if *(&(can_update) as &bool) {
            writer.write_str("<li>update</li>")?;
        }
        writer.write_str("</ul>")?;
    }
}
```

djc (approved):
LGTM, thanks!

Kijewski:
You're welcome :)

PR #610, opened 2022-01-15
Fix json/yaml features

### LINKED ISSUE #609: serde-json feature not working with cargo resolver 2
Using askama with the `serde-json` feature does not work with the new resolver.

I think this is because askama passes the json feature to askama_shared, but not to askama_derive. Thus the askama_shared of askama_derive does not have the json feature.

**Cargo.toml**

`askama = { version = "0.11.0", features = ["serde-json"] }`

**Error**:

`the json filter requires the serde-json feature to be enabled.`

jannik4:
This should fix #609. For now as a draft, since I am unsure about the overall dependency tree. Maybe there are other places that need to be addressed, too.

(Just tested in my project: It works for me now.)


djc (approved):
Makes sense, thanks!

Issue #398, opened 2020-12-09
Feature Request: Idiomatic way to mimic Option::map


msrd0:
This is very straight-forward in rust code, where we can simply write

```rust
// ver is of type Option<&str>
let suffix = ver.map(|ver| format!("-{}", var)).unwrap_or_default();
```

In askama, however, this code is much longer and less readable:

```jinja2
{% let suffix -%}
{% match ver -%}
  {% when Some with (ver) -%}
    {% let suffix = format!("-{}", ver) -%}
  {% when None -%}
    {% let suffix = String::new() -%}
{% endmatch -%}
```

It would be nice if askama had a better way to mimic `Option::map`, either supporting rust closures, or introducing a custom syntax for this.

vallentin:
If the feature request is only about closures, then this is a duplicate of #284. As djc already mentioned in that issue, the problem is that adding closure support essentially requires adding support for all valid Rust syntax.

If it's a reusability problem, then you could create a [macro] or implement it as a function in Rust.

[macro]: https://djc.github.io/askama/template_syntax.html#macros

```rust
fn suffix(ver: &str) -> String {
    format!("-{}", ver)
}
```

Then within your template you can do close to what you wanted originally.

```jinja
{% let suffix = ver.map(self::suffix).unwrap_or_default() %}
{{ suffix }}

{# or just #}

{{ ver.map(self::suffix).unwrap_or_default() }}
```

msrd0:
I think this is neither - I don't want above code anywhere in my template because my first impression would be _what is this?_ and then, after understanding the code, _why is that written so complicated?_. Neither am I asking askama to support closures. I just want some syntax that makes it easier to deal with `Option`s, without specifying how that's exactly going to look - personally, I'm a fan of the Kotlin-esque `ver.let { "-$it" } ?: ""` syntax, but that probably looks alien to people that are just used to Rust syntax.

What you suggested seems to me more like a workaround than a fix.

djc:
So far I've tried to avoid inventing too much syntax that isn't either (a) already in Rust or (b) already in Jinja et al and maps naturally to Rust code. If we invent syntax here, that becomes something that every user of Askama potentially has to learn, so I feel that the onus is on me to be careful about inventing things that fit in well with the rest of the design.

Another option here that hasn't been mentioned is a [custom filter](https://djc.github.io/askama/filters.html#custom-filters).

vallentin:
[...]

PR #611, opened 2022-01-18
Support single-expression closures in templates

### LINKED ISSUE #284: Support closures in templates
Currently working with `Option`s can be rather frustrating and clumsy because of the lack of support for closures inside templates. Add support for closures would significantly improve the library's usability.

jannik4:
Fixes #284 (and #398?)

## Example

```rust
use askama::Template;

#[derive(Debug, Clone)]
struct User { name: String }

impl User {
    fn ferris() -> Self { Self { name: "Ferris".to_string() } }
}

#[derive(Template)]
#[template(
    source = r#"Hello {{ user_opt.map(|user| user.name.as_str()).unwrap_or("World") }}"#,
    ext = "txt"
)]
struct ClosureTemplate<'a> {
    user_opt: Option<&'a User>,
}

#[test]
fn test_closure() {
    let user = User::ferris();
    let t = ClosureTemplate { user_opt: Some(&user) };
    assert_eq!(t.render().unwrap(), "Hello Ferris");

    let t = ClosureTemplate { user_opt: None };
    assert_eq!(t.render().unwrap(), "Hello World");
}
```

## Drawbacks

I think the only inconvenience of the current implementation is that `Copy` types can cause problems in combination with the automatic borrowing, see below. This should always be fixable by a clone, in this case `data.flag.clone()`. (Edit: Or by using `copied` on the Option as mentioned by @vallentin)

The alternative would be to not automatically borrow the closure expression. But then non `Copy` types would not work anymore.

```rust
struct Data {
    flag: bool,
}

#[derive(Template)]
#[template(
    source = r#"{{ data_opt.map(|data| data.flag).unwrap_or(false) }}"#,
    ext = "txt"
)]
struct DataTemplate {
    data_opt: Option<Data>,
}

// error[E0308]: mismatched types
//   --> testing\tests\closures.rs:85:10
//    |
// 85 | #[derive(Template)]
//    |          ^^^^^^^^ expected `&bool`, found `bool`
//    |
```

### DIFF
--- a/testing/tests/closures.rs
+++ b/testing/tests/closures.rs
@@ -1,7 +1,5 @@
 use askama::Template;
 
-const FALSE: &bool = &false;
-
 #[derive(Debug, Clone)]
 struct User {
     name: String,
@@ -59,7 +57,7 @@ fn test_closure_shadow() {
 
 #[derive(Template)]
 #[template(
-    source = r#"{{ user_opt.map(|user| user.flag).unwrap_or(FALSE) }}"#,
+    source = r#"{{ user_opt.map(|user| user.flag).copied().unwrap_or(false) }}"#,
     ext = "txt"
 )]
 struct ClosureBorrowTemplate<'a> {
--- a/testing/tests/closures.rs
+++ b/testing/tests/closures.rs
@@ -1,0 +1,79 @@
+use askama::Template;
+
+const FALSE: &'static bool = &false;
+
+#[derive(Debug, Clone)]
+struct User {
+    name: String,
+    flag: bool,
+}
+
+impl User {
+    fn ferris() -> Self {
+        Self {
+            name: "Ferris".to_string(),
+            flag: true,
+        }
+    }
+}
+
+#[derive(Template)]
+#[template(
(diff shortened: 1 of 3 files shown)


testing/tests/closures.rs
```diff
@@ -0,0 +1,79 @@
...
+    let t = ClosureShadowTemplate { user: Some(&user) };
+    assert_eq!(t.render().unwrap(), "Hello Ferris");
+
+    let t = ClosureShadowTemplate { user: None };
+    assert_eq!(t.render().unwrap(), "Hello World");
+}
+
+#[derive(Template)]
+#[template(
+    source = r#"{{ user_opt.map(|user| user.flag).unwrap_or(FALSE) }}"#,
```
vallentin, line 62: I understand why `FALSE` is needed, but it's still possible to do the following, right?

```suggestion
    source = r#"{{ user_opt.map(|user| user.flag).copied().unwrap_or(false) }}"#,
```

jannik4: Yes, that should work too. I used `FALSE` to statically ensure that there really is a reference. But since `copied` only works on references, that's probably better here.

Kijewski:
Please have a look at jannik4/askama#1. You don't need to add explicit lifetimes in all these places.

PR #612, opened 2022-01-24
Use Spans to report better error locations

Kijewski:
This PR might look huge and incomprehensible, but it actually shouldn't be too difficult to read. I hope.

Currently if the parsing fails, an error in `#[derive(Template)]` is reported, even though the library knows a more narrow location, e.g. at least that the error is actually in `#[template(…)]`, or in the `path = "…"` argument if the file is absent.

This PR is also done in anticipation of rust-lang/rfcs#3200, which might make it possible to report an error location in the actual template file; and #600, which adds multiple files as possible template file source of an error.


askama_shared/src/input.rs
```diff
@@ -59,76 +83,115 @@ impl TemplateInput<'_> {
...
+                        meta.span(),
+                        "unsupported attribute argument",
+                    ));
                 }
             };
 
-            if pair.path.is_ident("path") {
+            let ident = match pair.path.get_ident() {
+                Some(ident) => ident,
+                None => continue,
```
vallentin, line 97: Unsure if intentional, but the existing could would also trigger the "unsupported attribute key" error in this case

Kijewski: A proc_macro is [supposed to ignore](https://doc.rust-lang.org/reference/procedural-macros.html#derive-macro-helper-attributes) "derive macro helper attributes" it didn't define. The reason is that a derive macro is only ever additive, so you see attribute of other proc_macros, too. Maybe you want to build a template using clap::Parser, then Askama should not prevent you from doing so.

So IMO the old implementation was a bug, but fixing it was not my primary intention. :)

vallentin: Ahh makes sense. Just wanted to be sure :)

Kijewski: Sorry, I was wrong! I was thinking of line 36: `let ident = match attr.path.get_ident() {`

In here syn's parse_meta() already ensured that `pair.path.get_ident()` cannot be none, so I made this case unreachable!().

askama_shared/src/lib.rs
```diff
@@ -293,33 +292,6 @@ static DEFAULT_ESCAPERS: &[(&[&str], &str)] = &[
     (&["j2", "jinja", "jinja2"], "::askama::Html"),
 ];
 
-#[derive(Debug)]
```
djc, line 296: I don't like this. I think we should have a custom error type. We could redefine it as `CompileError(Cow<'static, str>)` and I agree that that would be an improvement. I would probably do it in two steps: use the `From` impls in favor of literal `Static` and `String` variants, then change the inner type. How about we take that as a separate PR for now to help shrink this one?

Issue #180, opened 2018-12-12
Print deprecation notices when using _parent field


djc:
I want to remove the special handling for `_parent` fields, but it would be nice to have one version where these still work but generate some sort of warning from the compiler. It appears there's no nice way to print deprecation notices from procedural macros, so I suppose we should just print a note to `stderr` that maybe is formatted a little bit like a compiler warning.

botika:
We can use [librustc-errors](https://github.com/rust-lang/rust/blob/master/src/librustc_errors/lib.rs#L664). 
```rust
if parent.is_some() {
   emit(...);
}
```

How about?

djc:
Sounds great! Is that crate available from crates.io?

botika:
It is not available in crate. I have made a small version.

PR #613, opened 2022-01-25
Remove support for deprecated `_parent` field

Kijewski:
The support for the magic `_parent` field is deprecated since v0.8.0
or issue #180. It's bothersome to keep this feature alive, when no-one
should be using it for 3 years.


djc:
Sounds good, but this should probably wait until we do another semver-incompatible release.

DamnWidget:
You should also probably refactor all the test cases where `_parent` is being used

michaelmelanson:
What should we be using instead?

djc:
@michaelmelanson what's your use case, and what have you tried that doesn't work?

Kijewski:
Instead of removing the inheritance test entirely, I made it use ops::Deref.

djc (approved):
Nice!

PR #614, opened 2022-01-25
Unify handling of calls

Kijewski:
Instead of having `Expr::VarCall`, `Expr::PathCall` and
`Expr::MethodCall`, this PR unifies the handling of calls by removing
the former three variants, and introducing `Expr::Call`.

This makes parsing postfix operators easier.

Extracted from #590.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -603,7 +603,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             // If `iter` is a call then we assume it's something that returns
             // an iterator. If not then the user can explicitly add the needed
             // call without issues.
-            Expr::MethodCall(..) | Expr::PathCall(..) | Expr::Index(..) => {
+            Expr::Call(..) | Expr::Index(..) => {
                 buf.writeln(&format!("let _iter = ({}).into_iter();", expr_code))
             }
             // If accessing `self` then it most likely needs to be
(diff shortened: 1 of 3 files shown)


askama_shared/src/generator.rs
```diff
@@ -1329,16 +1319,27 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                     _ => return Err("loop.cycle(…) expects exactly one argument".into()),
                 },
                 s => return Err(format!("unknown loop method: {:?}", s).into()),
+            },
+            left => {
+                match *left {
```
vallentin, line 1324: This seems like something that could easily be forgotten, to be updated, when changing the parser. Could it maybe be "simplified" or handled in the parser, by having some `enum CallKind`, which is a third value of `Expr::Call()`? :)

Kijewski: Please see my comment in #590: https://github.com/djc/askama/pull/590#discussion_r779486870. I think the list of special cases is exhaustive. I think not even the `Expr::Unary(..) | Expr::BinOp(..) | Expr::Range(..)` case is actually need, because the parser won't ever generate an `Expr::Call(Expr::Unary(..), ..)`, only `Call(Group(Unary(..)), ..)`.

djc (commented):
This is looking pretty good. One open question about the parser setup.

askama_shared/src/parser.rs
```diff
@@ -510,61 +498,42 @@ fn expr_single(i: &str) -> IResult<&str, Expr<'_>> {
...
-    for (aname, args) in attrs {
-        res = if let Some(args) = args {
-            Expr::MethodCall(Box::new(res), aname, args)
-        } else {
-            Expr::Attr(Box::new(res), aname)
-        };
-    }
-
-    Ok((i, res))
+fn expr_attr<'a>(i: &'a str) -> IResult<&'a str, Box<dyn 'a + FnOnce(Expr<'a>) -> Expr<'a>>> {
```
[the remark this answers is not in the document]
Kijewski: We don't. I guess I wanted to make it hyper-extensible for no proper reason. :D

djc (approved):
Ooh, nice improvement!

PR #615, opened 2022-01-31
Make is_shadowing_variable() failable

Kijewski:
Extracted from #612.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -804,26 +804,38 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         buf.writeln(";")
     }
 
-    fn is_shadowing_variable(&self, var: &Target<'a>) -> bool {
+    fn is_shadowing_variable(&self, var: &Target<'a>) -> Result<bool, CompileError> {
         match var {
             Target::Name(name) => {
                 let name = normalize_identifier(name);
                 match self.locals.get(&name) {
                     // declares a new variable
-                    None => false,
+                    None => Ok(false),
                     // an initialized variable gets shadowed
-                    Some(meta) if meta.initialized => true,
+                    Some(meta) if meta.initialized => Ok(true),
                     // initializes a variable that was introduced in a LetDecl before
-                    _ => false,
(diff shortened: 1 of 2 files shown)


askama_shared/src/generator.rs
```diff
@@ -804,26 +804,34 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
                 }
             }
             Target::Tuple(_, targets) => targets
                 .iter()
-                .any(|target| self.is_shadowing_variable(target)),
+                .find_map(|target| match self.is_shadowing_variable(target) {
+                    Ok(false) => None,
+                    outcome => Some(outcome),
+                })
+                .unwrap_or(Ok(false)),
```
djc, line 826: Transforming from `Ok(false)` to `None`, then back to `Ok(false)` doesn't seem like a great way to do this. At this point it might be better to just use a `for`-loop?

askama_shared/src/generator.rs
```diff
@@ -804,26 +804,34 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
             Target::Struct(_, named_targets) => named_targets
                 .iter()
-                .any(|(_, target)| self.is_shadowing_variable(target)),
-            _ => panic!("Cannot have literals on the left-hand-side of an assignment."),
+                .find_map(|(_, target)| match self.is_shadowing_variable(target) {
+                    Ok(false) => None,
+                    outcome => Some(outcome),
+                })
+                .unwrap_or(Ok(false)),
+            _ => Err("Cannot have literals on the left-hand-side of an assignment.".into()),
```
djc, line 834: Probably rewrite this as "literals are not allowed on the left-hand side of an assignment" (no capital, no period).

askama_shared/src/generator.rs
Kijewski: Yeah, a loop looks better. Changed.

askama_shared/src/generator.rs
Kijewski: Changed.

djc (approved):
Nice!

PR #617, opened 2022-01-31
Remove `panic!()` in `loop.cycle([])`

Kijewski:


askama_shared/src/generator.rs
```diff
@@ -1316,7 +1316,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                 "cycle" => match args {
                     [arg] => {
                         if matches!(arg, Expr::Array(arr) if arr.is_empty()) {
-                            panic!("loop.cycle(…) cannot use an empty array.");
+                            return Err("loop.cycle(…) cannot use an empty array.".into());
```
djc, line 1319: This should also not end in a period (I do realize it was there already).

djc (dismissed):

askama_shared/src/generator.rs
Kijewski: I removed the period.

Kijewski:
Fixed one more typo in my change.

djc (approved):

PR #620, opened 2022-01-31
Use exact trybuild version

Kijewski:
Sometimes for no obvious reason an old version is selected and the
output is different in just about every ui test. Just pin it to the
currently newest version and test if an updated version still works when
a new version gets released.

### DIFF
--- a/testing/Cargo.toml
+++ b/testing/Cargo.toml
@@ -15,7 +15,7 @@ serde_json = { version = "1.0", optional = true }
 
 [dev-dependencies]
 criterion = "0.3"
-trybuild = "1.0"
+trybuild = "1.0.55"
 version_check = "0.9"
 
 [[bench]]


testing/Cargo.toml
```diff
@@ -15,7 +15,7 @@ serde_json = { version = "1.0", optional = true }
 
 [dev-dependencies]
 criterion = "0.3"
-trybuild = "1.0"
+trybuild = "1.0.55"
```
djc, line 18: This is not pinning it. If you want to actually pin it it should be prefixed with =.

Kijewski: Does "1.0.55" mean "1.0.x" with "x >= 55" then? I guess this also OK, but it shouldn't be some ancient version.

djc: It actually means 1.x with x > 0.55.

djc (approved):

PR #622, opened 2022-02-01
Add markdown filter

Kijewski:


Kijewski:
"expected struct `std::time::Duration`, found `u64`" seems unrelated. `actix-test` was updated 2h ago. I guess you should not assume a 0.1-beta version to be stable. :)

vallentin:
The dependabot pr #623 also failed, so probably a change to actix-test, yeah :)

vallentin (dismissed):
This is really nice. I actually have had this filter locally for ages. Don't know why I never thought of adding it. But nice! 🚀

askama_shared/src/filters/markdown.rs
```diff
@@ -0,0 +1,17 @@
+use std::borrow::Borrow;
```
djc, line 1: I'd prefer to put this in an inline module in `filters/mod.rs`, it's a bit too short to merit its own file.

askama_shared/src/filters/markdown.rs
```diff
@@ -0,0 +1,17 @@
...
+use crate::error::Result;
+
+pub fn markdown<'a, E, S>(e: E, s: S) -> Result<MarkupDisplay<E, String>>
+where
+    E: Escaper,
+    S: Borrow<&'a str>,
+{
+    let mut buf = String::new();
+    push_html(&mut buf, Parser::new_ext(s.borrow(), Options::all()));
+    Ok(MarkupDisplay::new_safe(buf, e))
```
djc, line 16: Hmm, how do you reason about safe here? I seem to remember Markdown can include arbitrary HTML.

vallentin: The `Parser` is an `Iterator`, so technically we could escape all [`Event::Html`](https://docs.rs/pulldown-cmark/latest/pulldown_cmark/enum.Event.html#variant.Html). But then we should definitely have an `markdown_unsafe` filter variant

djc: Couldn't we just have `|markdown|safe`?

vallentin: But if by default `|markdown` is escaped. Then it also escapes all the HTML generated by `push_html`, right?
E.g. even just `# Title` which results in `<h1>Title</h1>` would then be escaped, or am I missing something?

Kijewski:
I replaced pulldown-cmark by comrak. Unsafe HTML is now sanitized by default, but the user may opt-out to allow unsafe HTML.

djc (commented):
Seems reasonable. Would you mind squashing the second commit?

askama_shared/src/filters/mod.rs
```diff
@@ -75,6 +75,8 @@ pub const BUILT_IN_FILTERS: [&str; 27] = [
     "wordcount",
     "json", // Optional feature; reserve the name anyway
     "yaml", // Optional feature; reserve the name anyway
+    #[cfg(feature = "markdown")]
```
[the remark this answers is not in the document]
Kijewski: If someone already has a custom markdown filter already, this PR would break their code.

askama_shared/src/filters/mod.rs
```diff
@@ -379,6 +381,61 @@ pub fn wordcount<T: fmt::Display>(s: T) -> Result<usize> {
     Ok(s.split_whitespace().count())
 }
 
+#[cfg(feature = "markdown")]
+pub fn markdown<'a, E: Escaper, S: std::borrow::Borrow<&'a str>>(
+    e: E,
+    s: S,
+    options: Option<&comrak::ComrakOptions>,
```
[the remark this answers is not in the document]
Kijewski: Options, yes. E.g. to allow [unsafe HTML](https://github.com/djc/askama/pull/622/commits/cbd2be745afbf42349e05c1d7f171ebd0cd76605#diff-032aef9600122904f8186dff77678b0eb8b440239d0adafb2f1062acb5f5468fR48-R59), or enable „smart quotes“.

Plugins, maybe. I think you can add e.g. Geshi-like syntax highlighting with plugins, which I don't use but I can see users wanting to use this feature.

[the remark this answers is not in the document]
djc: Let's ditch the plugin support for now, we can add it back once someone asks for it.

askama_shared/src/filters/mod.rs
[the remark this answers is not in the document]
djc: That's not really actually different from when the `json` and `yaml` filters were introduced, though... Also in the common case it probably wouldn't break their code since our implementation does the same thing anyway?

askama_shared/src/filters/mod.rs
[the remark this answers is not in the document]
Kijewski: Removed.

Kijewski:
I restructured the PR a bit, and I think it's quite pretty now. :)

testing/tests/optional/markdown.rs
```diff
@@ -0,0 +1,73 @@
+use askama::Template;
```
djc, line 1: We can just have `#![cfg()]` in here instead of the additional module?

Kijewski: Then everything needs to be annotated or everything needs to go into a ↓ block.

```rs
#[cfg(…)]
const _: () = {
    …
};
```

This would work for me, too, but I like the "optional" module better, actually. I would rather remove the `#[cfg()]`s from the other places in the testing module, and move the tests into "optional".

What do you think? Is this "over engineered"?

djc: I think you missed the `!` -- it's an inner attribute, that applies to the module as a whole.

Kijewski: Ahhhhh, yes, I did, because I simply assumed that disabling a whole file cannot be possible. Will change.

djc (approved):

PR #635, opened 2022-02-08
Untangle integrations from askama_shared

Kijewski:
This makes it easier to update integration crates on another schedule than the rest of the Askama crates.

askama_mendes is not yet converted, because I did not look into the changes needed to make mandes latest version work with the integration.

### DIFF
--- a/askama_tide/src/lib.rs
+++ b/askama_tide/src/lib.rs
@@ -30,3 +30,31 @@ pub fn into_response<T: Template>(t: &T, ext: &str) -> Response {
         }
     }
 }
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! impl_template {
+    (($ident:ident) ($($impl_generics:tt)*) ($($orig_generics:tt)*) ($($where_clause:tt)*)) => {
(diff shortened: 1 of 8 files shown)


djc:
Hmm, not sure I'm a fan. Might there be a way to achieve the same benefits with some traits?

Kijewski:
> Might there be a way to achieve the same benefits with some traits?

I don't think so, but I'm happy if you can show me otherwise, too. E.g. askama_tide  implements two traits.

askama_tide/src/lib.rs
```diff
@@ -30,3 +30,31 @@ pub fn into_response<T: Template>(t: &T, ext: &str) -> Response {
         }
     }
 }
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! impl_template {
+    (($ident:ident) ($($impl_generics:tt)*) ($($orig_generics:tt)*) ($($where_clause:tt)*)) => {
```
Kijewski, line 37: Hm, this line alone might prove as a conclusive counterpoint to my idea. :-/

PR #639, opened 2022-02-16
Port JSON safety changes to 0.10.x branch

djc:
I created a 0.10.x branch so we can get this out to folks still using 0.10.x.

@Kijewski can you review if I've merged this correctly? There were a bunch of conflicts.


djc:
@Kijewski do you understand why this test failure happens? Seems surprising.

djc:
Proactively merged this, since it mostly consists of commits we had already reviewed on main.

PR #640, opened 2022-02-16
Bump version for askama_derive

djc:


askama/Cargo.toml
```diff
@@ -37,9 +37,9 @@ mime = []
 mime_guess = []
 
 [dependencies]
-askama_derive = { version = "0.11.0", path = "../askama_derive" }
+askama_derive = { version = "0.11.2", path = "../askama_derive" }
 askama_escape = { version = "0.10", path = "../askama_escape" }
```
Kijewski, line 41: Explicitly use 0.10.3?

Kijewski:
Darn, I forgot how interwoven everything is, otherwise I had bumped the versions for all four crates in my PR.

askama/Cargo.toml
djc: Forgot, and I suppose it will pull that in anyway through askama_shared.

The bigger problem actually is that the bump for Askama 0.11.1 is  technically semver-incompatible because we added a const to a trait. While we usually think of our trait implementations as output, it's possible that someone  has actually written a manual implementation which will now break. I figured I'd take the risk, since the alternatives seem worse.

Kijewski (approved):

PR #641, opened 2022-02-19
Book: document markdown filter

Kijewski:


book/src/filters.md
```diff
@@ -323,7 +370,49 @@ Ugly: <script>var data = "{{data|json}}";</script>
 Ugly: <script>var data = '{{data|json|safe}}';</script>
 ```
 
+### `markdown`
+[#markdown]: #markdown
+
+Enabling the `markdown` feature fill enable the use of the `markdown` filter.
+This will render a value using a [Github flavored commonmark](https://docs.rs/comrak/0.12.*/comrak/) syntax.
```
vallentin, line 377:
```suggestion
This will render a value using a [GitHub flavored CommonMark](https://docs.rs/comrak/0.12.*/comrak/) syntax.
```

book/src/filters.md
```diff
@@ -323,7 +370,49 @@ Ugly: <script>var data = "{{data|json}}";</script>
...
+        unsafe_: true,
+        ..Default::default()
+    },
+    ..Default::default()
+};
+
+…
+
+{{ "<script>alert('Hello world')</script>"|markdown(options) }}
+```
```
vallentin, line 406: Maybe the usage example, should be more copy-pastable. I think it might be confusing, that it's half Rust and half Jinja :)

book/src/filters.md
```diff
@@ -323,7 +370,49 @@ Ugly: <script>var data = "{{data|json}}";</script>
 Ugly: <script>var data = '{{data|json|safe}}';</script>
 ```
 
+### `markdown`
+[#markdown]: #markdown
+
+Enabling the `markdown` feature fill enable the use of the `markdown` filter.
```
djc, line 376: fill -> will?

book/src/filters.md
Kijewski: :+1: I added a link to the unit tests. I think this should be the most readable variant, esp. since it should not be too interesting to most users.

djc (approved):

Issue #629, opened 2022-02-04
Latest release of askama_actix incompatible with askama (askama_shared 0.12.1 needed)


max-m:
Example code:
============
```rs
use actix_web::{web, App, HttpRequest, HttpServer, Responder};
use askama::Template;

#[derive(Template)]
#[template(path = "hello.html")]
struct HelloTemplate {
    name: String,
}

async fn greet(req: HttpRequest) -> impl Responder {
    let name = req.match_info()
        .get("name")
        .unwrap_or("World")
        .to_owned();

    HelloTemplate {
        name
    }
}

#[actix_web::main]
async fn main() -> std::io::Result<()> {
    HttpServer::new(|| {
        App::new()
            .route("/", web::get().to(greet))
            .route("/{name}", web::get().to(greet))
    })
    .bind(("127.0.0.1", 8080))?
    .run()
    .await
}
```
(this is a simple combination of the minimal Actix and Askama examples)

`cargo.toml` using the latest releases that work in combination:
================================================
```toml
[package]
name = "askama-test"
version = "0.1.0"
edition = "2021"

[dependencies]
actix-web = "3.3.3"
askama = "0.10.5"
askama_actix = "0.11.1"
```
Upgrading either dependency breaks:
=============================

Upgrading `askama` to `0.11` results in `actix_web::Responder` not being implemented on the template struct.
```
error: src/main.rs:10: the trait bound `HelloTemplate: actix_web::Responder` is not satisfied
help: src/main.rs:10: in this macro invocation
error: src/main.rs:10: the trait `actix_web::Responder` is not implemented for `HelloTemplate`
error: could not compile `askama-test` due to 2 previous errors
```

Upgrading `askama_actix` as well (`0.12`) results in `BoxBody` missing.
```
error: src/main.rs:4: type `Body` is not a member of trait `actix_web::Responder`
error: src/main.rs:4: not a member of trait `actix_web::Responder`
error: src/main.rs:4: cannot find type `BoxBody` in module `actix_web::body`
error: src/main.rs:4: not found in `actix_web::body`
error: src/main.rs:4: associated type `Body` not found for `Self`
error: src/main.rs:4: associated type `Body` not found
error: could not compile `askama-test` due to 4 previous errors
```

When pulling in the as of yet unreleased `askama_shared` 0.12.1 from commit 3b701f2 via the following configuration the test project builds and runs again.
```
[...]

PR #647, opened 2022-03-10
Move all of the template creation into askama_shared

Kijewski:
Before this PR the handling of integrations was done both by askama_shared and askama_derive. Let askama_shared do the work. This will prevent problems like https://github.com/djc/askama/issues/629, when both packages might come out of sync.

This PR removes the back-and-forth interaction between the two crates. Now askama_derive is a single re-export of `#[derive(Template)]` which has to be done in a proc_macro crate.

This most likely means that askama_derive is "final", unless another derive template needs to be introduced in the future.


askama_shared/src/derive.rs
```diff
@@ -0,0 +1,107 @@
+extern crate proc_macro;
```
djc, line 1: So I tried this in the past and this doesn't work. As far as I remember, you should not import `proc_macro` anywhere except in a `proc-macro = true` crate. However, what we can do is convert to proc_macro2 types in askama_derive and then pass that into askama_shared.

I don't love the `derive` name of this module, given that I think we might want an attribute macro at some point in the not too distant future, too. Can we use `macro` as the name, or is that a keyword?

djc (commented):
This is looking great, thanks for taking on a more systematic fix for this class of issues!

askama_shared/src/derive.rs
Kijewski: I replaced proc_macro by proc_macro2.

Actually, I guess I could move everything into generator.rs. It's only 100 lines as is.

djc: > Actually, I guess I could move everything into generator.rs. It's only 100 lines as is.

That probably makes sense for now, although it just means we should start splitting that up since it's already pretty big. Good start, though!

askama_shared/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "askama_shared"
-version = "0.12.2"
+version = "0.13.0"
```
djc, line 3: Why do we need this bump?

askama_derive/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "askama_derive"
-version = "0.11.2"
+version = "0.12.0"
```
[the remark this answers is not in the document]
Kijewski: I un-exported a bunch of symbols. Older `askama_shared` versions would break with this update. Every item in https://github.com/djc/askama/blob/main/askama_derive/src/lib.rs#L8-L10 is gone after the PR.

askama_shared/Cargo.toml
Kijewski: I un-exported a bunch of symbols. Older `askama_shared` versions would break with this update. Every item in https://github.com/djc/askama/blob/main/askama_derive/src/lib.rs#L8-L10 is gone after the PR.

askama_derive/Cargo.toml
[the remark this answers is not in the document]
Kijewski: I removed the features section, because they are now handles in askama_shared.

I am not 100% sure, but I guess if I re-added the features, incrementing the patch version only should be fine.

djc (approved):

PR #654, opened 2022-03-28
Expose the fact that templates implement Display

Kijewski:
Rebased on #653.

This PR consists of two parts:

*  Expose the fact that templates implement Display

   This is a quite useful feature, because you can use templates in format!(), format_args!(), etc.
*  Add io::writer helper methods

   It might not be immediately obvious to everyone how easy it is to use Askama template with std::io (e.g. files) instead of std::fmt, so this PR adds a few helper methods to make this more obvious to novice users.

(Nb. the "novice user" is me in this case :) )

### DIFF
--- a/askama_shared/src/lib.rs
+++ b/askama_shared/src/lib.rs
@@ -29,7 +29,7 @@ mod parser;
 /// Main `Template` trait; implementations are generally derived
 ///
 /// If you need an object-safe template, use [`DynTemplate`].
-pub trait Template {
+pub trait Template: fmt::Display {
     /// Helper method which allocates a new `String` and renders into it
     fn render(&self) -> Result<String> {
         let mut buf = String::with_capacity(Self::SIZE_HINT);
@@ -89,6 +89,12 @@ impl<T: Template> DynTemplate for T {
 
     fn mime_type(&self) -> &'static str {
         Self::MIME_TYPE
+    }
+}
+
+impl fmt::Display for dyn DynTemplate {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.dyn_render_into(f).map_err(|_| ::std::fmt::Error {})
     }
 }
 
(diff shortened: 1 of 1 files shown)


askama_shared/src/lib.rs
```diff
@@ -37,9 +37,21 @@ pub trait Template: fmt::Display {
         Ok(buf)
     }
 
-    /// Renders the template to the given `writer` buffer
+    /// Helper method which allocates a new `Vec<u8>` and renders into it
+    #[inline]
+    fn render_bytes(&self) -> Result<Vec<u8>> {
```
djc, line 42: This doesn't meet my cost/benefit bar, conversions with `String::into_bytes()` should be easy.

askama_shared/src/lib.rs
```diff
@@ -37,9 +37,21 @@ pub trait Template: fmt::Display {
...
+        Ok(self.render()?.into_bytes())
+    }
+
+    /// Renders the template to the given `writer` fmt buffer
     fn render_into(&self, writer: &mut (impl std::fmt::Write + ?Sized)) -> Result<()>;
 
+    /// Renders the template to the given `writer` io buffer
+    #[inline]
+    fn render_into_io(&self, writer: &mut (impl std::io::Write + ?Sized)) -> std::io::Result<()> {
```
djc, line 51: This one seems reasonable, though the name seems a little unwieldy. Maybe `write_into()`? That's technically somewhat ambitious between `impl std::fmt::Write` and `impl std::io::Write` but that doesn't seem like a big deal.

askama_shared/src/lib.rs
Kijewski: Removed.

askama_shared/src/lib.rs
Kijewski: Renamed.

djc (approved):

PR #659, opened 2022-04-04
Fixed Resolver issue that broke how libraries got features.

### LINKED ISSUE #656: Markdown filter requires markdown feature error, even though the feature is enabled
Hey, I'm not sure if I'm doing something wrong here or it's a bug, but I can't seem to get the `markdown` filter working, even though the relevant feature is enabled; I keep getting a proc macro compile error. Any help would be greatly appreciated!

My code looks something like this:

> `src/main.rs`

```rust
use askama::Template;

#[derive(Template)]
#[template(path = "basic.jinja")]
struct BasicTemplate {
    content: String,
}

fn main() {}
```

> `templates/basic.jinja`

```hbs
{{ content|markdown }}
```

> `Cargo.toml`

```toml
[package]
name = "demo"
version = "0.1.0"
edition = "2021"

[dependencies]
askama = { version = "0.11.1", features = ["markdown"] }
```

### LINKED ISSUE #658: Template is unable to make Axum into_response
I have everything set from the current git.

```
[dependencies]
axum = "0.5"
axum-macros = "0.2"
askama = { version = "0.11", git = "https://github.com/djc/askama.git", features = [
    "serde-json",
    "with-axum",
] }
askama_axum = { version = "0.1", git = "https://github.com/djc/askama.git" }
```

and This is my file thats having issues.
```
use crate::forms::User;
use crate::{
    extensions::{Flashes, FullState},
    forms::UserOnlineDisplay,
};
use askama::Template;
use axum::{routing::get, Router};
use axum_macros::debug_handler;
use axum_sessions_auth::Authentication;
use serde::*;

#[derive(Template, Debug, Deserialize, Serialize)]
#[template(path = "main/main.html")]
struct MainTemplate {
    #[serde(default)]
    current_user: User,
    #[serde(default)]
    online: Vec<UserOnlineDisplay>,
    #[serde(default)]
    flashes: Flashes,
}

pub fn main_routes() -> Router {
    Router::new().route("/main", get(main_page))
}

#[debug_handler]
async fn main_page(state: FullState) -> MainTemplate {
    let online_user = state.state.users_online().await;

    MainTemplate {
        current_user: state.auth.current_user.unwrap_or_default(),
        online: online_user,
        flashes: state.flashes,
    }
}
```

the macro output is 

https://pastebin.com/dsmSqB5t

tried with both stable and nightly fully updated.

It is causing the error 

```
the trait bound `site::main::MainTemplate: axum::response::IntoResponse` is not satisfied
the trait `axum::response::IntoResponse` is not implemented for `site::main::MainTemplate`
```

which is further also causing

```
the trait bound `fn(extensions::state::FullState) -> impl futures::Future<Output = site::main::MainTemplate> {site::main::main_page}: axum::handler::Handler<_, _>` is not satisfied\nthe trait `axum::handler::Handler<_, _>` is not implemented for `fn(extensions::state::FullState) -> impl futures::Future<Output = site::main::MainTemplate> {site::main::main_page}`
'''

even if I remove the part erroring out and it compiles without error it still does not build the into_response.
thank you for looking into the issue. Makes me wounder if the cfg is not working...

### PR DESCRIPTION
**Pull request by genusistimelord**
So basically the major issue here is Resolver 2 builds different versions of the same lib for different things. So for example Askama uses askama-derive to build the process macros, But Askama is linking to askama-shared with features set. However askama-derive does not have these features set and links to askama-shared as well.  askama-derive will then get its own version with the features disabled while askama will get its own version with them enabled. 

In the process of updating and fixing this I migrated the edition to 2021 for everything and updated everything to use the latest 0.11 version of askama. I then moved and added it so both askama-derive and askama build and get the same library. This has fixed the issue in my test code and should also fix other process macro derive issues as well. notably should fixes #656 and fixes #658 .

This issue is mostly caused by people using the newer resolver in projects which is on by default with newer Rust versions.

My test at https://github.com/genusistimelord/AskamaTest should also fully compile and work now due to this fix. Also id suggest upping the askama-axum version to 0.2 so people don't try to use it with axum 0.4.8 as it is not backwards compatible update.

### CODE REVIEW

File: askama/Cargo.toml [hunk 1]
```diff
@@ -24,13 +24,13 @@ urlencode = ["askama_shared/percent-encoding"]
...
-with-rocket = ["askama_shared/rocket"]
-with-tide = ["askama_shared/tide"]
-with-warp = ["askama_shared/warp"]
+with-actix-web = ["askama_derive/with-actix-web"]
+with-axum = ["askama_derive/with-axum"]
+with-gotham = ["askama_derive/with-gotham"]
+with-mendes = ["askama_derive/with-mendes"]
+with-rocket = ["askama_derive/with-rocket"]
+with-tide = ["askama_derive/with-tide"]
+with-warp = ["askama_derive/with-warp"]
```
jplatte: It makes sense to me that these feature dependencies are needed, same goes for the ones in askama_derive. However, all of the other changes of this PR seem like a distraction at best, and I don't really get why you reverted all of my changes from #657.

.gitignore
```diff
@@ -1,3 +0,0 @@
-target
-Cargo.lock
-.DS_Store
```
jplatte, line 3: By the way, you deleted this file which was probably not intentional 😉

genusistimelord: added it back and yeah somehow it disappeared XD.

Kijewski: Um, the file is still missing from the last commit. :)

genusistimelord:
OK let me go about reducing the changes then. I will PR this again with what is necessary.

PR #660, opened 2022-04-04
Moved Features into derive, Updated Askama versions to 0.11

genusistimelord:
So basically the major issue here is Resolver 2 builds different versions of the same lib for different things. So for example Askama uses askama-derive to build the process macros, But Askama is linking to askama-shared with features set. However askama-derive does not have these features set and links to askama-shared as well. askama-derive will then get its own version with the features disabled while askama will get its own version with them enabled.

In the process of updating and fixing this I  updated everything to use the latest 0.11 version of askama. I then moved and added it so both askama-derive and askama build and get the same library. This has fixed the issue in my test code and should also fix other process macro derive issues as well. notably should fixes https://github.com/djc/askama/issues/656 and fixes https://github.com/djc/askama/issues/658 .

This issue is mostly caused by people using the newer resolver in projects which is on by default with newer Rust versions.

My test at https://github.com/genusistimelord/AskamaTest should also fully compile and work now due to this fix. Also id suggest upping the askama-axum version to 0.2 so people don't try to use it with axum 0.4.8 as it is not backwards compatible update.

Redid to fix too many commits and changes.


Kijewski:
Thank you for working on fixing my errors! 

Just for the record: I do not like the new resolver. :)

genusistimelord:
> Thank you for working on fixing my errors!
> 
> Just for the record: I do not like the new resolver. :)

it in itself is both Good and bad. Good as it does help prevent some issues by doing what it does by also bad because it breaks old ways people use to use. XD. It is not your fault though more of they should have better laid out what the new one does and that everyone will be using it by default now.  This is why i though switching to Edition 2021 would make better sense as then it is also enabled by default for the libs but that is ok.

dnaka91:
Looking forward to this being merged. Currently, have the same issue, as I tried to update to `axum` 0.5 and using the latest `askama` from git.

askama_actix/Cargo.toml
```diff
@@ -14,7 +14,7 @@ edition = "2018"
 
 [dependencies]
 actix-web = { version = "4", default-features = false }
-askama = { version = "0.11.2", path = "../askama", default-features = false, features = ["with-actix-web"] }
+askama = { version = "0.11", path = "../askama", default-features = false, features = ["with-actix-web"] }
```
jplatte, line 17: Why did you change this?

genusistimelord: because they should all be matching the latest update as those are for bug fixes rather than API additions or changes. unless they are not following any standard for incrementing the version? otherwise when he releases an update to askama that fixes a bug that did not change the api then he does not need to republish all of the smaller libraries pointing to askama and they all see the latest updates for regular bug fixes.

Kijewski (changes requested):

askama_actix/Cargo.toml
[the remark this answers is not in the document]
genusistimelord: generally i thought due to the new resolver that it would auto pull in the latest patch version of askama. but i guess i should add this back just in case they might be using the old resolver.

djc: @genusistimelord 0.11.2 is a minimum bound, the ^ semver operator is implicit in Cargo.toml. So this already supports anything 0.11.x with x >= 2.

Kijewski:
Huh, I did not intend to close the PR?? I only wanted to unpick the last commit and now I cannot push anymore, because the PR is closed. :-/

PR #664, opened 2022-04-07
Add suppress_whitespace config option and add `config` parameter to the derive macro

GuillaumeGomez:
I recently disovered in jinja (with Flask in python) that we could enable `automatic_block_trim`, allowing us to remove all the `-` and also preventing us to forget them. I saw it was missing in this crate so I thought it'd be a good addition.

So now, about the newly added `config` parameter in the derive macro: I realized when writing the tests that I couldn't have a specific configuration for a given test, making it quite complex. So the easiest solution was to create this new parameter. If you prefer, I can open a separate PR for it (its implementation is in the second commit). Of course, it's only if both features make sense to you.

Finally: I didn't find out where to put the documentation for `config`. Do you have a suggestion?

In any case: thanks a lot for writing this crate! It's really nice to use.


askama_shared/src/generator.rs
```diff
@@ -129,6 +253,11 @@ struct Generator<'a, S: std::hash::BuildHasher> {
     buf_writable: Vec<Writable<'a>>,
     // Counter for write! hash named arguments
     named: usize,
+    // If set to true, the meaning of the minus sign (`-`) will be reverted on all blocks
```
vallentin, line 256:
```suggestion
    // If set to true, the meaning of the minus sign (`-`) will be inverted on all blocks
```

Right? :)

testing/tests/gen_ws_tests.py
```diff
@@ -40,7 +40,7 @@ def pairwise(iterable):
     return zip(a, b)
 
 
-def write_cond(conds, active_branch):
+def write_cond(conds, active_branch, revert):
```
vallentin, line 43:
```suggestion
def write_cond(conds, active_branch, invert):
```

testing/tests/gen_ws_tests.py
```diff
@@ -182,5 +186,6 @@ def write_match_tests(f):
     with open("ws.rs", "w") as f:
         f.write("// This file is auto generated by gen_ws_tests.py\n\n")
         f.write("use askama::Template;\n")
-        write_cond_tests(f)
+        write_cond_tests(f, "", "")
+        write_cond_tests(f, "_revert", ', config = "test_trim.toml"')
```
vallentin, line 190:
```suggestion
        write_cond_tests(f, "_inverted", ', config = "test_trim.toml"')
```

askama_shared/src/generator.rs
GuillaumeGomez: Absolutely! :D

GuillaumeGomez:
Fixed the typos. :)

askama_shared/src/generator.rs
```diff
@@ -20,6 +20,121 @@ pub fn derive_template(input: TokenStream) -> TokenStream {
     }
 }
 
+#[derive(Default)]
```
djc, line 23: I'd like this refactoring split out in a separate commit, as this commit history makes it very hard to review the changes that were done to add the `config` argument.

Also I'd like the `TemplateArgs` definition and `impl` to go below `build_template()`, in order to be consistent with the rest of the files top-down ordering.

book/src/configuration.md
```diff
@@ -18,6 +18,7 @@ Here is an example that defines two custom syntaxes:
 ```toml
 [general]
 default_syntax = "foo"
+automatic_trim_block = true
```
djc, line 21: I think there should be some explanation of the behavior of this option in the text below.

djc (commented):
Did you see the earlier discussion in #598? In addition to the smaller questions below, I'm not sure I like the specified behavior much where the only effect of a `-` when `automatic_block_trim` is enabled is to **disable** the trim on the other side of the block. IMO the design I think we discussed earlier where a `+` could be added to locally revert the default seems much clearer, since it is explicit and does not cause sort spooky action at a distance (even if the distance is limited).

What do you think? I forget what Jinja and Tera do exactly...

GuillaumeGomez:
> Did you see the earlier discussion in #598? In addition to the smaller questions below, I'm not sure I like the specified behavior much where the only effect of a `-` when `automatic_block_trim` is enabled is to **disable** the trim on the other side of the block. IMO the design I think we discussed earlier where a `+` could be added to locally revert the default seems much clearer, since it is explicit and does not cause sort spooky action at a distance (even if the distance is limited).
> 
> What do you think? I forget what Jinja and Tera do exactly...

I just checked and they indeed use `+` to revert the automatic trimming (my bad, I thought the `-` was doing it). So if you agree, I can make the change.

askama_shared/src/generator.rs
GuillaumeGomez: Sure!

book/src/configuration.md
GuillaumeGomez: Agreed!

djc:
I think adding a `config` argument to the template attribute makes sense in principle, will review the changes in more detail once you've split the commit.

GuillaumeGomez:
Thanks for confirming. Then I'll make the changes as soon as possible.

djc:
Sounds great! I'll review this tomorrow.

askama_shared/src/generator.rs
```diff
@@ -129,6 +135,11 @@ struct Generator<'a, S: std::hash::BuildHasher> {
     buf_writable: Vec<Writable<'a>>,
     // Counter for write! hash named arguments
     named: usize,
+    // If set to true, the meaning of the minus sign (`-`) will be inverted on all blocks
```
djc, line 255: I believe this comment is out of date with regards to the `+` syntax.

askama_shared/src/generator.rs
```diff
@@ -1688,7 +1701,12 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     // prefix whitespace suppressor from the given argument, flush that whitespace.
     // In either case, `next_ws` is reset to `None` (no trailing whitespace).
     fn flush_ws(&mut self, ws: Ws) {
-        if self.next_ws.is_some() && !ws.0 {
+        if self.next_ws.is_none() {
+            return;
+        }
+        // If `automatic_block_trim` is enabled, we keep the whitespaces only if there is a `-`.
```
djc: This comment, too. While you're in there, mind adding an empty line after the `}` above?

askama_shared/src/parser.rs
```diff
@@ -129,6 +129,9 @@ pub(crate) enum Target<'a> {
     Path(Vec<&'a str>),
 }
 
+/// First field is "minus sign was used on the left part of the item".
+///
+/// Seonc field is "minus sign was used on the right part of the item".
```
djc: "Seonc" -> second?

askama_shared/src/parser.rs
```diff
@@ -129,11 +129,27 @@ pub(crate) enum Target<'a> {
     Path(Vec<&'a str>),
 }
 
-/// First field is "minus sign was used on the left part of the item".
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub(crate) enum Sign {
```
djc, line 133: Hmm, I'd prefer to give this more semantic meaning, like:

```rust
enum Whitespace { // Not sure about this name...
    Preserve, // Alternatively, `Keep`, ...
    Trim, // Alternatively, `Strip`, `Skip`, ....
}
```

askama_shared/src/parser.rs
```diff
@@ -660,6 +676,10 @@ expr_prec_layer!(expr_compare, expr_bor, "==", "!=", ">=", ">", "<=", "<");
 expr_prec_layer!(expr_and, expr_compare, "&&");
 expr_prec_layer!(expr_or, expr_and, "||");
 
+fn expr_minus_or_plus(i: &str) -> IResult<&str, Sign> {
```
djc, line 679: This name, too, is a bit too literal for me, something like `expr_handle_ws` maybe?

askama_shared/src/generator.rs
```diff
@@ -1697,16 +1697,20 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         self.prepare_ws(ws);
     }
 
+    fn should_trim_ws(&self, ws: Option<Sign>) -> bool {
+        ws == Some(Sign::Minus) || self.automatic_block_trim && ws != Some(Sign::Plus)
```
djc, line 1824: I feel like this could be clearer as a `match`?

```rust
match ws {
    Some(Sign::Minus) => true,
    Some(Sign::Plus) => false,
    None => self.automatic_block_trim,
}
```

askama_shared/src/generator.rs
```diff
@@ -66,7 +67,7 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
         &contexts,
         heritage.as_ref(),
         MapChain::new(),
-        config.automatic_block_trim,
+        automatic_block_trim,
```
djc, line 70: Can we just pass `config.automatic_block_trim` directly, or did pass ownership of the config somewhere?

askama_shared/src/lib.rs
```diff
@@ -327,12 +327,18 @@ struct RawEscaper<'a> {
     extensions: Vec<&'a str>,
 }
 
-fn read_config_file() -> std::result::Result<String, CompileError> {
+fn read_config_file(config_path: &Option<String>) -> std::result::Result<String, CompileError> {
     let root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
-    let filename = root.join(CONFIG_FILE_NAME);
+    let filename = if let Some(config_path) = config_path {
```
djc, line 332: I feel like a `match` expression would make this easier to read. (And please add a newline after it.)

testing/tests/gen_ws_tests.py
```diff
@@ -10,7 +10,10 @@
 IF, ELSE_IF, ELSE, END_IF = 0, 1, 2, 3
 
 NL = "\\n"
-dash = lambda ws: [" ", "-"][ws]
+
+
+def dash(ws, sign="-"):
```
djc, line 15: Let's rename `sign` here according to what the AST has. I'm not sure what `dash()` means here?

book/src/configuration.md
```diff
@@ -11,8 +11,37 @@ This example file demonstrates the default configuration:
 [general]
 # Directories to search for templates, relative to the crate root.
 dirs = ["templates"]
+# Unless you add a `-` in a block, whitespace won't be trimmed.
+automatic_trim_block = false
 ```
 
+About `automatic_trim_block`, if you want to trim whitespace before or after a
```
djc, line 18: This sentence is a bit awkward. Maybe something like, "In the default configuration, you can use the `-` operator to indicate that whitespace should be suppressed before or after a block. For example:".

book/src/configuration.md
```diff
@@ -11,8 +11,37 @@ This example file demonstrates the default configuration:
...
+
+
+{%- if something %}
+Hello
+(% endif %}
+```
+
+In this case, only the content between `<div>` and `{%-` will be trimmed. If
+this option is enabled, all content before and after each jinja block will be
+trimmed.
```
djc, line 32: Then we introduce the option here, "In the template above, only the whitespace between `<div>` and `{%-` will be suppressed. If you enable `automatic_trim_block`, whitespace before and after each block will be suppressed by default. To preserve the whitespace, you can use the `+` operator:"

book/src/configuration.md
```diff
@@ -11,8 +11,37 @@ This example file demonstrates the default configuration:
...
+If the option is enabled but you don't want a part to be trimmed, use the plus
+(`+`) sign:
+
+```jinja
+{% if something +%}
+Hello
+(%+ endif %}
+```
+
+In this example, `Hello` will be surrounded with backlines.
```
djc, line 43: I'm not familiar with "backlines", use "newline characters" instead?

askama_shared/src/parser.rs
djc: Having thought about this some more, I believe this feature is called "whitespace suppression" in Jinja and I think that's maybe clearer as to what we're doing than `trim`, so maybe the variants should be called `Suppress` and `Preserve`? Maybe the config option should be `whitespace_handling = "Preserve"`, too?

askama_shared/src/parser.rs
GuillaumeGomez: Indeed. ^^'

askama_shared/src/parser.rs
GuillaumeGomez: Sounds like a good idea. For the config, I think a boolean would be better, but renaming is ok I guess (I took the name from `flask`). What do you think about `preserve_whitespace`?

askama_shared/src/generator.rs
GuillaumeGomez: That is indeed much better.

book/src/configuration.md
GuillaumeGomez: :+1:

askama_shared/src/parser.rs
GuillaumeGomez: Ok!

askama_shared/src/parser.rs
djc: I feel like `suppress_whitespace` (flipping it) would be better because not suppressing whitespace is (a) the default and (b) more natural. What do you think?

GuillaumeGomez: Sounds good to me!

askama_shared/src/generator.rs
GuillaumeGomez: Yes we can! I think at some point it wasn't possible because of how I was trying to parse the derive proc macro. Fixed it,

testing/tests/gen_ws_tests.py
GuillaumeGomez: Renaming it to `sign` would be problematic for the parameters so I instead renamed it `sign_to_use`.

djc: No, I meant it should be more like `suppress`/`preserve` and less like `sign`/`minus`/`plus`.

GuillaumeGomez: The function or the parameter? :laughing: I guess you meant both. So in this case, the parameter will be `suppress`, but what about the function name?

djc: Oh, missed that the name was already dash. I guess the function could be called `suppress` and the param could be called `action` or `kind`?

GuillaumeGomez: Seems weird to have a function called `suppress` if an argument would actually be doing the opposite no?

djc: Maybe `handle_ws()`?

GuillaumeGomez: Let's go with this one then! :)

GuillaumeGomez:
Applied suggestions!

askama_shared/src/generator.rs
```diff
@@ -1688,7 +1704,13 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
     // In either case, `next_ws` is reset to `None` (no trailing whitespace).
     fn flush_ws(&mut self, ws: Ws) {
-        if self.next_ws.is_some() && !ws.0 {
+        if self.next_ws.is_none() {
+            return;
+        }
+
+        // If `suppress_whitespace` is enabled, we keep the whitespace characters only if there is
+        // a `+` character.
+        if self.automatic_block_trim == ws.0 {
```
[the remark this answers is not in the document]
GuillaumeGomez: I failed the rebase for this commit I guess. Fixing it.

GuillaumeGomez:
Fixed the invalid rebase. Sorry for the noise.

djc (approved):
This looks great, thanks!

GuillaumeGomez:
Any news here? :)

djc:
Given that it's been the Easter weekend, I'll give them like 24 more hours to respond. :)

GuillaumeGomez:
That is a very good point. We'll see if they survived the chocolate day. :laughing:

askama_shared/src/generator.rs
```diff
@@ -60,14 +61,135 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
...
+
+            if ident == "path" {
+                if let syn::Lit::Str(ref s) = pair.lit {
+                    if args.source.is_some() {
+                        return Err("must specify 'source' or 'path', not both".into());
+                    }
+                    args.source = Some(Source::Path(s.value()));
+                } else {
+                    return Err("template path must be string literal".into());
+                }
```
vallentin, line 144: I know it's just old code being moved. But the nested ifs could be replaced with a single match. _(Just did a quick copy, paste, format by hand.)_

@djc which do you prefer?

```suggestion
                match pair.lit {
                    syn::Lit::Str(ref s) = pair.lit if args.source.is_some() => return Err("must specify 'source' or 'path', not both".into()),
                    syn::Lit::Str(ref s) = pair.lit => args.source = Some(Source::Path(s.value())),
                    _ => return Err("template path must be string literal".into()),
                }
```

vallentin (commented):
I really like the direction this is heading. Unless I'm misinterpreting something, and please correct me if I am. For HTML I'd love if it was possible to say "Trim all but 1 space". Given that in HTML `<span>Foo</span><span>Bar</span>` vs `<span>Foo</span> <span>Bar</span>` are rendered differently.

While `<span>Foo</span> <span>Bar</span>`, `<span>Foo</span>         <span>Bar</span>`, `<span>Foo</span>\n\n<<span>Bar</span>`, `<span>Foo</span>\n\n   \n    \n\n\n<span>Bar</span>` are all rendered the same.

Anyways, this is more of a side-comment, but a feature I'd love to see extended later :)

askama_shared/src/generator.rs
djc: Let's do it in a separate PR. I think a `match` expression makes sense here, I think the `args.source` case could maybe be handled with an `ok_or()` or `ok_or_else()`?

vallentin: Separate PR it is 👍 Not sure how it would work with `ok_or()`/`ok_or_else()` as the error case is if it's `Some` not `None`? :)

djc: Ah yes, that would be the other way around.

djc: Depending on the formatting rustfmt comes up with, an inner `match` might be nicer than using guards on the outer match arms.

GuillaumeGomez:
> I really like the direction this is heading. Unless I'm misinterpreting something, and please correct me if I am. For HTML I'd love if it was possible to say "Trim all but 1 space". Given that in HTML `<span>Foo</span><span>Bar</span>` vs `<span>Foo</span> <span>Bar</span>` are rendered differently.
> 
> While `<span>Foo</span> <span>Bar</span>`, `<span>Foo</span> <span>Bar</span>`, `<span>Foo</span>\n\n<<span>Bar</span>`, `<span>Foo</span>\n\n \n \n\n\n<span>Bar</span>` are all rendered the same.
> 
> Anyways, this is more of a side-comment, but a feature I'd love to see extended later :)

It only impacts jinja block, not pure HTML. So unless you have a jinja block between your two `<span>`, it won't change the whitespace there (I hope you understood correctly ^^').

vallentin:
> It only impacts jinja block, not pure HTML. So unless you have a jinja block between your two `<span>`, it won't change the whitespace there (I hope you understood correctly ^^').

Yeah, I interpreted that much. But as in example, I have some loops like `<ul> {% for %} <li>...</li> {% endfor %} </ul>`, then there would be no spacing between the `<li>` elements, right? :)

GuillaumeGomez:
Absolutely! But if you want to keep them, you can use `+` like this: `<ul> {% for %+} <li>...</li> {+% endfor %} </ul>`.

vallentin:
> Absolutely! But if you want to keep them, you can use `+` like this: `<ul> {% for %+} <li>...</li> {+% endfor %} </ul>`.

Yeah, of course, and it's really nice, don't get me wrong :)

My point was "a feature I'd love to see extended later". I'd love to see an extension of this feature in the future, where it would be possible, to trim all but 1 space between all items :)

GuillaumeGomez:
Oh I see. Sorry I misunderstood your point. Yes, it would be a nice extension! Don't forget to open an issue so it's not forgotten. :wink:

Kijewski (approved):
I like the idea, and the implementation looks good to me. Nice!

GuillaumeGomez:
Thanks a lot everyone for the careful reviews and suggestions, it was very nice contributing here! I'll try to take a look at @vallentin's suggestion too.

djc:
@GuillaumeGomez thanks to you for doing the work on this change and dealing with my feedback!

PR #665, opened 2022-04-13
Remove `unsafe { … }` code from askama_escape

Kijewski:
Using only safe code is actually same as fast as the previous "unsafe"
code according to the crate's benchmark.

The code was extracted from [markup]'s escape function in [escape.rs],
written by Utkarsh Kukreti <utkarshkukreti@gmail.com>, licensed as
`MIT OR Apache-2.0`.

[markup]: https://crates.io/crates/markup
[escape.rs]: https://github.com/utkarshkukreti/markup.rs/blob/8ec40428483790b2c296e907e7be4147b157fe8f/markup/src/escape.rs#L1-L21

### DIFF
--- a/askama_escape/src/lib.rs
+++ b/askama_escape/src/lib.rs
@@ -107,40 +107,31 @@ where
 
 pub struct Html;
 
-macro_rules! escaping_body {
-    ($start:ident, $i:ident, $fmt:ident, $bytes:ident, $quote:expr) => {{
-        if $start < $i {
-            $fmt.write_str(unsafe { str::from_utf8_unchecked(&$bytes[$start..$i]) })?;
-        }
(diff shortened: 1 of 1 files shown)


djc (approved):
Nice!

Kijewski:
This is the only remaining unsafe line in the whole project AFAICT:
https://github.com/djc/askama/blob/d4c674064e0e95bbc68c99f7e458d0128020ac7f/askama_escape/src/lib.rs#L176 

What do you think, should I replace it with its safe variant, too? I don't think that anyone will put tons of JSON in their output, so the difference might be neglectable.

djc:
I'm honestly not too bothered about having some unsafe code, as long as the invariant is clearly upheld.

PR #668, opened 2022-04-19
Skip caching calls (Resolves #667)

### LINKED ISSUE #667: Cached methods call
It uses value of the first method call instead of calling the method each time. E.g.
```rust
use askama::Template;
use rand::prelude::*;

#[derive(Template)]
#[template(
    source = "{{ rnd.gen() }} {{ rnd.gen() }}",
    ext = "txt",
    print = "code"
)]
struct Foo {
    rnd: Rnd,
}

struct Rnd;
impl Rnd {
    fn gen(&self) -> i32 {
        rand::thread_rng().gen()
    }
}

fn main() {
    let foo = Foo { rnd: Rnd {} };
    dbg!(foo.render().unwrap());
}
```

Generates 
```rust
"{expr0} {expr0}",
expr0 = &::askama::MarkupDisplay::new_unsafe(&(self.rnd.gen()), ::askama::Text),
```

As you can see `{expr0}` is calculated only once.

vallentin:
This is just a quick fix to address #667, and open up discussion.

Should we take any assumptions about interior mutability, statics and side-effects? i.e. never cache calls. Someone could potentially even make a "funky" `impl Deref`, making caching field accesses "wrong".

Alternative suggestion, we could introduce a config to disable caching.

(I'll add tests, update docs, etc. when I know whether we want a config or just take no assumptions and not cache)

-----

New behavior:

```jinja
{{ foo() }} {{ foo() }} {{ bar }}  {{ bar }}
```

Now generates:

```rust
::std::write!(
    writer,
    "{expr0} {expr1} {expr2}  {expr2}",
    expr0 = &::askama::MarkupDisplay::new_unsafe(&(self.foo()), ::askama::Text),
    expr1 = &::askama::MarkupDisplay::new_unsafe(&(self.foo()), ::askama::Text),
    expr2 = &::askama::MarkupDisplay::new_unsafe(&(self.bar), ::askama::Text),
)?;
```

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -7,7 +7,7 @@ use crate::CompileError;
 use proc_macro::TokenStream;
 use quote::{quote, ToTokens};
 
-use std::collections::HashMap;
+use std::collections::hash_map::{Entry, HashMap};
 use std::path::{Path, PathBuf};
 use std::{cmp, hash, mem, str};
 
@@ -1196,11 +1196,9 @@ impl<'a> Generator<'a> {
                             expr_buf.buf, self.input.escaper
                         ),
                     };
-                    let is_cacheable = !matches!(s, Expr::Call(..));
 
-                    use std::collections::hash_map::Entry;
                     let id = match expr_cache.entry(expression.clone()) {
-                        Entry::Occupied(e) if is_cacheable => *e.get(),
+                        Entry::Occupied(e) if s.is_cachable() => *e.get(),
                         e => {
                             let id = self.named;
                             self.named += 1;
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1196,11 +1196,12 @@ impl<'a> Generator<'a> {
                             expr_buf.buf, self.input.escaper
                         ),
                     };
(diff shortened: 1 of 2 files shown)


askama_shared/src/generator.rs
```diff
@@ -1058,11 +1058,12 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
                             expr_buf.buf, self.input.escaper
                         ),
                     };
+                    let is_cacheable = !matches!(s, Expr::Call(..));
```
Kijewski, line 1061: I'm not sure if this doesn't need to be a "deep" check. E.g. `foo().0` should still be uncachable, right?

vallentin: Oh yeah, you're right. Guess my quick fix, was too quick. I'll fix it later 😅

djc: Should I review before you fix it or do you want to fix it before I review? Not sure what's more appropriate here.

Kijewski:
Rebased on main. I added a `Expr::is_cachable()` to test the recursive case.

Kijewski:
@vallentin, I hope it's okay that I took this over. :)

djc (approved):

djc:
Thanks!

Issue #671, opened 2022-04-21
Trim all but 1 whitespace jinja condition


GuillaumeGomez:
@vallentin suggested [here](https://github.com/djc/askama/pull/664#pullrequestreview-945253861) that we could extend the "trim_all_whitespace" feature to be able to trim all whitespace characters but one.

So currently, we have:

```jinja
<div>    {%- if foo -%} blabla {%- endif -%} </div>
```

Which will remove all whitespace characters before and after (by using `-`) and we have:

```jinja
<div>    {%+ if foo +%} blabla {%+ endif +%} </div>
```

Which means that the whitespace should be saved (only useful when the "trim all whitespace" feature is enabled).

The idea would now to add a new character to strip all whitespace characters but one. I propose `~`:

```jinja
<div>    {%~ if foo %} blabla {% endif %} </div>
```

djc:
I'm on board with all that, including the proposed operator.

GuillaumeGomez:
Ok, then I'll make a PR soon. :)

Kijewski:
I like that! For readability I'd like it best if a run of whitespaces becomes a single '\n' if it contains an '\n', or a space otherwise.

GuillaumeGomez:
I'm fine with implementing it this way. Anyone else has an opinion maybe?

djc:
Yeah, I think that approach makes sense.

vallentin:
I'd love, if it is also configurable like #664. So you can configure it through the config file or part of the `#[template(...)]`. As minifying HTML is part of my rendering pipeline, so if it was possible to enable it globally, without needing to sprinkle `~` everywhere, then that would be lovely :)

GuillaumeGomez:
Sure. What should we name the config parameter then?

GuillaumeGomez:
Also, if both "trim_all" and "trim_all_but_one" are enabled, what should we do?

vallentin:
Maybe instead of having them be individual. Maybe it might be possible to do `whitespace = "trim_all"` and `whitespace = "trim_all_but_one"`

GuillaumeGomez:
Maybe @djc was right: I should have gone with a string argument for "trim_whitespace" so then we wouldn't have this problem.

djc:
> Maybe @djc was right: I should have gone with a string argument for "trim_whitespace" so then we wouldn't have this problem.

Yup!

GuillaumeGomez:
Ok, then I'll change the parameter as well. :)

[...]

PR #672, opened 2022-04-21
Rename `suppress_whitespace` into `whitespace`

GuillaumeGomez:
As discussed in https://github.com/djc/askama/issues/671.

That'll make what follows simpler.

### DIFF
--- a/askama_shared/src/lib.rs
+++ b/askama_shared/src/lib.rs
@@ -118,7 +118,7 @@ struct Config<'a> {
     syntaxes: BTreeMap<String, Syntax<'a>>,
     default_syntax: &'a str,
     escapers: Vec<(HashSet<String>, String)>,
-    suppress_whitespace: bool,
+    whitespace: WhitespaceHandling,
 }
 
 impl Config<'_> {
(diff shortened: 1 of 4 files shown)


askama_shared/src/lib.rs
```diff
@@ -302,12 +306,33 @@ impl RawConfig<'_> {
...
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "serde", derive(Deserialize))]
+#[cfg_attr(feature = "serde", serde(field_identifier, rename_all = "lowercase"))]
+pub(crate) enum WhitespaceHandling {
+    /// The default behaviour. It will leave the whitespace characters "as is".
+    Preserve,
+    /// It'll remove all the whitespace characters before and after the jinja block.
+    Trim,
```
djc, line 316: `Suppress`, right?

askama_shared/src/lib.rs
```diff
@@ -118,7 +118,7 @@ struct Config<'a> {
     syntaxes: BTreeMap<String, Syntax<'a>>,
     default_syntax: &'a str,
     escapers: Vec<(HashSet<String>, String)>,
-    suppress_whitespace: bool,
+    whitespace_handling: WhitespaceHandling,
```
djc, line 121: Let's just call this `whitespace`, too.

askama_shared/src/lib.rs
GuillaumeGomez: Oh yes indeed! Sorry about that.

GuillaumeGomez:
Updated!

djc (approved):

PR #673, opened 2022-04-21
Add `minimize` config for `whitespace` and add handling of `~` in the jinja parser

### LINKED ISSUE #671: Trim all but 1 whitespace jinja condition

GuillaumeGomez:
Fixes #671.

### DIFF
--- a/askama_shared/src/generator.rs
+++ b/askama_shared/src/generator.rs
@@ -246,7 +246,7 @@ struct Generator<'a, S: std::hash::BuildHasher> {
     next_ws: Option<&'a str>,
     // Whitespace suppression from the previous non-literal. Will be used to
     // determine whether to flush prefix whitespace from the next literal.
-    skip_ws: bool,
+    skip_ws: WhitespaceHandling,
     // If currently in a block, this will contain the name of a potential parent block
     super_block: Option<(&'a str, usize)>,
     // buffer for writable
@@ -272,7 +272,7 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
             heritage,
             locals,
             next_ws: None,
-            skip_ws: false,
+            skip_ws: WhitespaceHandling::Preserve,
             super_block: None,
             buf_writable: vec![],
             named: 0,
@@ -1230,13 +1230,22 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     fn visit_lit(&mut self, lws: &'a str, val: &'a str, rws: &'a str) {
         assert!(self.next_ws.is_none());
         if !lws.is_empty() {
-            if self.skip_ws {
-                self.skip_ws = false;
-            } else if val.is_empty() {
-                assert!(rws.is_empty());
-                self.next_ws = Some(lws);
-            } else {
-                self.buf_writable.push(Writable::Lit(lws));
+            match self.skip_ws {
+                WhitespaceHandling::Suppress => {
+                    self.skip_ws = WhitespaceHandling::Preserve;
+                }
+                _ if val.is_empty() => {
+                    assert!(rws.is_empty());
+                    self.next_ws = Some(lws);
+                }
+                WhitespaceHandling::Preserve => self.buf_writable.push(Writable::Lit(lws)),
+                WhitespaceHandling::Minimize => {
+                    self.buf_writable
+                        .push(Writable::Lit(match lws.contains('\n') {
+                            true => "\n",
+                            false => " ",
+                        }))
+                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


askama_shared/src/generator.rs
```diff
@@ -1230,13 +1230,19 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
     fn visit_lit(&mut self, lws: &'a str, val: &'a str, rws: &'a str) {
         assert!(self.next_ws.is_none());
         if !lws.is_empty() {
```
djc, line 1232: We should flip the condition and turn it into an early return.

askama_shared/src/generator.rs
```diff
@@ -1230,13 +1230,19 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
-            if self.skip_ws {
-                self.skip_ws = false;
+            if self.skip_ws == WhitespaceHandling::Suppress {
```
djc, line 1233: Can we do a `match` here somehow? I don't remember exactly how `self.skip_ws` and `val` interact.

askama_shared/src/generator.rs
```diff
@@ -1819,11 +1825,12 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
         self.prepare_ws(ws);
     }
 
-    fn should_trim_ws(&self, ws: Option<Whitespace>) -> bool {
+    fn should_trim_ws(&self, ws: Option<Whitespace>) -> WhitespaceHandling {
         match ws {
-            Some(Whitespace::Trim) => true,
-            Some(Whitespace::Preserve) => false,
-            None => self.whitespace == WhitespaceHandling::Suppress,
+            Some(Whitespace::Trim) => WhitespaceHandling::Suppress,
```
djc, line 1830: We should rename `Whitespace::Trim` to `Suppress`, too.

askama_shared/src/parser.rs
```diff
@@ -1447,12 +1447,12 @@ mod tests {
     #[test]
     fn change_delimiters_parse_filter() {
         let syntax = Syntax {
-            expr_start: "{~",
-            expr_end: "~}",
+            expr_start: "{=",
```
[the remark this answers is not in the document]
djc: A warning is probably a good idea!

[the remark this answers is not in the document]
GuillaumeGomez: https://github.com/djc/askama/issues/674 :)

askama_shared/src/generator.rs
```diff
@@ -1230,13 +1230,19 @@ impl<'a, S: std::hash::BuildHasher> Generator<'a, S> {
...
+            if self.skip_ws == WhitespaceHandling::Suppress {
+                self.skip_ws = WhitespaceHandling::Preserve;
             } else if val.is_empty() {
                 assert!(rws.is_empty());
                 self.next_ws = Some(lws);
-            } else {
+            } else if self.skip_ws == WhitespaceHandling::Preserve {
                 self.buf_writable.push(Writable::Lit(lws));
+            } else if self.skip_ws == WhitespaceHandling::Minimize {
+                if lws.contains('\n') {
```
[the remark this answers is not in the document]
GuillaumeGomez: Sounds good to me!

askama_shared/src/generator.rs
GuillaumeGomez: Adding a new commit then. :)

askama_shared/src/generator.rs
GuillaumeGomez: We can't make it return early, some code after this condition needs to be run.

askama_shared/src/generator.rs
GuillaumeGomez: I replaced this with a `match`. Please tell me if it looks good to you.

GuillaumeGomez:
Updated!

djc (approved):
LGTM! Any other opinions?

askama_shared/src/parser.rs
```diff
@@ -132,14 +132,16 @@ pub(crate) enum Target<'a> {
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub(crate) enum Whitespace {
```
vallentin, line 133: Couldn't `enum WhitespaceHandling` be renamed into `enum Whitespace` and then reuse the same `enum Whitespace` for both things? Is there any real need to have two different types in this case? :)

GuillaumeGomez: I don't like too much to mix parser and config although they are the same type in here. Well, if @djc agrees with you, I'll merge them. :)

vallentin: To be fair, I do agree with that point-of-view. I was just curious if there was anything beyond that :)

GuillaumeGomez: No, just this "mental split". Nothing else. :smile:

djc: The main downside here is the ugly name of `WhitespaceHandling`, but I think it makes sense to keep them separate for now.

djc:
Thanks!

Issue #676, opened 2022-04-25
Why is virtually all the macro code inside askama_shared


jplatte:
I was surprised to see that syn, quote and proc-macro2 are not just build dependencies of my project, they're regular dependencies too through askama_shared. When I looked into it, I found all of the macro code being in askama_shared. What is the reason for this?

djc:
The reason is to minimize coupling between some of the other crates, which created some semver hazards before.

I'm struggling to come up with a reason for desiring a split here. What's the concern here? If you use serde for example, it will depend on serde_derive which depends on syn/quote/proc-macro2. Is there an issue with binary sizes?

jplatte:
If compiling for a different target triple than the host's, these libraries that are only ever used as part of the proc-macro code (as a compiler plugin on the host) will be compiled for both host and target. I haven't been cross-compiling so far, I just noticed this when analyzing dependencies of turbo.fish with cargo-depgraph. It seems wrong to me to have these be part of the runtime dependency graph.

An alternative to moving the code into the proc-macro crate would be moving it behind a feature flag that's activated by the proc-macro crate, though then you end up compiling the shared crate twice (at least with `resolver = "2"`) even for target=host because the build-dependencies graph has a different feature set on askama_shared than the regular dependency graph.

djc:
I think the problem is that proc-macro crates can only export procedural macros, and no other kinds of symbols. That may also complicate the testing situation. That said, I'm not opposed to someone experimenting with moving a bunch of code (back) into the askama_derive crate, but it probably won't make my priority list (though I'd be happy to review).

thburghout:
Thanks for fixing this issue! We ran into this in our cross-compiling toolchain with an error like https://github.com/meta-rust/meta-rust/issues/266 . It was fixed by using the git dependency of askama. When are you planning on releasing these changes?

Great project btw!

djc:
Going to close this for now. Will probably release after #700 has been merged.

PR #677, opened 2022-04-26
 Merge askama_derive into askama 

Kijewski:


Kijewski:
Ping @jplatte.

Per #676, IMO this makes the project much easier to understand.

djc:
We should keep askama_escape separate, that's actually used by other projects as well. I'll review the rest of this later.

askama_derive/src/input.rs
```diff
@@ -181,7 +181,7 @@ impl Default for Print {
 }
 
 #[doc(hidden)]
```
jplatte, line 183: Not needed now that the function is private.

askama_derive/src/generator.rs
```diff
@@ -1,21 +1,18 @@
...
 
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::{cmp, hash, mem, str};
 
 /// The actual implementation for askama_derive::Template
-#[doc(hidden)]
-pub fn derive_template(input: TokenStream) -> TokenStream {
-    let ast: syn::DeriveInput = syn::parse2(input).unwrap();
+pub(crate) fn derive_template(input: TokenStream) -> TokenStream {
```
jplatte, line 14: Can be a simple `pub` since proc-macro crates can't export anything other than proc macros anyways.

askama/Cargo.toml
```diff
@@ -17,20 +17,20 @@ maintenance = { status = "actively-developed" }
...
-num-traits = ["askama_derive/num-traits", "askama_shared/num-traits"]
-with-actix-web = ["askama_derive/with-actix-web", "askama_shared/actix-web"]
-with-axum = ["askama_derive/with-axum", "askama_shared/axum"]
-with-gotham = ["askama_derive/with-gotham", "askama_shared/gotham"]
-with-mendes = ["askama_derive/with-mendes", "askama_shared/mendes"]
-with-rocket = ["askama_derive/with-rocket", "askama_shared/rocket"]
-with-tide = ["askama_derive/with-tide", "askama_shared/tide"]
-with-warp = ["askama_derive/with-warp", "askama_shared/warp"]
+config = ["askama_derive/config"]
+humansize = ["askama_derive/humansize", "dep:humansize"]
```
jplatte, line 21: This raises the MSRV for 1.60, although it seems like no MSRV than stable is enforced anyways? Otherwise could rename the dependency (and rename it back in source via `external crate humansize_cr as humansize;`).

jplatte (commented):
So there wasn't really a need for a shared dependency between proc-macro and runtime code at all? 😮

askama/Cargo.toml
djc: Even though we haven't made the MSRV explicit, I'd prefer not to consciously require 1.60 yet.

askama_derive/src/generator.rs
djc: I'd prefer to use `pub(crate)` anyway, to make it clear what's happening to the human reader.

Kijewski:
>We should keep askama_escape separate, that's actually used by other projects as well. I'll review the rest of this later. 

Oh, it's used quite often actually. I did not notice. I unpicked c9e7657b4c5ac710457bffb1cd1b5771c569f085.

> So there wasn't really a need for a shared dependency between proc-macro and runtime code at all?

Once upon the time maybe there was. :)

> This raises the MSRV for 1.60, although it seems like no MSRV than stable is enforced anyways?

I accepted cargo's suggestion to use `dep:*` without thinking. I use renaming now. The tests now run with 1.53 (any maybe older toolchains, too, but well … 1.53 is the oldest one I have installed except the ancient 1.36 + 1.38). Except for the ui tests (tests that are supposed to fail).  Somewhen we broke them for the old versions even though we use version_check.

askama_derive/src/input.rs
Kijewski: Removed.

askama_derive/src/generator.rs
Kijewski: I think `#![deny(unreachable_pub)]` had a problem with `pub`, but I can be mistaken. IMO `pub(crate)` is clearer.

askama/src/lib.rs
```diff
@@ -73,7 +73,8 @@ pub use askama_shared::{
 pub mod mime {
     #[cfg(all(feature = "mime_guess", feature = "mime"))]
     #[deprecated(since = "0.11.1", note = "Use Template::MIME_TYPE instead")]
-    pub use crate::shared::extension_to_mime_type;
+    #[allow(non_upper_case_globals)]
+    pub const extension_to_mime_type: () = ();
```
djc, line 214: What's the point of this? It's the same symbol, but it has a different type.

askama_derive/src/lib.rs
```diff
@@ -2,9 +2,628 @@
...
+mod parser;
 
+/// The actual implementation for askama_derive::Template
 #[proc_macro_derive(Template, attributes(template))]
 pub fn derive_template(input: TokenStream) -> TokenStream {
-    askama_shared::derive_template(input.into()).into()
+    generator::derive_template(input)
+}
+
+#[derive(Debug)]
```
djc, line 27: Would you mind doing a separate commit before this one that moves all the config stuff into a separate `config` module? I think that would make a lot of this stuff clearer.

askama/Cargo.toml
```diff
@@ -39,7 +39,14 @@ mime_guess = []
 [dependencies]
 askama_derive = { version = "0.12.0", path = "../askama_derive" }
 askama_escape = { version = "0.10.3", path = "../askama_escape" }
-askama_shared = { version = "0.13.0", path = "../askama_shared", default-features = false }
+
+comrak = { version = "0.12", optional = true, default-features = false }
+humansize_ = { package = "humansize", version = "1.1.0", optional = true }
```
djc, line 44: Not a fan of the `_` suffixes here. Is there another way to do this?

djc (commented):
I think this is a good direction! Don't remember why I didn't do it this way in the first place...

Some nits, I'd like to take a closer look again after this.

askama/Cargo.toml
jplatte: Yes, use a different prefix / suffix 😛

The prefix / suffix doesn't need to be visible in the code, you can rename the dependency back via `extern crate humansize_ as humansize;` ([related post on my blog](https://blog.turbo.fish/cargo-features/#reusing-the-name-of-an-optional-dependency-for-a-feature)).

The cleanest solution would be to not rename, and use `dep:humansize` syntax for the dependency of the feature on the dependency. That is however a [new feature from 1.60](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features), so would raise the MSRV.

askama/src/lib.rs
Kijewski: To keep the deprecation notice, so the user gets a message how to proceed even thought their compilation will fail.

djc: It's an interesting idea, but I think we should just get rid of the symbol instead. Probably not a lot of people using it anyway.

askama/Cargo.toml
djc: We can rename the features instead of the dependencies, right?

jplatte: Sure, that works too. But it often causes a lot of confusion, especially when the name of the optional dependency was previously a feature name. Every optional dependency is automatically a feature (unless shadowed by an explicit feature as of 1.60), so there will be no error for activating the `humansize` feature.

askama/src/filters/mod.rs
```diff
@@ -7,23 +7,23 @@
...
 
 #[allow(unused_imports)]
 use crate::error::Error::Fmt;
 use askama_escape::{Escaper, MarkupDisplay};
 #[cfg(feature = "humansize")]
-use humansize::{file_size_opts, FileSize};
+use mod_humansize::{file_size_opts, FileSize};
 #[cfg(feature = "num-traits")]
-use num_traits::{cast::NumCast, Signed};
+use mod_num_traits::{cast::NumCast, Signed};
```
Kijewski, line 26: A `mod_` prefix is better than a `_` suffix, I guess. :) I'm open for other ideas.

jplatte: There is also the possibility of having overlapping module / external crate names and disambiguating using `::humansize` for the crate and `self::humansize` for the module.

djc: But the problem here isn't module vs crate names but crate name vs feature name, IIRC?

jplatte: That's what the other review thread is about, this looks separate to me.

Kijewski: No, it's the same issue.

jplatte: Oh, so `mod_` as a crate name prefix? Yeah I think that's confusing, I would prefer `crate` if there is going to be a prefix (or suffix).

Kijewski: To me a `crate` {pre,suf}fix sound like it's an alias for `crate::`. How about `pkg_*`, since the Cargo.toml option is called `package`?

djc:
This has a conflict now, mind rebasing? I'll merge it soon!

Kijewski:
Rebased.

djc (approved):

PR #678, opened 2022-04-26
Fix invalid check

GuillaumeGomez:
`block_start` is already checked just above.

Also a question about this code: shouldn't the character in common be between `*_start` and `*_end` instead of the current code?

### DIFF
--- a/askama_shared/src/lib.rs
+++ b/askama_shared/src/lib.rs
@@ -275,8 +275,8 @@ impl<'a> TryFrom<RawSyntax<'a>> for Syntax<'a> {
         let be = syntax.block_start.as_bytes()[1];
         let cs = syntax.comment_start.as_bytes()[0];
         let ce = syntax.comment_start.as_bytes()[1];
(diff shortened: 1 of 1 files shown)


Kijewski (approved):

Kijewski:
I'm not even sure if the two characters requirement couldn't be relaxed. I guess it speeds up the parsing, but I don't think that parsing the templates will take up much of the compiling time of a non-trivial project.

GuillaumeGomez:
Considering it's config setup, I'm not even sure if it'd have any impact at all...

Kijewski:
Hehe, I meant the code of the parser. But no, e.g. https://github.com/djc/askama/blob/7238fe2971917f547498e761ff252244cd5c992d/askama_shared/src/parser.rs#L213-L217 does not care how long the delimiters are. Nevertheless, this could be be changed in another PR.

GuillaumeGomez:
Oh indeed. Please open an issue! :D

Issue #674, opened 2022-04-21
Emit a warning in case a newly defined delimiter is actually an already existing delimier


GuillaumeGomez:
For example, if you define `{+` as `expr_start`, then it would be ignored because `askama` already handles it internally meaning it cannot be overloaded.

Currently it is silently ignored. Emitting a warning would be better.

PR #679, opened 2022-04-26
Emit an error if a syntax config already exists

GuillaumeGomez:
Fixes https://github.com/djc/askama/issues/674.

I realized that warnings were easy to ignore so I decided to turn it into an error. What do you think?

### DIFF
--- a/askama_shared/src/lib.rs
+++ b/askama_shared/src/lib.rs
@@ -281,8 +281,72 @@ impl<'a> TryFrom<RawSyntax<'a>> for Syntax<'a> {
             return Err(format!("bad delimiters block_start: {}, comment_start: {}, expr_start: {}, needs one of the two characters in common", syntax.block_start, syntax.comment_start, syntax.expr_start).into());
         }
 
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
The checks I'm running are far from enough. I'll update it.

djc:
Out of curiosity, is the custom syntax you're working with intended for rustdoc or some other project? What's the use case?

GuillaumeGomez:
No, it's just that I don't like having settings being silently ignored. That's pretty much it. ^^'

GuillaumeGomez:
It should be ready now.

djc:
So can you explain again what problem this is trying to solve? This seems like a bunch of complicated code trying to improve diagnostics for what's like an edge case of an edge case -- in other words, extremely niche...

GuillaumeGomez:
When your config adds a syntax which already exists. A problem I actually encountered in a test of this crate. :laughing:

djc:
Okay, I'm inclined not to take this.

GuillaumeGomez:
Understandeable!

PR #686, opened 2022-05-23
Fix `clippy::derive_partial_eq_without_eq` warning

Kijewski:


djc:
Thanks; I should have mentioned that I filed a clippy issue for this:

https://github.com/rust-lang/rust-clippy/issues/8875

Kijewski:
Yea, that should be fixed upstream. Interestingly enough only in askama_rocket the bug shows.

djc:
Because Rocket runs against nightly.

djc:
I guess with Rocket 0.5 rcs we actually don't need nightly anymore?

PR #689, opened 2022-05-24
Cache generated code

Kijewski:
* Speeds up compilation of a project a lot, when only one, a few, or no templates were changed.
* Speeds up compilation a little even for an initial compilation.
* Far better error messages in generated code.
* The generated code is easily debuggable.

### DIFF
--- a/askama_derive/Cargo.toml
+++ b/askama_derive/Cargo.toml
@@ -29,11 +29,14 @@ with-tide = []
 with-warp = []
 
 [dependencies]
+base64 = "=0.13"
+blake2 = "=0.10"
+indexmap = "=1"
 mime = "0.3"
 mime_guess = "2"
 nom = "7"
 proc-macro2 = "1"
 quote = "1"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-syn = "1"
+syn = { version = "1", features = ["extra-traits"] }
 toml = { version = "0.5", optional = true }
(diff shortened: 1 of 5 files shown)


Kijewski:
I'll tidy up the PR after y'all tell me what you think, e.g. introducing the use of IndexMap, and deriving Hash in their own commits.

Kijewski:
I could suppress the newly generated clippy warnings, too, but actually clippy is quite right with its warnings. Same condition in multiple elif-blocks, calculation X / X, etc. are 99% programming errors (outside of tests).

Kijewski:
I could suppress the newly generated clippy warnings, too, but actually clippy is quite right with its warnings. Same condition in multiple elif-blocks, calculation X / X, etc. are 99% programming errors (outside of tests).

djc:
Do you have benchmark numbers? In a bunch of other projects, I've adopted a different approach that I think might be better: generate code in an integration test. Here's one example, I've done a few others if you're interested:

https://github.com/mattsse/chromiumoxide/pull/80

So the idea would be that we have an askama_build crate that contains the generator and other stuff, which offers a builder API that lets you pass a string like `crate::foo::Bar` with a `Path` to the template for that type, and this would generate the `Template` implementations into a file (formatted with rustfmt) written into the `src` dir as part of an integration test. This is nice because:

* No build script, build scripts slow down compilation pipelines and are a supply chain hazard
* No recompilation of the templates unless needed, no procedural macro expansion except at test time
* Integration test run in CI makes sure that the generated code is never out of date

PR #697, opened 2022-06-16
Update to Rocket 0.5

Kijewski:
Though Rocket 0.5 still only a release candidate, Rocket 0.4 severely outdated, and depends on a bunch of crates with active security advisories. Rocket 0.5 updates its dependencies to fixes versions.

Also Rocket 0.4 needs a nightly Rust, which caused multiple problems.


djc (approved):
Nice!

PR #698, opened 2022-06-16
Add caching action to shrink CI times

ludicast:
- Adds GitHub caching step to shorten installation and build times.
- Extracts toolchain step into setup action to absorb commonalities

The cached folders are the suggested ones from the github's official [caching action](https://github.com/actions/cache/blob/main/examples.md#rust---cargo).

The most recent run with caching is at: https://github.com/ludicast/askama/actions/runs/2508931388.  Makes a large dent in CI time (admittedly the screenshotted example is a more pathological one :)):

from
Screenshot from 2022-06-16 08-26-52

to
Screenshot from 2022-06-16 08-30-16

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -33,4 +33,3 @@ runs:
         restore-keys: |
           ${{ inputs.key }}-${{ runner.os }}-${{ inputs.toolchain }}-${{ steps.toolchain-install.outputs.rustc_hash }}-
           ${{ inputs.key }}-${{ runner.os }}-${{ inputs.toolchain }}-
-          
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -3,13 +3,21 @@ name: Setup Rust Environment
 inputs:
   os:
     default: ubuntu-latest
-  rusthash:
-    description: Unique hash from toolchain action
-    required: true
+  toolchain:
+    default: stable
+  components:
+
 
 runs:
   using: composite
   steps:
+    - uses: actions-rs/toolchain@v1
+      id: toolchain-install
+      with:
+        profile: minimal
+        override: true
+        toolchain: ${{ inputs.toolchain }}
+        components: ${{ inputs.components }}  
     - uses: actions/cache@v3
       with:
         path: |
@@ -18,4 +26,4 @@ runs:
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
           target/
-        key: ${{ inputs.os }}-${{ inputs.rusthash }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+        key: ${{ inputs.os }}-${{ steps.toolchain-install.outputs.rustc_hash }}-cargo-${{ hashFiles('**/Cargo.lock') }}
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,10 +1,9 @@
 name: Setup Rust Environment
 
 inputs:
-  os:
-    description: Runner Type for generating hash-key
-    default: ubuntu-latest
-    required: false
+  key:
+    description: Cache key
+    required: true
   toolchain:
     description: Pass-through to toolchain on actions-rs
     default: stable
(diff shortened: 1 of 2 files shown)


djc (dismissed):
Looks nice, thanks!

Kijewski (changes requested):
Very nice, thank you! I'd make [some minor changes](https://github.com/Kijewski/nate/blob/41d4d0cc91bab8131bf55ae6a3b7c6db724b8254/.github/actions/setup-rust/action.yaml) to action.yaml in here, too:

* Remove the OS input, as it's known through the runner,
* add a key to tell the jobs apart,
* use restore-keys, and reorder the cache key a bit to make better use of the restore-key,
* and don't cache the binaries.

The binaries are downloaded from github as is, and if you put the into the cache, then rustup won't update components in subsequent runs.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,36 @@
...
+      with:
+        path: |
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: ${{ inputs.key }}-${{ runner.os }}-${{ inputs.toolchain }}-${{ steps.toolchain-install.outputs.rustc_hash }}-${{ hashFiles('**/Cargo.lock', '**/Cargo.toml') }}
+        restore-keys: |
+          ${{ inputs.key }}-${{ runner.os }}-${{ inputs.toolchain }}-${{ steps.toolchain-install.outputs.rustc_hash }}-
+          ${{ inputs.key }}-${{ runner.os }}-${{ inputs.toolchain }}-
```
ludicast, line 35: @Kijewski  - believe the `toolchain` is used as part of the rustc_hash output, are you sure you want both?

https://github.com/actions-rs/toolchain#outputs

.github/workflows/rust.yml
```diff
@@ -40,23 +39,19 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - uses: actions-rs/toolchain@v1
+      - uses: ./.github/actions/setup
         with:
-          profile: minimal
-          toolchain: stable
-          override: true
+          key: actix-web
```
ludicast, line 44: the rest of the jobs seem to invoke clippy and include it as a `component` input.  this job does not include it as a toolchain component but still executes clippy,

is the clippy component 
- needed here
- not needed anywhere
- fine as is (not used here, but used elsewhere)

.github/actions/setup/action.yml
Kijewski: So that when a the patch version of rustc is released, not everything has to be built anew.

.github/workflows/rust.yml
Kijewski: That's interesting, according to the [rustup book](https://rust-lang.github.io/rustup/concepts/profiles.html) the minimal profile should not contain clippy. I would add clippy in here, simply so we can be sure that it still works tomorrow.

Kijewski: > The minimal profile includes as few components as possible to get a working compiler (rustc, rust-std, and cargo).

> The default profile includes all of components in the minimal profile, and adds rust-docs, rustfmt, and clippy.

So, only cargo {build,test,check} should work with the minimal profile.

ludicast:
> Very nice, thank you! I'd make [some minor changes](https://github.com/Kijewski/nate/blob/41d4d0cc91bab8131bf55ae6a3b7c6db724b8254/.github/actions/setup-rust/action.yaml) to action.yaml in here, too:
> 
> * Remove the OS input, as it's known through the runner,
> * add a key to tell the jobs apart,
> * use restore-keys, and reorder the cache key a bit to make better use of the restore-key,
> * and don't cache the binaries.
> 
> The binaries are downloaded from github as is, and if you put the into the cache, then rustup won't update components in subsequent runs.

alright, these suggested changes are incorporated - thanks

Kijewski (approved):
LGTM, a big improvement! Thank you very much!

What do you think regarding the clippy component, @djc?

djc:
I think this is good to go. I'm sorry, I don't have time to dig out the context of the clippy issue. If you still want an opinion, can someone restate the full issue for me with all the necessary context?

PR #714, opened 2022-08-06
[DONT MERGE] Always take reference when accessing value passed as argument to a macro

### LINKED ISSUE #708: Iterating over elements of an iterator inside a macro moves the inner collection
For some reason, askama won't use references of the inner collection when using macros, resulting in a compiler error

```
error[E0507]: cannot move out of `foo.list` which is behind a shared reference
   --> src/main.rs:
    |
280 |     #[derive(Template)]
    |              ^^^^^^^^
    |              |
    |              `foo.list` moved due to this method call
    |              move occurs because `foo.list` has type `BTreeSet<std::string::String>`, which does not implement the `Copy` trait
    |
note: this function takes ownership of the receiver `self`, which moves `foo.list`
   --> /usr/lib/rustlib/src/rust/library/core/src/iter/traits/collect.rs:261:18
    |
261 |     fn into_iter(self) -> Self::IntoIter;
    |                  ^^^^

error[E0507]: cannot move out of `foo.list` which is behind a shared reference
   --> src/main.rs:
    |
280 |     #[derive(Template)]
    |              ^^^^^^^^
    |              |
    |              `foo.list` moved due to this method call
    |              move occurs because `foo.list` has type `BTreeSet<std::string::String>`, which does not implement the `Copy` trait
    |
```

When taking the code from the macro body and placing it at the call side, it works as expected.

This is a minimal template to reproduce the problem:

```jinja2
{%- macro list(list) -%}
<ul>
	{%- for item in list %}
	<li>{{item}}</li>
	{%- endfor %}
</ul>
<ol>
	{%- for item in list %}
	<li>{{item}}</li>
	{%- endfor %}
</ol>
{%- endmacro -%}

{%- for foo in foos %}
{% call list(foo.list) %}
{%- endfor %}
```

```rust
#[derive(Template)]
#[template(path = "list.j2", print = "code")]
struct List {
	foos: Vec<Foo>
}

struct Foo {
	list: BTreeSet<String>
}
```

<details><summary>Generated code by askama</summary><p>

```rust
impl ::askama::Template for List {
        include_bytes! ("/path/to/list.j2") ;
        {
            let mut _did_loop = false;
            let _iter = (&self.foos).into_iter();
            for (foo, _loop_item) in ::askama::helpers::TemplateLoop::new(_iter) {
                _did_loop = true;
                writer.write_str("\n")?;
                {
                    writer.write_str("<ul>")?;
                    {
                        let mut _did_loop = false;
                        let _iter = (foo.list).into_iter();
                        for (item, _loop_item) in ::askama::helpers::TemplateLoop::new(_iter) {
                            _did_loop = true;
                            ::std::write!(
                                writer,
                                "\n\t<li>{expr0}</li>",
                                expr0 = &::askama::MarkupDisplay::new_unsafe(&(item), ::askama::Html),
                            )?;
                        }
                        if !_did_loop {
                        }
                    }
                    writer.write_str("\n</ul>\n<ol>")?;
                    {
                        let mut _did_loop = false;
                        let _iter = (foo.list).into_iter();
                        for (item, _loop_item) in ::askama::helpers::TemplateLoop::new(_iter) {
                            _did_loop = true;
                            ::std::write!(
                                writer,
                                "\n\t<li>{expr1}</li>",
                                expr1 = &::askama::MarkupDisplay::new_unsafe(&(item), ::askama::Html),
                            )?;
                        }
                        if !_did_loop {
                        }
                    }
                    writer.write_str("\n</ol>")?;
                }
            }
            if !_did_loop {
            }
        }
        ::askama::Result::Ok(())
    }
    const EXTENSION: ::std::option::Option<&'static ::std::primitive::str> =
    Some("j2")
    ;
    const SIZE_HINT: ::std::primitive::usize =
    0
    ;
    const MIME_TYPE: &'static ::std::primitive::str =
    "application/octet-stream"
    ;
}
impl ::std::fmt::Display for List {
    #[inline]
    }
}
```

</p></details>

I'm using askama version 0.11.1 from crates.io.

msrd0:
This was proposed by @djc to fix #708 - but just blindly taking references does not seem to work

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -2034,7 +2034,7 @@ impl MapChain<'_, &str, LocalMeta> {
         let name = normalize_identifier(name);
         self.get(&name).map(|meta| match &meta.refs {
             Some(expr) => expr.clone(),
-            None => name.to_string(),
+            None => format!("&{name}"),
         })
     }
 
--- a/testing/templates/macro-iter.html.j2
+++ b/testing/templates/macro-iter.html.j2
@@ -1,0 +1,16 @@
+{%- macro list(list) -%}
+<ul>
+	{%- for item in list %}
+	<li>{{item}}</li>
+	{%- endfor %}
+</ul>
+<ol>
+	{%- for item in list %}
+	<li>{{item}}</li>
+	{%- endfor %}
+</ol>
+{%- endmacro -%}
+
+{%- for foo in foos %}
+{% call list(foo.list) %}
+{%- endfor %}
--- a/testing/tests/macro.rs
+++ b/testing/tests/macro.rs
@@ -1,4 +1,5 @@
 use askama::Template;
+use std::collections::BTreeSet;
 
 #[derive(Template)]
 #[template(path = "macro.html")]
@@ -72,4 +73,14 @@ struct StrCmpTemplate;
 fn str_cmp() {
     let t = StrCmpTemplate;
     assert_eq!(t.render().unwrap(), "AfooBotherCneitherD");
+}
+
+#[derive(Template)]
+#[template(path = "macro-iter.html.j2")]
+struct MacroIterTemplate {
+    foos: Vec<Foo>,
+}
+
+struct Foo {
+    list: BTreeSet<String>,
 }


djc:
Sorry, I don't have much time for this so if you want more help can you comment here with some of the generated code and point out where the problem happens?

msrd0:
It fails because it cannot compare `&str` with `str` (i.e. when the macro argument was already a borrowed string, it becomes a double borrow which cannot be compared to a literal anymore). You can see it in the CI here: https://github.com/djc/askama/runs/7705017639?check_suite_focus=true

Issue #717, opened 2022-08-28
Coerce Option<String> to ""


alper:
Is there an easy way to coerce an Optional<String> to be `""` if it's None? That would really make prototyping a lot faster than having to write out Match sequences.

Is something like: `{{ t.as_ref().unwrap() }}` the preferred way?

djc:
Well, I would do `{{ t.as_ref().unwrap_or("") }}` which doesn't seem too bad.

I'd also be open to a built-in filter, maybe `t|or_empty`?

septatrix:
When I try to use `unwrap_or("")` I get an error during compilation:

```
error[E0308]: mismatched types
   --> src/main.rs:90:10
    |
90  | #[derive(Template)] // this will generate the code...
    |          ^^^^^^^^
    |          |
    |          expected `&String`, found `&str`
    |          arguments to this method are incorrect
    |
    = note: expected reference `&std::string::String`
               found reference `&'static str`
help: the return type of this call is `&'static str` due to the type of the argument passed
   --> src/main.rs:90:10
    |
90  | #[derive(Template)] // this will generate the code...
    |          ^^^^^^^^ this argument influences the return type of `unwrap_or`
note: method defined here
   --> /home/septatrix/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:955:12
    |
955 |     pub fn unwrap_or(self, default: T) -> T {
    |            ^^^^^^^^^
    = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0308`.
error: could not compile `teamviewer-monitor` (bin "teamviewer-monitor") due to previous error
```

djc:
Yeah, well, if you do `unwrap_or("")` on an `Option<String>` the types don't line up (because `""` is a `&str` not a `String`). So I suppose you'd need `.as_deref().unwrap_or("")` for that case.

PR #718, opened 2022-08-28
Automatically unwrap Options for display

Kijewski:
It's the same as in Jinja: for a None value, simply write nothing.

### DIFF
--- a/askama/src/helpers.rs
+++ b/askama/src/helpers.rs
@@ -1,4 +1,7 @@
+use std::fmt;
 use std::iter::{Enumerate, Peekable};
+
+use askama_escape::MarkupDisplay;
 
 pub struct TemplateLoop<I>
 where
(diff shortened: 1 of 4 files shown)


Kijewski:
Please have a look, @alper, if that is what you meant.

alper:
Definitely what I want. I think this is an appropriate level of correctness for rendering templates.

djc:
Sorry, I think this is a little bit too magic. See my suggestion in #717.

alper:
I'm not exactly sure how Jinja templates work in this situation but the behaviour I'm used to is from Django.

I have to say that templates which are fully compiled and error out if variables are wrong in them is something that takes getting used to so maybe yeah, philosophically these will be a lot more strict than their equivalent python versions.

djc:
> I have to say that templates which are fully compiled and error out if variables are wrong in them is something that takes getting used to so maybe yeah, philosophically these will be a lot more strict than their equivalent python versions.

If I understand correctly, this wouldn't error out but would silently put an empty string in. That seems less problematic than yielding an error but I think something more explicit better matches the Rust experience.

alper:
I think a filter that catches most None cases and prints "" could be nice and would make it explicit. Maybe even printing "<EMPTY>" if some `DEBUG` value is turned on.

Kijewski:
> Sorry, I think this is a little bit too magic in a Rust context. See my suggestion in https://github.com/djc/askama/issues/717.

Yeah, you are right. This kind of magic is par for the course in Python, but very unrusty.

PR #725, opened 2022-09-23
Apply clippy suggestions from Rust 1.64

djc:


Kijewski (approved):

Kijewski:
Interesting, I didn't notice that the method was stabilized.

Issue #720, opened 2022-09-11
Documentation site shows unreleased features


natemcmaster:
There is gap between documented features in https://djc.github.io/askama/ and the actual behavior of the library when installing 0.11 (latest release on crates.io at time of writing.) Have you considered making it clearer which version is being documented?

Indicating more clearly the library version would help reduce confusion for new developers. For example, https://djc.github.io/askama/configuration.html documents a `whitespace` option in TOML config which doesn't work in 0.11. It took some digging to find the root cause. Root cause was that this feature was implemented in April (f997ba97bd298e219f3286fa792ec10874a34079) but hasn't been released yet.

Just a suggestion -- something like https://glebbahmutov.com/blog/versioned-doc-pages/ might help :) If this is something you would be open to but need help implementing, I could help.

By the way, this project is great! I look forward to the next release.

djc:
I'm definitely open to better versioned documentation, but unlikely to get to implementing it myself, so your help would be much appreciated!

natemcmaster:
Ok, here is a pull request which implements this. https://github.com/djc/askama/pull/726

Note: my source of inspiration was https://arctic-hen7.github.io/blog/dev/mdbook-versioning/ by @arctic-hen7

natemcmaster:
So, I see #726 was rejected. 

Can you give more specific guidance about what you would accept? I'm happy to help contribute, and I don't want to waste time trying to guess.

djc:
Yes, sorry that I've changed my mind on this. I think there are mostly two things we can do: only deploy released book https://djc.github.io/askama/ and/or mention the version at which some feature was introduced in the documenation for that feature (though I don't think we should go through all the old features and do that).

natemcmaster:
Ok. Both seem reasonable. The main motivation for me starting this thread was I thought the `whitespace` option in TOML was broken, but the real reason is that the feature hasn’t been released. Either option you suggested would have avoided the confusion.

avlec:
0.11.2 is in the "book" and doesn't work as it's not published on crates.io

djc:
Can you send a PR to fix it? Honestly in my mind it's not a big deal, these days I think it's okay to assume that people know how to add a library to their `Cargo.toml`.

PR #726, opened 2022-09-25
Create versioned documentation

natemcmaster:
Addresses #720 

This creates a versioned documentation site. I've deployed this in my fork for example:

* Stable docs: https://natemcmaster.github.io/askama/ (notice how this redirects to 0.11.x)
* Unstable docs: https://natemcmaster.github.io/askama/next/index.html
* Outdated releases: https://natemcmaster.github.io/askama/0.10.x/index.html

**Where did docs in 0.10.x/ and 0.11.x/ come from?**

I checked out these folders of documentation from the book/ folder at https://github.com/djc/askama/releases/tag/0.11.0 and https://github.com/djc/askama/releases/tag/0.10.0


djc:
Thanks for doing the work on this. It seems relatively ugly to have multiple copies of the documentation checked out -- and I'm also not a big fan of having a bunch of shell script orchestrating this. Relative to the required complexity, I feel like I'm okay with the current solution of only having mainline docs? Askama evolves fairly slowly anyway...

Kijewski:
I wonder if we could convert the book to a crate like https://docs.rs/dtolnay/0.0.11/dtolnay/ … So it would be versioned, could be easily linked to, would have the same look and feel as other crates, etc.

natemcmaster:
I didn’t spend much time in this, so no worries about closing and going another direction. 

Some alternatives to explore:

- just have one version of docs, but put annotations inline about features and when they were added. Downside: this can be cumbersome to maintain and easy to misinterpret. 
- bundle all docs into the crates
- Use git to checkout tags instead of putting copies side by side in the main branch
- just maintain a “stable” and “nightly” version of docs 
- release updates frequently so there isn’t a big gap between features and docs.

djc:
I prefer the user experience of mdbooks over manual/user guide documentation on docs.rs. Checking out tags would make it a little bit better, I guess? I definitely don't care about docs for older versions than stable -- but still feels like a lot of indirection when Askama evolves only slowly -- and the documented interface even more slowly.

PR #727, opened 2022-09-26
Add render benchmark

Kijewski:
The code is a simple copy of `template-benchmarks-rs`.

So you can tell if a change is good, bad, great or terrible.


djc:
Like https://github.com/djc/askama/blob/main/testing/benches/all.rs?

Kijewski:
Oops, yeah. I did not look into `testing` to find the benchmarks. :)

What do you think about adopting the big_table code to use the iteration count to size the dataset instead of iterating more often? This my we'd measure the inner loop speed instead of the outer.

djc:
> What do you think about adopting the big_table code to use the iteration count to size the dataset instead of iterating more often? This my we'd measure the inner loop speed instead of the outer.

Can you make an argument for why the inner loop speed is more interesting/representative than the outer loop speed?

PR #734, opened 2022-11-04
Fix some minor issues in escape documentation.

AndrolGenhald:
- Remove forward slash and backslash from the list of escaped characters in the book since they [aren't escaped](https://github.com/djc/askama/blob/1b18bab91ba773425e521abc276a1278e0d61f3c/askama_escape/src/lib.rs#L126-L130)
 - Update the OWASP link, which used to include forward slash in the list of recommended escapes, but no longer does (askama's HTML escaper currently matches the OWASP example)
 - Update the `escape` filter documentation to remove the incomplete list of escaped characters.
 - Add example for specifying alternate escaper to `escape` filter documentation.

### DIFF
--- a/askama/src/filters/mod.rs
+++ b/askama/src/filters/mod.rs
@@ -59,11 +59,14 @@ where
     Ok(MarkupDisplay::new_safe(v, e))
 }
 
-/// Escapes `&`, `<` and `>` in strings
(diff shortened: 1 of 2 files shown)


djc (approved):
Thanks!

PR #736, opened 2022-11-04
chore: apply clippy suggestions for 1.65

djc:


askama/src/lib.rs
```diff
@@ -163,7 +163,7 @@ impl fmt::Display for dyn DynTemplate {
 }
 
 #[cfg(test)]
-#[allow(clippy::blacklisted_name)]
+#[allow(clippy::disallowed_names)]
```
Kijewski, line 166: Not strictly part of this PR, but maybe it would be good to comment which names exactly are disallowed?

Otherwise: LGTM

Kijewski (approved):

PR #737, opened 2022-11-04
bump axum-core

NicolaLS:
running `cargo audit` I got this:
Crate:     axum-core
Version:   0.1.2
Title:     No default limit put on request bodies
Date:      2022-08-31
ID:        RUSTSEC-2022-0055
URL:       https://rustsec.org/advisories/RUSTSEC-2022-0055
Solution:  Upgrade to >=0.2.8, <0.3.0-rc.1 OR >=0.3.0-rc.2
Dependency tree:


djc:
I don't think there is a need to bump this dependency. All crates that update Askama will likely also update axum-core and get 0.2.8 as part of that update.

PR #739, opened 2022-11-05
Allow inserting arbitrary code

### LINKED ISSUE #682: Support rust imports
While you probably won't have a lot of rust code inside templates, having the possibility to import e.g. a trait like `Borrow` (https://github.com/djc/askama/issues/330#issuecomment-1120411565).

You currently have to `use` the trait in the surrounding rust file, but when e.g. having  super template used across different modules, it would be helpful to allow adding some `use`'s at the top of a template.

```
{% use std::borrow::Borrow %}
``` 

I'm not familiar with the parsing of askama, but it should be possible to parse the `use` statement with syn and just add them to the expansions.

The only issue could be conflicts, but those should occur very rarely and relatively easy to understand for a rust user.

Kijewski:
This can be used in any cases when Askama's built-in features are too restrictive for a very special corner case, e.g. to `use` some library. It should be used seldom, e.g. for prototyping before you clean up your code.

Resolves #682.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -662,6 +662,12 @@ impl<'a> Generator<'a> {
                     self.write_buf_writable(buf)?;
                     buf.writeln("continue;")?;
                 }
+                Node::Code(ws, code) => {
+                    self.handle_ws(ws);
+                    self.write_buf_writable(buf)?;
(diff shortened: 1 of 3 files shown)


Kijewski:
I'll add documentation to the book if the PR is accepted.

Amy7944 (approved):

PR #742, opened 2022-11-09
Fail to parse if keyword is not longest ident

Kijewski:
This PR makes e.g. `{% leta = b %}` a parsing error. To the reader it would appear that `leta` should be a meaningful expression in Askama, which it is not. Before this PR, `leta` was tokenized as `let` + `a`. This PR makes the parser try to find the longest identifier at a parsing positions and only compare the outcome against the expected keyword.

This is potentially a breaking change, because code that should always have been invalid will now fail to compile, when it was accepted before.

### DIFF
--- a/askama_derive/src/parser.rs
+++ b/askama_derive/src/parser.rs
@@ -238,6 +238,17 @@ fn skip_till<'a, O>(
     }
 }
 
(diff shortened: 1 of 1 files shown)


djc:
Ohh, good call, thanks!

Issue #743, opened 2022-11-10
[Feature request] Turbo-fish syntax in custom filters


saona-raimundo:
I have a custom filter that is naturally parametrized by a `usize` and I was hoping to implement it in a turbo-fish manner with a `const` parameter.

My custom filter indents the string by prefixing a string, a number of times, to each line. The code is the following.
```rust
    pub fn indent_by<const N: usize>(s: &str) -> ::askama::Result<String> {
    	let mut replace = "\n".to_string();
    	replace.extend(["    "; N]);
        Ok(s.replace("\n", &replace))
    }
```

I would like to call this filter by writing the filter `indent_by::<2>` in the Askama template (which is a stateless function). 

## Use-case

I am translating the Rule-Based approach for transformations in XLST to Rust. The idea is to code a function that takes an (XML) input and outputs another text format. Sometimes, this can be done by <em>delegating</em> the processing to the children. In light of #575, I implemented it as you suggested: having fields that are templates themselves. the result makes a lot of sense from the Rule-Based approach. 

By using `{{ child_template.render()?|safe }}`, I can locate the children result in a template, but to have full control of the output, I need to indent it like this `{{ child_template.render()?|indent_by::<X>|safe }}`.

I could indent as a post-processing, but I lose a lot of control over the output.

### Example

Note: I tried to come up with a minimal example, so I had to strip a few things. 
We want to display a poem, written in XML, in HTML. The input is
```xml
<poem>
   <author>Rupert Brooke</author>
   <title>Song</title>
</poem>
```
The desired output is
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <title>
        Song
    </title>
</head>
<body>
    <article>
        <h1>
            Song
        </h1>
        <address>
            Rupert Brooke
        </address>
    </article>
</body>
</html>
```

My approach is to have 3 templates: `File`, `Title`, and `Author`, where  `File` delegates to the others to achieve a Rule-Based approach. These are the templates.
```
{# file.html.askama -#}

<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <title>
        {{ title.render()?|indent_by::<2>|safe }} {#- delegation #}
    </title>
</head>
<body>
    <article>
        <h1 class="display">
            {{ title.render()?|indent_by::<3>|safe }} {#- delegation #}
        </h1>
        {{ author.render()?|indent_by::<2>|safe }} {#- delegation #}
    </article>
</body>
</html>
[...]

Issue #102, opened 2018-07-04
Add documentation about custom filters


djc:

djc:
I should also document the list of built-in filters.

Cokemonkey11:
Would be great to have a more complete example of a custom filter.

djc:
Fair point! For now, have a look at this test case:

https://github.com/djc/askama/blob/master/testing/tests/filters.rs#L39

Cokemonkey11:
Thanks! I managed to make some filters with this test case as reference:

```
/// Any filters defined in `mod filters` are accessible in your template documents.
mod filters {
    pub fn extension(s: &str) -> ::askama::Result<String> {
        Ok(s.split('.').last().expect("Couldn't split by '.'").into())
    }

    pub fn uscore_to_space(s: &str) -> ::askama::Result<String> {
        Ok(s.replace("_", " ").into())
    }
}
```

PR #744, opened 2022-11-10
Extend documentation of custom filters

saona-raimundo:
Related to #743 and #102 

Extends documentation of custom filters.

### DIFF
--- a/book/src/filters.md
+++ b/book/src/filters.md
@@ -13,7 +13,8 @@ is passed to the next.
 {{ "HELLO"|lower }}
 ```
 
-Askama has a collection of built-in filters, documented below, but can also include custom filters. Additionally, the `json` and `yaml` filters are included in the built-in filters,
+Askama has a collection of built-in filters, documented below, but can also include custom filters. 
+Additionally, the `json` and `yaml` filters are included in the built-in filters,
 but are disabled by default. Enable them with Cargo features (see below for more information).
 
 **Table of contents**
@@ -105,7 +106,10 @@ Output:
 Escape &lt;&gt;&amp;
 ```
 
-Optionally, it is possible to specify and override which escaper is used. Consider a template where the escaper is configured as [`escape = "none"`]. However, somewhere escaping using the HTML escaper is desired. Then it is possible to override and use the HTML escaper like this:
+Optionally, it is possible to specify and override which escaper is used. 
+Consider a template where the escaper is configured as [`escape = "none"`]. 
+However, somewhere escaping using the HTML escaper is desired. 
+Then it is possible to override and use the HTML escaper like this:
 
 ```jinja
 {{ "Don't Escape <>&"|escape }}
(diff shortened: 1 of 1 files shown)


book/src/filters.md
```diff
@@ -408,18 +408,25 @@ This will output formatted YAML for any value that implements the required
 ## Custom Filters
 [#custom-filters]: #custom-filters
 
-To define your own filters, simply have a module named filters in scope of the context deriving a `Template` impl.
+To define your own filters, simply have a module named `filters` in scope of the context deriving a `Template` impl and defines the filters as functions within this module. The functions must have a first argument `&str`, but otherwise can have other inputs. The output must be `::askama::Result<String>`.
```
djc, line 411: There are actually no restrictions on the arguments, and while the return type must be an `::askama::Result<T>`, `T` can be of any type.

djc (commented):
Thanks, some great improvements here!

book/src/filters.md
saona-raimundo: Thank you! 
I was checking it again and wanted to say that "the first argument to the filter is a reference to the expression it is applied to", but it is not totally true and found some unexpected (for me) behaviour:

Consider the templates
```
{{ 2|my_filter }}
```
and
```
{{ (1 + 1)|my_filter }}
```

I would expect both to work the same, but they do not: the first case is `2` is given directly to `my filter`, while in the second case `&(1 + 1)` is given as reference.

The problem arises for the signature of the filter: should it take a reference or not?
```
fn my_filter(n: usize) -> ::askama::Result<String>
```
works for the first case and
```
fn my_filter(n: &usize) -> ::askama::Result<String>
```
works for the second.

**Question.**
Is this expected for you, or should `2` be past a reference `&2`?

saona-raimundo (commented):
Question on macro expansion: should expressions always be given by reference to filters?

book/src/filters.md
djc: I think this is expected, although I agree that it is quite subtle -- would be great to document it! I think the recommended way to abstract over the difference would be to use a trait. For example, `std::fmt::Display` has a blanket implementation for `&T where T: std::fmt::Display`, so you could take an `impl std::fmt::Display` and receive either a `&T` or a `T`. In other cases, you might be able to define your own trait in a similar way.

saona-raimundo: Okay!
To document what values are given as references and which are not, where can I look?
I would assume that everything that needs evaluation, is the result of a filter, or is a field is passed by reference. But literal values are passed without reference. Where should I look for these "literal values"?

djc: I think you should not document the exact rules, since they might change across otherwise semver-compatible releases. See `Expr::is_copyable()` for the heuristics we're using. That method is actually slightly misnamed -- naming it the other way around might make more sense (something like `needs_borrow()`).

saona-raimundo: Perfect! I changed the wording, I hope it is clear enough without too much detail.

book/src/filters.md
```diff
@@ -408,19 +408,29 @@ This will output formatted YAML for any value that implements the required
 ## Custom Filters
 [#custom-filters]: #custom-filters
 
-To define your own filters, simply have a module named filters in scope of the context deriving a `Template` impl.
+To define your own filters, simply have a module named `filters` in scope of the context deriving a `Template` impl and define the filters as functions within this module. The functions must have at least one argument and the return type must be `::askama::Result<T>` where `T` is some type that implements `Display`.
```
djc, line 411: I'd prefer if you use some hard line wrapping (at about 100 columns).

Also, I think it's not true that a filter must return a `T` that implements `fmt::Display` -- that's what can be used in a template expression, but since filters can be chained any particular filter doesn't have to yield that.

book/src/filters.md
```diff
@@ -408,19 +408,29 @@ This will output formatted YAML for any value that implements the required
 ## Custom Filters
 [#custom-filters]: #custom-filters
 
-To define your own filters, simply have a module named filters in scope of the context deriving a `Template` impl.
+To define your own filters, simply have a module named `filters` in scope of the context deriving a `Template` impl and define the filters as functions within this module. The functions must have at least one argument and the return type must be `::askama::Result<T>` where `T` is some type that implements `Display`. 
 
-Note that in case of name collision, the built in filters take precedence.
+The arguments to the filters are passed as follows. The first argument corresponds to the expression they are applied to. Subsequent arguments, if any, must be given directly when calling the filter. The first argument <em>may</em> be passed as reference whenever Askama can not determine that it can be passed directly. If you want a filter to accept a first argument with or without reference, you may use traits to bound the argument. For example, the `trim` built-in filter accepts any value implementing `Display`. Its signature is similar to `fn trim<T: fmt::Display>(s: T) -> ::askama::Result<String>`.
```
djc, line 413: "may be passed as a reference whenever Askama cannot determine that it can be passed directly" seems a bit awkward. How about something like:

> The first argument may or may not be a reference, depending on the context in which the filter is called. To abstract over ownership, consider defining your argument as a trait bound.

In your `trim()` example, I would these days `impl fmt::Display` instead of the indirection through a type variable.

djc (commented):
Great work, some minor clarifications/edits suggested.

book/src/filters.md
saona-raimundo: Thanks! That is totally true!
Changed to mention that is only needed at the end of a chain of filters.

book/src/filters.md
saona-raimundo: Perfect, much better :) Thanks!

djc (approved):

Issue #425, opened 2021-01-03
Feature: ability to render .html file updates without rebuilding?


seenickcode:
I understand that Askama uses compiled templates but maybe is there a way to have a debug mode that renders the template when the page is loaded?

djc:
There is currently not. Askama is really two things: a parser for a template language to an AST, and a code generator that transpiles the AST to Rust code. I'd be open to some other "backend" for the AST (maybe one that recasts the Askama AST as Tera AST, or some other dynamic template engine), but it's a relatively large undertaking which I probably won't prioritize in the near future unless someone wants to pay for the efforts.

djc:
What's the use case, BTW? Just the compile time benefits, or actual "dynamic" usage, something else?

Another approach could be to run the Askama-generated Rust code through a JIT compiler or interpreter at run-time, but I'm not aware of any efforts in this direction (other than Miri, but it has a very different use case).

seenickcode:
In short, I'm looking for a faster dev workflow, than "update .html file (askama template)", stop server, `cargo build`, `cargo run`.

I absolutely love it that templates are compiled and you get the checks that come out of that (something that Go Templates do not have out of the box at least). But OTHO I'm seeking a more convenient way of doing Askama + Rust dev. 

(Context: I'm re-writing fluttercrashcourse.com, which is effectively a bespoke ecommerce application written in Vue/Nuxt + Go into Rust + actix-web + Askama. Going excellent thus far btw with Askama, and I'm even brand new to Rust).

djc:
Fair enough. I'm not even sure the interpreting approach could work short of something that can interpret Rust -- I think it would require "flattening" context types into something easier to use dynamically. So this is definitely a bunch of work.

Another approach that might work is to write templates that also work in Tera, I guess -- we would likely accept patches that would make this kind of thing easier.

vallentin:
I'd definitely be up for an interpreter, to be able to dynamically render templates. However, I immediately see two issue.

[...]

PR #748, opened 2022-11-23
Extract parser

couchand:
A rough pass at extracting the parser code from the derive generator.  A basic AST provides what's needed for the generator, but  much more care would need to be given to the public API.

A bare-bones AST printer demonstrates a second client.  Slight modifications were made to the parser to ensure all structures could be round-tripped.

Motivations include a personal project to implement an alternate Askama backend, as well as #425 maybe?


couchand:
Questions:

- [ ] threshold question: should the refactor be done?
- [ ] api design question: what should the AST data structures look like?
- [ ] fmt crate question: should this be finished?  what should it look like/do, beyond proving the concept?
- [ ] others, i'm sure

couchand:
> What's the goal of the alternate backend?

More or less #108, and then seeing where things could go from there.

djc:
#108 seems a relatively small feature to write an entirely new backend for, though... Why is this so important to you?

So I'm not opposed in principle to exposing the parser API through a separate crate (this actually used to be the case with the askama_shared crate). An AST printer doesn't seem particularly valuable to me; usually the `Debug` impl of the AST node types we already got is a decent start when pretty-printed.

For a larger change like this, if you want me to review this I'd need a clean commit history with small commits that make one logical change each. I guess my question is, why do you want to do this work upstream rather than just fork it? I have a lot of crates to maintain, and a lot of incoming PRs to review so the question for me is, if you're going out and build your own backend, how much work do you are you effectively asking me to do to support your project, and how is that work valuable to me?

couchand:
This is exactly the kind of discussion I wanted to open this PR to start.  My thinking on it is, if it's something you'd generally be doing anyway, I'm happy to help drive it forward.  If it's something you see as getting in your way, I won't bother pursuing it.

The fact that it used to be separate might counsel in favor of or against it being a good change to make at this point.

----

Asides, to answer your various questions:

> #108 seems a relatively small feature to write an entirely new backend for, though... Why is this so important to you?

If that's all it were, yes.  It's a minor annoyance but not difficult to `|safe`.

But I think there's opportunity in viewing the model slightly differently, making templates into components... The child template render as not just sticking a string in the right place, really none of it should be.  The Askama parser is the first step in a tokenizer for a templated markup parser.  (Apologies about the winding  and vague explanation, I'm not sure I have a clear idea yet!)

> if you want me to review this I'd need a clean commit history with small commits that make one logical change each. 

Roger that, sorry to assault you with this terrible history.

> why do you want to do this work upstream rather than just fork it?

Short answer, because Askama is awesome!  There's a community and tooling and people know it.  Better to glom on to all that if possible.  Also, it seems like the sort of refactor that's abstractly good to upstream.

> how much work do you are you effectively asking me to do to support your project, and how is that work valuable to me?

Totally valid question.  First of all, thanks for spearheading this and so many crates (loving quinn btw), and thanks for engaging with my wandering analysis, I'd have written a shorter PR but I didn't have the time :-p

As I mentioned above, if this is going to be a drain on your maintenance then let's close it right away.  I'm more than happy to drive the implementation, deal with rebases over changes that happen meanwhile, and of course please tag me if I end up causing you a bug or something.  But if you see the refactor being a net-negative on the project long term, close it without prejudice.

As far as the fmt utility is concerned, I really only included that to prove that the implementation was complete.  I'm not sure it's worth publishing on its own, but I do want to work toward a more configurable one that I can use to auto-format the legions of templates I've now accumulated.  And if one day I could have a tool that auto-applies fixes to all of my templates that would be really cool, too.  Anyway, that's the direction I was thinking with this stuff....

djc:
I'm probably okay with exposing the parser as a separate crate and I would then be open to reviewing individual changes (one per commit) that attempt to improve the parser API. Past that, not sure.

couchand:
:+1:  given the immediate future of child care I can confidently say I won't be picking this up again until next year.

In the meantime, if you have any insight or thoughts on the shape of the public API of the parser, please do write them down!

djc:
Nope, all those insights and thoughts have been paged out for a while now; I'll await your proposals.

couchand:
Closed in favor of #782

PR #766, opened 2023-01-24
Update `toml` dependency

Kijewski:
This PR replaces `toml` by `toml_edit`. Since version 0.6, `toml` is a wrapper around `toml_edit`, and the more basic library already meets our needs.

Also the configuration is made `'static`, because `toml` and `toml_edit` both needs to implement serde's `DeserializeOwned` by now. We allocate the strings once per template, so it is very unlikely that this change will have any measurable impact, neither in compile time nor RAM usage.

### DIFF
--- a/askama_derive/Cargo.toml
+++ b/askama_derive/Cargo.toml
@@ -14,7 +14,7 @@ rust-version = "1.58"
 proc-macro = true
 
 [features]
-config = ["serde", "toml"]
+config = ["serde", "toml_edit"]
 humansize = []
 markdown = []
 urlencode = []
@@ -38,4 +38,4 @@ proc-macro2 = "1"
 quote = "1"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 syn = "1"
-toml = { version = "0.5", optional = true }
+toml_edit = { version = "0.19", optional = true, features = ["serde"] }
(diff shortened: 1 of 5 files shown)


djc:
Would you mind splitting this into one change to facilitate `DeserializeOwned` and a separate change to bump the version? Seems a little ugly that the toml crate no longer supported borrowed data...

Kijewski:
I split the PR into two commits.

As far as I can tell from a quick glance into the code, they are not parsing the input like other serde libraries anymore in one step, but they are building a document first, and then extract the information from it. Rather like the difference between an SAX and a DOM parser.

askama_derive/Cargo.toml
```diff
@@ -37,4 +37,4 @@ proc-macro2 = "1"
 quote = "1"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 syn = "1"
-toml = { version = "0.5", optional = true }
+toml = { version = "0.6", optional = true }
```
djc: So this should go in the next commit, right?

djc (approved):

askama_derive/Cargo.toml
Kijewski: This change was not need at all anymore, because the next commit replaces `toml` with `toml_edit`. I removed the change.

djc: The point of the first commit was to change the code to not depend on types borrowing data from the input TOML `&str` without changing any other code (for the update).

Kijewski:
Yes, I updated the description of the first commit, and upgraded toml_edit to 0.19 in the second commit.

djc (approved):
Thanks! It's pretty ugly that toml_edit now depends on a nom fork...

PR #767, opened 2023-01-24
Update `ui` test for trybuild 1.0.76

Kijewski:
In <https://github.com/dtolnay/trybuild/pull/219> the output of error messages was subtly changed, because they introduced a subdirectoy in their temp path.

This PR fixes the mismatch between the expected and the actual output.


djc (approved):
Thanks! Should we update the trybuild dependency specification to make sure we get the correct version?

Kijewski:
Ah, yes. I updated the dependency specification.

djc (approved):

PR #769, opened 2023-01-27
Apply clippy suggestions for 1.67

djc:
Set an explicit `rust-version` to avoid further clippy-induced MSRV bumps.

### DIFF
--- a/testing/tests/filters.rs
+++ b/testing/tests/filters.rs
@@ -122,8 +122,7 @@ mod filters {
     }
     // for test_nested_filter_ref
     pub fn mytrim(s: &dyn (::std::fmt::Display)) -> ::askama::Result<String> {
-        let s = format!("{}", s);
-        Ok(s.trim().to_owned())
+        Ok(s.to_string().trim().to_owned())
     }
 }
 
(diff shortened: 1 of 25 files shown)


testing/tests/filters.rs
```diff
@@ -122,7 +122,7 @@ mod filters {
     }
     // for test_nested_filter_ref
     pub fn mytrim(s: &dyn (::std::fmt::Display)) -> ::askama::Result<String> {
-        let s = format!("{}", s);
+        let s = format!("{s}");
         Ok(s.trim().to_owned())
```
vallentin, line 126: Alternatively:

```suggestion
        Ok(s.to_string().trim().to_owned())
```

vallentin (approved):
One tiny nit, but otherwise looks fine :)

Kijewski (approved):

testing/tests/filters.rs
Kijewski: Since it's just a test, I have no opinion either way. :) The one line is a bit more idiomatic.

djc: Nice, squashed that into my change!

Kijewski (approved):

PR #771, opened 2023-01-30
derive: Shrink dependency list

Kijewski:
We don't need the more advanced features that `quote` and `syn` offer. This PR opts out of using them.

Until <https://github.com/rust-lang/rust/issues/92565> is resolved, we cannot give a better error location than `#[derive(Template)]` itself. So we don't need to keep track of the span in our `struct CompileError`, and we can remove the dependency of `proc_macro2`.

### DIFF
--- a/askama_derive/Cargo.toml
+++ b/askama_derive/Cargo.toml
@@ -34,8 +34,7 @@ with-warp = []
 mime = "0.3"
 mime_guess = "2"
 nom = "7"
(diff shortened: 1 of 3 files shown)


Kijewski:
I didn't notice that the extra features were pulled in by some other dependency and they were, if fact, needed. :D

The direct dependency to `proc-macro2` could still be removed, but it gets pulled in by `syn` anyway.

PR #772, opened 2023-01-30
derive: refactor parser

Kijewski:
`parser.rs` was a single file containing almost 2000 lines.

This PR split the file into multiple, smaller files. `Expr`, `Node`, and `Target` each get an own file. Each struct gets a `parse()` method that return `Result<Self>`, and every other detail is private to the file.

This PR should make this essential part of Askama more easy to understand, and make future modifications easier.


djc:
Good stuff! Some thoughts on the new structure:

* I'd like to drop the `parse_` prefix on modules within `parser`, since it seems a bit stuttery to have both.
* I think we stuff in `parse_common` could just move into `parser/mod.rs` directly.
* I'd prefer to move the contents of `parse_target` into `parse_node`.
* I'd then also like to move the `Expr` definition into the `expr` module and the `Node` and `Target` definitions into `node`, so that we no longer need a separate `ast` module.

Kijewski:
> * I'd like to drop the parse_ prefix on modules within parser, since it seems a bit stuttery to have both.
> * I think we stuff in parse_common could just move into parser/mod.rs directly.
> * I'd prefer to move the contents of parse_target into parse_node.

Done.

 > * I'd then also like to move the Expr definition into the expr module and the Node and Target definitions into node, so that we no longer need a separate ast module.

I was thinking that this could be the starting point of a more universal interface if someone wants to implement another syntax, e.g. handlebars (#395). But I guess that's not going to happen anytime soon (not by me, anyway :) ). If someone were to implement another syntax, then they can move the structs into their own file.

djc (approved):

djc:
Nice, thanks!

PR #776, opened 2023-02-10
Add derive for impl From<{Template}> for hyper::Body 

valkum:
The `with-hyper` feature will now, when enabled, generate a From implementation allowing `Templates` to be passed to functions with trait bounds for `Into<hyper::Body>`.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -426,12 +426,13 @@ impl<'a> Generator<'a> {
         buf.writeln("}")
     }
 
-    // Implement `From<Template> for hyper::Response<Body>`.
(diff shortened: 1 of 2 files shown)


askama_derive/src/generator.rs
```diff
@@ -447,6 +447,25 @@ impl<'a> Generator<'a> {
         ))?;
         buf.writeln("::askama_hyper::respond(value)")?;
         buf.writeln("}")?;
+        buf.writeln("}")?;
+        buf.writeln(&format!(
```
djc, line 451: At least an empty line here would be nice, maybe add a coment here and before the other impl writing, too?

djc (approved):
Nice! Maybe change the PR to target the 0.11.x branch? Then we can push it out faster.

PR #779, opened 2023-02-13
exchange hyper feature's From<Template> for hyper::Body with TryFrom

valkum:
The previous implementation hid the error returned by render, which made it impossible for callers to react to rendering errors.

While having a simple way to just call `into` would be nice, there is no way to have both implementations without adding any manual specialization because of the blanket impl for `TryFrom`. Thus, the `From` impl is replaced by the `TryFrom` implemenation.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -450,24 +450,22 @@ impl<'a> Generator<'a> {
         buf.writeln("}")?;
         buf.writeln("}")?;
 
-        // From<Template> for hyper::Body
+        // TryFrom<Template> for hyper::Body
         buf.writeln(&format!(
             "{} {{",
             quote!(
-                impl #impl_generics ::core::convert::From<&#ident #orig_ty_generics>
+                impl #impl_generics ::core::convert::TryFrom<&#ident #orig_ty_generics>
                 for ::askama_hyper::hyper::Body
                 #where_clause
             )
         ))?;
+        buf.writeln("type Error = ::askama::Error;")?;
         buf.writeln("#[inline]")?;
         buf.writeln(&format!(
             "{} {{",
-            quote!(fn from(value: &#ident #orig_ty_generics) -> Self)
+            quote!(fn try_from(value: &#ident #orig_ty_generics) -> Result<Self, Self::Error>)
         ))?;
-        buf.writeln(
-            "::askama::Template::render(value).ok().map(Into::into)\
-            .unwrap_or_else(|| ::askama_hyper::hyper::Body::empty())",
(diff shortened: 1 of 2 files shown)


djc:
I'm honestly not sure now if this is a good idea. I guess hyper doesn't have a trait we can plug into that has fallibility plugged in. If that's the case, maybe we should just not implement `From<T: Template> for Body` at all? I don't see much value in having this `TryFrom` impl.

valkum:
I see more value in the `TryFrom impl` than the `From impl` as it gives API caller more control/feedback. 
The caller needs to call `Into::into` in any case, as there is no method with a trait bound for `Into<Body>`. But at the same page, the amount of code the caller needs without this impl is very small. Body implements `From<String> for Body`. So having this convenience is not a huge gain. I guess most will just `render()`, process the result and then call `Body::from` on the `String`.

I'm ok with removing the `From impl` again and closing this PR.

djc:
Okay, so I think what makes sense here is that we have `impl_hyper_into_response()` generate a `TryFrom<T: Template> for Response<Body>`, instead of the `From` impl we have now.

valkum:
I am not sure how handy this would be. The current `impl From<T: Template>  for Response<Body>` covers errors already and returns a 500. Which is good to get going. If we change this to a `TryFrom` we only gain more control over the returned Error response.

I make the case that someone does not use the `From impl` at all if they care about the returned error response. Because in these cases, you are probably at a scale where you want to change headers for the successful response, too. So, you end up using a response `Builder` in these cases nevertheless.

djc (approved):

djc:
Thanks!

FrankReh:
The Askama book does not currently discuss the `hyper` integration, particularly in the "Integrations" chapter. It might be beneficial for users to clarify the distinct roles that the generated `From` and `TryFrom` implementations serve. Looking at the test code didn't immediately help either.

It took me some time to understand the difference between these two implementations and where the difference came from. It wasn't immediately apparent the difference was more than the ability to return an error - one returns a `hyper::Response<hyper::Body>` upon successful rendering, while the other returns a  simpler`hyper::Body` allowing more control and requiring more code by the caller in their route handler.

Also, the existence of the `try_respond` public function added to my confusion. Although it is tasked with supporting the `From` implementation, its name could mistakenly suggest a connection to the `TryFrom` implementation (until one looks at the generated code). It might help to clarify its specific role in the hyper integration as well - maybe it is meant as an example - I think a useful example.

djc:
@FrankReh would you be able to submit a PR clarifying some of the issues you've found? I'd be happy to review!

PR #780, opened 2023-02-20
Replace `toml_edit` with `basic-toml`

Kijewski:
This removes a bunch of transitive dependencies (e.g. "nom8`), and we can use borrowed strings again.


djc (approved):
Awesome, thanks!

Issue #786, opened 2023-03-06
Is write_cond trying to return a size_hint?


couchand:
It looks like `write_cond` is accumulating a contribution to `size_hint`:
https://github.com/djc/askama/blob/417cb924ae1f94d54e2eb13ebc7e9fab91b84588/askama_derive/src/generator.rs#L760

But it seems to be dropped on the floor:
https://github.com/djc/askama/blob/417cb924ae1f94d54e2eb13ebc7e9fab91b84588/askama_derive/src/generator.rs#L633

Should this instead be `size_hint += self.write_cond(ctx, buf, conds, ws)?;`?

couchand:
Same question for `write_match`, `write_loop`, `write_block`, maybe others?  I'm starting to think there's more to accumulating the `size_hint` than it looks like at first glance?

PR #788, opened 2023-03-06
Propogate size_hint from sub-blocks

couchand:
I've opened this based on the assumption that #786 is a simple oversight, and added test cases to catch it.  The values in those assertions were simply blessed on the first run of the test.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -630,16 +630,16 @@ impl<'a> Generator<'a> {
                     self.write_let(buf, ws, var, val)?;
                 }
                 Node::Cond(ref conds, ws) => {
-                    self.write_cond(ctx, buf, conds, ws)?;
+                    size_hint += self.write_cond(ctx, buf, conds, ws)?;
                 }
                 Node::Match(ws1, ref expr, ref arms, ws2) => {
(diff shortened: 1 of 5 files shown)


Kijewski (approved):
LGTM, thanks!

djc:
@Kijewski although I can see how this change seems obviously correct I'd liked to have a bit more time to comment before you merged this. I seem to remember that there is maybe some backstory? It seems likely to me that this was more wired up at some point and there was some reason to undo that? Might make sense to look at some commits that touched the size_hint stuff to see if we can piece together more context.

Kijewski:
Sorry, I'll leave the PR open for longer next time.

E.g. write_loop() was introduced in https://github.com/djc/askama/commit/36b06915177ccc8da1cdefa46fdde7ff44938ef8  (2017-09-07) when there were no size hints. SIZE_HINT was only introduced in https://github.com/djc/askama/commit/9f04869be2afd7d8d1b6fff44eb1657d2121a772 (2019-01-12), and it did not use the size hints of e.g. write_loop() then. So unless I'm using git blame wrong, I think there was nothing more to it, or at least it did not make it into a commit on the current branch.

djc:
Okay, thanks for digging into it, looks like I misremembered!

PR #791, opened 2023-03-07
Add whitespace argument in template derive proc-macro

GuillaumeGomez:
This PR adds the possibility to set `whitespace = "..."` directly into the template derive proc-macro.


djc:
Is this mainly for testing or do you have another use case for it?

GuillaumeGomez:
Convenient for sub-templates mostly, where you want to override what is present in the parent template.

Need to add a `cfg(toml)` to fix the test too.

GuillaumeGomez:
Fixed the CI. :)

djc (approved):

PR #794, opened 2023-03-20
Upgrade to syn 2

djc:


Kijewski (approved):
LGTM, thanks. As I read syn's changelog I assumed that this would be a huge change.

PR #795, opened 2023-03-20
Version bump askama_derive to 0.12.1

djc:
Since compiling two versions of syn is not cool, let's get this out (including a few other recent fixes).

### DIFF
--- a/askama_derive/Cargo.toml
+++ b/askama_derive/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "askama_derive"
-version = "0.12.0"
+version = "0.12.1"
 description = "Procedural macro package for Askama"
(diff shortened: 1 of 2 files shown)


djc:
Turns out the syn upgrade was actually a bit harder than it seemed! But not too many changes needed...

Kijewski:
I guess this is fine. I have no idea how to understand the description of [syn::ExprGroup](https://docs.rs/syn/2.0.3/syn/struct.ExprGroup.html), though.

> An expression contained within invisible delimiters.

Does this mean that `Expr::Group` is generated instead of `Expr::Lit` if there are spaces around the literal? Or what does this mean?

Kijewski (approved):
In the end, I guess it does not matter if I understand the description or not.

The change is obviously fine.

PR #798, opened 2023-03-27
Use `try_reserve()` instead of `with_capacity()`

Kijewski:
`String::with_capacity()` panics if the requested memory could not be allocated. `Template::render()` is a fallible method, and the fact that it can panic is not documented.

This commit uses `String::try_reserve()` instead, so even for an exceedingly large `SIZE_HINT` the method should not panic. In the generated code, `write!()` calls may fail with `Err(std::fmt::Error)` instead.

I do not test if `try_reserve()` returned an error, because the rendering might succeed anyway, if less bytes are written than estimated.


djc (approved):
Seems reasonable.

Issue #588, opened 2021-12-28
Enabling "with-axum" feature on askama causes build errors


qwandor:
If I enable the `with-axum` feature in my `Cargo.toml` like

```toml
askama = {version = "0.11.0", features = ["with-axum"] }
```

then I get lots of build errors like

```
error[E0433]: failed to resolve: could not find `askama_axum` in the list of imported crates
   --> src/controllers/users.rs:134:10
    |
134 | #[derive(Template)]
    |          ^^^^^^^^ could not find `askama_axum` in the list of imported crates
    |
    = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Maybe this is because the `askama_axum` crate has not been published to crates.io?

djc:
When using the integration you should depend on askama_axum directly. It re-exports the contents of the Askama crate.

qwandor:
I see, it still seems to missing on crates.io though. And why does https://djc.github.io/askama/integrations.html say to use the `with-axum` feature in that case?

djc:
Ah sorry, the release is waiting for #583 and the docs are probably outdated. 😔

EdmundsEcho:
In reviewing this issue, it seems that the dependent issues have been resolved.  Is that right?

I ask because I'm having a tough time resolving how a working example from`axum` uses the dependency: `askama = "0.11"`.  However, when I tried to build on the work, I started to get errors requesting `askama_axum`.  The docs indicate a need to use  features `with-axum`; not what the working example uses.

Finally, one of the posts in this issues suggests the documentation is *not* current.  

Thus, the request for clarification and specifically how to use this template engine using `axum`.

Thank you!!

djc:
You should depend on askama_axum only, this should take care of any feature flags that need to be enabled. askama_axum re-exports all the symbols from askama, so you should usually not need to depend on anything else explicitly.

EdmundsEcho:
@djc Thank you for your prompt reply.  This helps streamline and focus my debugging efforts.

And, thank you for your work here.  It's a pleasure to work with these well-thought-out packages.

\- E

EdmundsEcho:
Agree.

Quick follow-up.  I'm getting the following error 

```rust
#[derive(askama_axum::Template)]
    |          ^^^^^^^^^^^^^^^^^^^^^ not found in `askama`
```

... from the following code:
```rust
use askama_axum::Template;
...

#[derive(askama_axum::Template)]
[...]

PR #799, opened 2023-03-27
Updating the axum integration doc (issue #588)

DGotshalk:
When using the askama feature "with-axum" you need to also depend on the "askama_axum" crate directly.

### DIFF
--- a/book/src/integrations.md
+++ b/book/src/integrations.md
@@ -22,7 +22,7 @@ from the Askama test suite for more on how to integrate.
 
 ## Axum integration
 
-Enabling the `with-axum` feature appends an implementation of Axum's
+Enabling the `with-axum` feature and depending on the `askama_axum` crate appends an implementation of Axum's
 `IntoResponse` trait for each template type. This makes it easy to trivially
 return a value of that type in a Axum handler. See
 [the example](https://github.com/djc/askama/blob/main/askama_axum/tests/basic.rs)
--- a/book/src/integrations.md
+++ b/book/src/integrations.md
@@ -22,7 +22,7 @@ from the Askama test suite for more on how to integrate.
 
 ## Axum integration
 
-Enabling the `with-axum` feature and depending on the `askama_axum` crate appends an implementation of Axum's
+Depending on the `askama_axum` crate appends an implementation of Axum's
 `IntoResponse` trait for each template type. This makes it easy to trivially
 return a value of that type in a Axum handler. See
 [the example](https://github.com/djc/askama/blob/main/askama_axum/tests/basic.rs)


book/src/integrations.md
```diff
@@ -22,7 +22,7 @@ from the Askama test suite for more on how to integrate.
 
 ## Axum integration
 
-Enabling the `with-axum` feature appends an implementation of Axum's
+Enabling the `with-axum` feature and depending on the `askama_axum` crate appends an implementation of Axum's
```
djc, line 25: If you depend on askama_axum I don't think you have specifically enable the `with-axum` flag, have you tried that?

DGotshalk: If the flag is not there, it doesn't seem to use the askama_axum crate at all. Whatever axum specific integration you have done with the askama_axum crate isn't seemingly touched without that flag

DGotshalk: Hello,
I just tried it. If I remove the `with-axum` flag, then having the `askama_axum` crate seemingly makes no difference?

DGotshalk: Yea, can confirm that. 

If you have the `with-axum`  feature, you need to have that dependency. However,  askama seems to works with axum even without the `askama_axum` crate? 

I am confused!

djc: I'm sorry, I don't have much time to dig into this. The feature flag is forward to the askama_derive crate, which in its `generator` generates some source code specific to the Axum integration. For many of the integrations, the generated source code depends on the integration crate but that may not be true here? Please have a look at the relevant source code yourself.

DGotshalk: Sure thing, I'll use that information to the best of my abilities.

DGotshalk: I can confirm that when the `with-axum` flag is used, the generated source code depends on the `askama_axum` integration crate. 

I can also confirm, that without the `with-axum` feature flag, the integration code is still generated. I can call the `into_reponse` function that is defined in the `askama_axum` on a Template.

DGotshalk:
I can confirm that when the with-axum flag is used, the generated source code depends on the askama_axum integration crate.

I can also confirm, that without the with-axum feature flag, the integration code is still generated. I can call the into_reponse function that is defined in the askama_axum on a Template.

If you check integration doc files now, I have updated the pull request file to say that you only need to depend on the askama_axum crate, and don't need the feature flag. I don't know if you want the feature flag mentioned in the integration docs.

Also interesting note. I cannot depend on askama_axum only. When using the Templates, it still looks for the askama crate in my dependencies even though it should get re-exported as a part of askama_axum but that may be an issue for another pull request?

PR #805, opened 2023-03-30
Escape HTML faster

Kijewski:
Escaped HTML characters vary in length. So, in order to select the correct replacement two variables need to be loaded: The pointer to the new substring and its length. Because of this the generated code is less dense than it could be.

With this PR instead of selecting the appropriate `&str`, an `&&str` is selected. The former consumes two words while the latter consumes only one. Intuitively one might assume that the double dereference makes the code slower, but the optimized lookup seems to be so much faster, so that the change is worth its weight.

Comparing the result of `cargo bench` (best out of three runs for both):

```text
Old:  [4.3592 µs 4.3675 µs 4.3764 µs]
New:  [3.8691 µs 3.8766 µs 3.8860 µs]
Diff: [-11.24 %  -11.24 %  -12.21 % ]
```

### DIFF
--- a/askama_escape/src/lib.rs
+++ b/askama_escape/src/lib.rs
@@ -113,21 +113,18 @@ impl Escaper for Html {
     {
         let mut last = 0;
         for (index, byte) in string.bytes().enumerate() {
-            macro_rules! go {
-                ($expr:expr) => {{
-                    fmt.write_str(&string[last..index])?;
-                    fmt.write_str($expr)?;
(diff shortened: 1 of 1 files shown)


Kijewski:
The diff in the JSON escaper does not actually change anything, because the replacement is and was always 6 characters long. But I think the new version without macros is easier on the eyes.

djc (approved):
Nice work!

PR #806, opened 2023-03-30
WIP: Implement latex escaper

Kijewski:


askama_escape/src/lib.rs
```diff
@@ -150,6 +150,57 @@ impl Escaper for Text {
...
+    where
+        W: Write,
+    {
+        let mut last = 0;
+        for (index, byte) in string.bytes().enumerate() {
+            const MIN_CHAR: u8 = b'"';
+            const MAX_CHAR: u8 = b'~';
+            const TABLE: [Option<&&str>; (MAX_CHAR - MIN_CHAR + 1) as usize] = {
+                let mut table = [None; (b'~' - b'"' + 1) as usize];
+                table[(b'"' - MIN_CHAR) as usize] = Some(&r#"{\textquotedbl}"#);
```
vallentin, line 166:
```suggestion
            const TABLE: [Option<&str>; (MAX_CHAR - MIN_CHAR + 1) as usize] = {
                let mut table = [None; (b'~' - b'"' + 1) as usize];
                table[(b'"' - MIN_CHAR) as usize] = Some(r#"{\textquotedbl}"#);
```

I assume it might be an artifact from copying and pasting another escaper. But is there a reason for `&&str` instead of just `&str`?

djc:
The question here is, what is the value of maintaining this within this crate?

PR #811, opened 2023-04-03
Allow using arbitrary delimiters

Kijewski:
* Allow delimiters with arbitrary length
* Allow setting expr_start, etc. in Rust source

### DIFF
--- a/askama_derive/src/config.rs
+++ b/askama_derive/src/config.rs
@@ -168,24 +168,29 @@ impl<'a> TryFrom<RawSyntax<'a>> for Syntax<'a> {
             comment_end: raw.comment_end.unwrap_or(default.comment_end),
         };
 
-        if syntax.block_start.len() != 2
-            || syntax.block_end.len() != 2
-            || syntax.expr_start.len() != 2
-            || syntax.expr_end.len() != 2
-            || syntax.comment_start.len() != 2
-            || syntax.comment_end.len() != 2
-        {
-            return Err("length of delimiters must be two".into());
+        for s in [
+            syntax.block_start,
+            syntax.block_end,
+            syntax.expr_start,
+            syntax.expr_end,
+            syntax.comment_start,
+            syntax.comment_end,
+        ] {
+            if s.len() < 2 {
+                return Err("Delimiters must be at least two characters long".into());
+            } else if s.chars().any(|c| c.is_whitespace()) {
+                return Err("Delimiters may not contain white spaces".into());
+            }
         }
 
-        let bs = syntax.block_start.as_bytes()[0];
-        let be = syntax.block_start.as_bytes()[1];
-        let cs = syntax.comment_start.as_bytes()[0];
-        let ce = syntax.comment_start.as_bytes()[1];
-        let es = syntax.expr_start.as_bytes()[0];
-        let ee = syntax.expr_start.as_bytes()[1];
-        if !((bs == cs && bs == es) || (be == ce && be == ee)) {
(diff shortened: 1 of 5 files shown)


Kijewski:
Actually, using a single character as delimiter (even such a character as newline) (w|sh)ould work, but then it would be difficult to extend the language in the future if we have to assume that any and all characters are used in the wild.

djc:
What are you trying to achieve here? Conceptually I don't really like the idea of giving `Config` access to the `TemplateArgs`.

Kijewski:
Well, 1st and foremost I wanted to remove the restriction that you have to use exactly two ASCII characters as delimiters. With this PR you can use any and all unicode characters.  And then I find setting up a config file cumbersome. Right now you can already set the whitespace handling in the code where you derive the template, without needing to resort to using a config file. So I added the option to set the other parts of the configuration, too. I pass &TemplateArgs instead of adding six new parameters, because until Rust adds named parameters having so many parameters is quite error prone, esp. when every parameter has the same type.

Kijewski:
In the end I want to use askama to format a single latex file. Using braces as delimiters does not work very well in this context. Adding a config file would make the code less readable, IMO, because I have no code duplication anyway.

askama_derive/src/config.rs
```diff
@@ -168,24 +168,29 @@ impl<'a> TryFrom<RawSyntax<'a>> for Syntax<'a> {
...
-        let ee = syntax.expr_start.as_bytes()[1];
-        if !((bs == cs && bs == es) || (be == ce && be == ee)) {
-            return Err(format!("bad delimiters block_start: {}, comment_start: {}, expr_start: {}, needs one of the two characters in common", syntax.block_start, syntax.comment_start, syntax.expr_start).into());
+        for (s1, s2) in [
+            (syntax.block_start, syntax.expr_start),
+            (syntax.block_start, syntax.comment_start),
+            (syntax.expr_start, syntax.comment_start),
+        ] {
+            if s1.starts_with(s2) || s2.starts_with(s1) {
+                return Err("A delimiter may not be the prefix of another delimiter".into());
```
djc, line 211: Nit: I think our error message usually don't start with capitals.

djc (commented):
Sorry for the slow feedback.

I think the goal here is good, but I still think that allowing `Config` to inspect the `TemplateArgs` is a layering violation. It should really be set up the other way around.

askama_derive/src/generator.rs
```diff
@@ -163,37 +180,13 @@ impl TemplateArgs {
...
-                    return Err("whitespace value must be string literal".into());
-                }
-            } else {
+            } else if !capture_str_lit!(
+               "escape": args.escaping,
+               "ext": args.ext,
+               "syntax": args.syntax,
+               "config": args.config_path,
+               "whitespace": args.whitespace,
+            ) {
```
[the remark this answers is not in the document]
Kijewski: Yes, good idea!

askama_derive/src/config.rs
Kijewski: You're right. Will change.

Kijewski:
> Sorry for the slow feedback.

No problem at all!

What do you think about adding a new syntax config struct `{whitespace, block_start, ..}` that I would pass around? Instead of `template_whitespace` in the current version, or the whole `TemplateArgs` I in my PR proposed?

djc:
> What do you think about adding a new syntax config struct `{whitespace, block_start, ..}` that I would pass around? Instead of `template_whitespace` in the current version, or the whole `TemplateArgs` I in my PR proposed?

Sounds like a decent path forward? I'd have to see it in more detail to be sure, though.

Kijewski:
PR needs to be redone entirely, because the codebase changed entirely.

PR #813, opened 2023-04-17
Update to rocket 0.5-rc.3

Vypo:
Splits the lifetime for the derived impl into two, and adds a struct deriving `Responder` to hopefully catch a similar change in the future.


djc (approved):

djc:
Thanks!

PR #822, opened 2023-06-10
Allow macros to be defined and called without arguments

mataha:
This PR proposes a shorthand for defining and calling macros when using them as a reusable substitute for variables assigned complex values (e.g. string literals with or without newline escapes). The use-case is formatting - from my experience it's easier to visually parse a `macro`-`endmacro` block than a multiline variable assignment.

I might be alone in that regard, though - opinions?


djc (approved):
Thanks!

PR #824, opened 2023-06-18
Feature: Block fragments

wrapperup:
Since https://github.com/djc/askama/pull/568 hasn't had an update in a long time, and it is a nice feature, this is a new PR to implement the proposed changes from that discussion.

This adds block fragments/partials as part of the Template derive. Here's an example of how to use it:

```rust
#[derive(Template)]
#[template(path = "...", block = "my_block")]
struct MyBlockFragment<'a> {
    name: &'a str,
}
```

This also properly supports inheritance, nesting, and the `super()` macro. It does this by abstracting `buf_writable` in the generator to add the ability to discard output, except for the block that needs to be rendered.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::collections::hash_map::{Entry, HashMap};
+use std::ops::Deref;
 use std::path::Path;
 use std::rc::Rc;
 use std::{cmp, hash, mem, str};
@@ -34,8 +35,8 @@ pub(crate) struct Generator<'a> {
     skip_ws: WhitespaceHandling,
     // If currently in a block, this will contain the name of a potential parent block
     super_block: Option<(&'a str, usize)>,
-    // buffer for writable
-    buf_writable: Vec<Writable<'a>>,
+    // Buffer for writable
+    buf_writable: WritableBuffer<'a>,
     // Counter for write! hash named arguments
     named: usize,
 }
@@ -55,7 +56,10 @@ impl<'a> Generator<'a> {
             next_ws: None,
             skip_ws: WhitespaceHandling::Preserve,
             super_block: None,
-            buf_writable: vec![],
+            buf_writable: WritableBuffer {
+                discard: input.block.is_some(),
+                ..Default::default()
+            },
             named: 0,
         }
     }
@@ -891,6 +895,13 @@ impl<'a> Generator<'a> {
         name: Option<&'a str>,
         outer: Ws,
     ) -> Result<usize, CompileError> {
+        let block_fragment_write = self.input.block == name && self.buf_writable.discard;
+
+        // Allow writing to the buffer if we're in the block fragment
+        if block_fragment_write {
+            self.buf_writable.discard = false;
+        }
+
         // Flush preceding whitespace according to the outer WS spec
         self.flush_ws(outer);
 
@@ -939,6 +950,12 @@ impl<'a> Generator<'a> {
         // succeeding whitespace according to the outer WS spec
         self.super_block = prev_block;
         self.prepare_ws(outer);
+
+        // Restore the original buffer discarding state
+        if block_fragment_write {
+            self.buf_writable.discard = true;
+        }
+
         Ok(size_hint)
     }
 
@@ -2053,6 +2070,37 @@ enum AstLevel {
 enum DisplayWrap {
     Wrapped,
     Unwrapped,
+}
+
+#[derive(Default, Debug)]
+struct WritableBuffer<'a> {
+    buf: Vec<Writable<'a>>,
+    discard: bool,
+}
+
+impl<'a> WritableBuffer<'a> {
+    fn push(&mut self, writable: Writable<'a>) {
+        if !self.discard {
+            self.buf.push(writable);
+        }
+    }
+}
+
+impl<'a> IntoIterator for WritableBuffer<'a> {
+    type Item = Writable<'a>;
+    type IntoIter = <Vec<Writable<'a>> as IntoIterator>::IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.buf.into_iter()
+    }
+}
+
+impl<'a> Deref for WritableBuffer<'a> {
+    type Target = [Writable<'a>];
+
+    fn deref(&self) -> &Self::Target {
+        &self.buf[..]
+    }
 }
 
 #[derive(Debug)]
--- a/askama_derive/src/lib.rs
+++ b/askama_derive/src/lib.rs
@@ -66,7 +66,15 @@ pub(crate) fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileEr
 
     let ctx = &contexts[&input.path];
     let heritage = if !ctx.blocks.is_empty() || ctx.extends.is_some() {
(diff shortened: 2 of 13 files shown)


askama_derive/src/generator.rs
```diff
@@ -242,6 +259,52 @@ fn find_used_templates(
...
 }
 
+#[derive(Default)]
+struct WritableBuffer<'a> {
+    inner: Vec<Writable<'a>>,
+    discard: bool,
+}
+
+impl<'a> WritableBuffer<'a> {
+    fn new(discard: bool) -> Self {
```
couchand, line 269: Two named constructors rather than a `bool` parameter, please.

book/src/creating_templates.md
```diff
@@ -75,6 +75,14 @@ recognized:
   #[template(path = "hello.html", print = "all")]
   struct HelloTemplate<'a> { ... }
   ```
+* `block` (as `block = "block"`): renders a block by itself, useful for partial
```
couchand, line 78: This would be clearer if the name of the example block were something other than "block".

book/src/template_syntax.md
```diff
@@ -183,6 +183,21 @@ blocks from the base template with those from the child template. Inside
 a block in a child template, the `super()` macro can be called to render
 the parent block's contents.
 
+### Block Fragments
```
couchand, line 186: I think it makes sense to mention the feature here since it relates to these blocks, but I feel like it makes sense to put the bulk of the documentation for this in `book/src/creating_templates.md` rather than here.

couchand (commented):
I'm really impressed at how effective the strategy of abstracting the `WritableBuffer` is.  Very nice.

Personally this feature seems a bit odd to me.  It feels like its overloading the block inheritance mechanism to do something very different that there's already other ways to do.  I did re-read the thread from #568.  I wonder if perhaps a more complete motivating example would be helpful, in the tests or the docs or just PR comments?

wrapperup:
> Personally this feature seems a bit odd to me. It feels like its overloading the block inheritance mechanism to do something very different that there's already other ways to do.

I couldn't think of a cleaner way here. Perhaps adding another node to do it, but `block` ended up being suggested as the preferred way in that thread.

> I wonder if perhaps a more complete motivating example would be helpful, in the tests or the docs or just PR comments?

It's very much a convenience feature. There are already ways to do achieve the same result (macros, importing other templates). When using a library like `htmx` that send down small sections of html to update the page, it's pretty convenient.

It's part of this whole "locality of behavior" idea, where you keep things as close together as possible (ideally, in the same file). htmx, Alpine.js, Tailwindcss, and all of those that are HTML first trend towards this kind of design. This is also a pretty good read: https://htmx.org/essays/locality-of-behaviour/

testing/tests/block_fragments.rs
```diff
@@ -0,0 +1,88 @@
+use askama::Template;
+
+// fragment-nested-block.html
+// fragment-nested-super.html
+// fragment-simple.html
+// fragment-super.html
+
+#[derive(Template)]
```
djc, line 8: Keep the types close to the tests please, instead of keeping types together.

askama_derive/src/generator.rs
```diff
@@ -309,7 +383,14 @@ impl<'a> Generator<'a> {
         heritage: Option<&'n Heritage<'_>>,
         locals: MapChain<'n, &'n str, LocalMeta>,
         whitespace: WhitespaceHandling,
+        discard_output: bool,
     ) -> Generator<'n> {
+        let buf_writable = if discard_output {
+            WritableBuffer::new_discarding()
+        } else {
+            WritableBuffer::new()
+        };
```
djc, line 392: Please use a `match` expression for this, and inline it into the initializer expression directly instead of binding here. Let's ditch the constructors for `WritableBuffer` and initialize its fields directly here (and rename the argument to the `Generator::new()` constructor as just `discard` to match).

askama_derive/src/generator.rs
```diff
@@ -241,6 +258,63 @@ fn find_used_templates(
     Ok(())
 }
 
+#[derive(Default)]
```
djc, line 261: This type should go towards the bottom, just above the `Writable` type definition. I think we should ditch pretty much all the methods and trait impls for it other than the `push()` method, in favor of direct access to its fields.

djc (commented):
The direction here seems like it offers a decent amount of functionality especially relative to the limited complexity cost, so it seems reasonable to me. @vallentin / @Kijewski any thoughts?

wrapperup:
The parser crate changes had no effect on this PR. Could I get eyes on these changes again?

wsantos:
@djc what do we need to merge this? I'm considering using it but since my site is going to be built in htmx this is a requirement, thanks in advance.

nathanaelhuffman:
I'm also considering using this with htmx and would love to see this land.

pbouzakis:
Same here; looking to leverage this in htmx project. Currently, I'm using minijinja and render_block, which is working nicely. Would love to move to Askama if I can, but need this feature.

teenjuna:
Hey @Kijewski, sorry to bother, but it looks like @djc is looking for your opinion on this. Can you please take a look? 🙏

djc:
(No is also an okay answer, but would like to have that be made explicit!)

Kijewski:
Sorry, I won't find the time to wrap my head around this PR for the foreseeable time. At a very (_very_) quick glance the code looks fine, and the feature seems useful to me.

James-Rhodes:
Hi all is there any updates on this feature? Very much looking forward to seeing it land :)

djc:
This is waiting for another round of review from me now that we have some positive acknowledgement from @Kijewski (thanks!) -- I'm having a very busy time so it may take me a bit.

(If this is important to you, maybe consider [sponsoring](https://github.com/sponsors/djc) my work on Askama and other projects.)

GuillaumeGomez (approved):

GuillaumeGomez:
The feature is pretty nice and code looks good to me! Let's wait for @djc to have time to come back to it.

askama_derive/src/lib.rs
```diff
@@ -62,8 +70,14 @@ pub(crate) fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileEr
...
     }
 
-    let code = Generator::new(&input, &contexts, heritage.as_ref(), MapChain::default())
-        .build(&contexts[input.path.as_path()])?;
+    let code = Generator::new(
+        &input,
+        &contexts,
+        heritage.as_ref(),
+        MapChain::default(),
+        input.block.is_some(),
```
djc, line 78: Given that we're already passing `input` to `Generator::new()`, there's no reason to pass `input.block.is_some()` separately -- it should derive the value for `WritableBuffer::discard`.

djc:
Please just rebase and squash all your changes -- right now "this branch cannot be rebased due to conflicts".

djc:
(Failing CI now.)

Kijewski:
The error is unrelated to this PR, and should be fixed in the current main branch. Can you please rebase, @wrapperup?

wrapperup:
> I think the error is unrelated to this PR, and it should be fixed in the current main branch. Can you please rebase, @wrapperup?

Yep! Was just waiting on it to be fixed in main.

djc:
Thanks, and sorry for the long delays in getting this merged!

PR #827, opened 2023-06-20
Added example on how to use super() in the docs

### LINKED ISSUE #825: How to use super()?
I'm trying to get the `super()` macro to work.

rustc 1.70.0 (90c541806 2023-05-31)

Cargo.toml
```toml
[dependencies]
askama = {version = "0.11.1"}
```

page.html
```html
{% extends "base.html" %}

{% block content %}
child
{{super()}}
{% endblock %}
```

base.html
```html
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="UTF-8">
	<meta http-equiv="X-UA-Compatible" content="IE=edge">
	<meta name="viewport" content="width=device-width, initial-scale=1.0">
	<title>Document</title>
</head>
<body>
{% block content %}
parent
{% endblock %}
</body>
</html>
```

main.rs
```rust
use askama::Template;

#[derive(Template)]
#[template(path="page.html")]
struct Page {}

fn main() {
	println!("{}", Page{}.render().unwrap());
}
```

The error:
```
error[E0609]: no field `super` on type `&Page`
 --> src\main.rs:3:10
  |
3 | #[derive(Template)]
  |          ^^^^^^^^ unknown field
  |
```

I've tried with `{ super() }`, `{{ super }}`and `{% super() %}` but those don't work either. I haven't managed to find a working example on the internet...

reportingissue:
Resolves #825 

The default jinja syntax for `super()` doesn't require `call` and someone looking through the docs quickly might not realize that its the only way to call it, as macros are a slightly more advanced topic, more likely to be skipped over.

### DIFF
--- a/book/src/template_syntax.md
+++ b/book/src/template_syntax.md
@@ -141,7 +141,7 @@ A base template defines **blocks** that child templates can override.
   </head>
   <body>
     <div id="content">
-      {% block content %}{% endblock %}
+      {% block content %}<p>Placeholder content</p>{% endblock %}
     </div>
   </body>
 </html>
@@ -171,6 +171,7 @@ Here's an example child template:
 {% block content %}
   <h1>Index</h1>
   <p>Hello, world!</p>
+  {% call super() %}
 {% endblock %}
 ```
 


djc (approved):
Thanks!

Issue #826, opened 2023-06-20
Compile times scale with number of templates


wrapperup:
Hi, I'm enjoying askama, it's fantastic. I noticed that compile times for templates grow considerably if you have many templates, which I don't think is too surprising considering it's a proc-macro.

I thought of having templates cached in the user's project (under a `.askama` directory) without using a build script. But then I saw https://github.com/djc/askama/pull/689, but it doesn't seem to have gone anywhere. Has there been any plans or additional thoughts on that to continue on this work?

IMO, it would be great if it could be done purely in the proc-macro. In my limited testing, just skipping the compile step already sped things up considerably, so it seems like the overhead of invoking and parsing a proc-macro wasn't big enough to notice (testing with about 100 templates). I saw there was some unstable Rust features for tracking files in proc-macros, which would be useful for this, but who knows how long that would take to stabilize.

djc:
> The main issue is when making incremental changes, askama will recompile all templates, even if they didn't change. This hurts iteration times a bit.

I find this surprising. Have you verified this? Do you understand why this is the case?

The work in #689 hasn't gone anywhere as far as I know. I still think the code generation at test time would be the optimal approach, but if you have concrete benchmark numbers to demonstrate the benefits of a caching approach I'd be open to considering that as well.

wrapperup:
So, my test implements 50 templates templates in a single module.
```rust
#[derive(Template)]
#[template(path = "index_001.html")]
struct Index001 {
   test: String,
}

...

#[derive(Template)]
#[template(path = "index_050.html")]
struct Index050 {
   test: String,
}
```
Making a change to one or more of these templates causes the compile time to be about ~3.10s.
I ran self profile to get a flamegraph, and curiously the proc-macro stage was the longest.

![image](https://github.com/djc/askama/assets/7478134/ab68c434-9c2f-406d-aa81-1c914a04c1d8)

I also stubbed out `include_bytes!` to see if maybe there was some side-effects with file tracking that may have caused all the macros to run again, but that didn't seem to help much. (I got similar results to above)

I made a very quick-and-dirty prototype (hashes entire proc-macro AST, but not template itself), and the build time dropped to 0.43s for one template, and about 1.78s for compiling all the templates.

Flamegraph for one template change with the cache (much better!):
![image](https://github.com/djc/askama/assets/7478134/f2632028-9be2-49bd-b6c2-d47f479662fc)

and all:
[...]

PR #828, opened 2023-06-22
Feature: Cache compiled templates

wrapperup:
More info: https://github.com/djc/askama/issues/826

If the "cache" feature is enabled, then the compiled template sources will be cached as a build artifact. This is done purely in the derive macro.

Assuming you are alright with how it works now, then this works pretty well as is. Improves iteration times quite a bit!


askama_derive/src/generator.rs
```diff
@@ -73,9 +90,73 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
...
+
     Ok(code)
 }
 
+#[cfg(feature = "cache")]
+fn find_cached_template(digest: &str) -> Option<String> {
+    let root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    let cache_dir = root.join(".askama");
+    let path = cache_dir.join(digest);
+
```
djc, line 105: I don't think we should `unwrap()` here. We can just return `None` instead so that everything will still work. Can write the tail expression something like `path.exists().then(|| fs::read_to_string(path).ok()).flatten()`.

askama_derive/src/generator.rs
```diff
@@ -73,9 +90,73 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
...
+    #[cfg(feature = "cache")]
+    cache_template(&digest, &code).unwrap();
+
     Ok(code)
 }
 
+#[cfg(feature = "cache")]
+fn find_cached_template(digest: &str) -> Option<String> {
+    let root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
```
djc, line 102: Use `CARGO_TARGET_DIR` instead.

djc (commented):
Looking good!

askama_derive/src/generator.rs
wrapperup: `CARGO_TARGET_DIR` isn't intended to be used this way unfortunately. And on Windows, this isn't set during compilation. I think instead, we should re-export `OUT_DIR` in a build script.

wrapperup: I just went with `OUT_DIR` since I can guarantee that it's set.

wrapperup:
With those changes, I would be comfortable making `cache` a default feature. What do you think?

djc:
Why do we need the build script? Not a big fan of build scripts.

wrapperup:
> Why do we need the build script? Not a big fan of build scripts.

`CARGO_TARGET_DIR` is not available to crates, or not intended to be anyway. On Windows it wasn't available.

Instead, I'm using `OUT_DIR`, but it's only defined if a build script exists. The build script doesn't do anything else, and it only reruns if the `OUT_DIR` env changes or if the build script changes (which even that part is probably unnecessary).

djc:
How about `CARGO_TARGET_TMPDIR` instead?

couchand (commented):
I love this.  It's going to make my life much easier, thanks!

askama_derive/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    // This build script only exists so that OUT_DIR is set.
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-env-changed=OUT_DIR");
```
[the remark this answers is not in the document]
wrapperup: `cargo:rerun-if-env-changed` reruns only if the environment variable changes (ie, switch targets). It doesn't care if something inside that directory changed. Even with that, I'm not sure having anything here is actually necessary, would need some testing.

askama_derive/src/generator.rs
```diff
@@ -73,9 +93,66 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
...
+
+    path.exists()
+        .then(|| std::fs::read_to_string(path).ok())
+        .flatten()
+}
+
+#[cfg(feature = "cache")]
+fn cache_template(digest: &str, source: &str) -> Result<(), CompileError> {
+    let root =
+        PathBuf::from(option_env!("OUT_DIR").ok_or(CompileError::from("OUT_DIR is not defined"))?);
```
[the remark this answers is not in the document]
wrapperup: Makes sense. There is potential that someone may be using askama without cargo, but that is kind of at their peril anyway, and it's probably better to have the compile error appear there instead of as a compile error.

askama_derive/src/generator.rs
```diff
@@ -73,9 +93,66 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
...
+
+    let path = root.join(digest);
+
+    std::fs::write(path, source)
+        .map_err(|err| CompileError::from(format!("failed to write template cache: {}", err)))
+}
+
+#[cfg(feature = "cache")]
+fn hash_features(hasher: &mut blake3::Hasher) {
+    #[cfg(feature = "humansize")]
```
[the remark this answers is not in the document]
wrapperup: I'm not 100% sure what you mean. If you mean potential here to create collision hash here with two similar features ("x" vs "with-x), it's highly unlikely.

askama_derive/build.rs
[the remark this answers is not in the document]
wrapperup: Seems like none of it is necessary, so I'll just remove the two lines here.

askama_derive/src/generator.rs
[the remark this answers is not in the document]
couchand: Yes, the whole list of features is right here, so it "shouldn't" be a problem.  But it's already long enough that verifying that it's not possible is a fair amount of effort.  Sticking in a byte that is not a valid byte in a feature name could be prudent?

askama_derive/build.rs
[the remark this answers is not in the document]
couchand: Oh! I missed the `env` there!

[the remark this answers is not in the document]
couchand: Oh you do still want the first one.

> If the build script inherently does not need to re-run under any circumstance, then emitting cargo:rerun-if-changed=build.rs is a simple way to prevent it from being re-run (otherwise, the default if no rerun-if instructions are emitted is to scan the entire package directory for changes). 

https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed

askama_derive/src/generator.rs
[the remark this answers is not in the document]
couchand: Wow, very niche case but perhaps worth a callout in the docs?  As long as they set some valid, stable directory it will all work...

askama_derive/src/generator.rs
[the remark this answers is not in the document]
wrapperup: Ah I see what you mean. Could stick "feature-" in front of all of these.

askama_derive/build.rs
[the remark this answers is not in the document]
wrapperup: Ah nice. Added it back.

djc:
> Instead, I'm using `OUT_DIR`, which is only defined if a build script exists. The build script itself doesn't do anything, except signal to Cargo to make a directory and set `OUT_DIR`.

That seems like a pretty ugly solution to me. Maybe have a look at how something like sqlx does this stuff?

askama_derive/src/generator.rs
```diff
@@ -73,9 +93,61 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
...
+}
+
+#[cfg(feature = "cache")]
+fn cache_template(digest: &str, source: &str) -> Result<(), CompileError> {
+    let root = PathBuf::from(env!("OUT_DIR"));
+
+    let path = root.join(digest);
+
+    std::fs::write(path, source)
+        .map_err(|err| CompileError::from(format!("failed to write template cache: {}", err)))
```
Kijewski, line 120: Does this function need to be fallible? If the file could not be written, then we could simply return the source like we'd do without a cache.

askama_derive/Cargo.toml
```diff
@@ -31,6 +32,7 @@ with-tide = []
 with-warp = []
 
 [dependencies]
+blake3 = "1.4.0"
```
Kijewski, line 35: Could you benchmark your PR with `blake3` and `ahash`? I guess `blake3` is faster, but `ahash` is much more likely to already be a dependency in your `cargo tree`. It's a non-optional dependency of `actix-web` and `sqlx-core`.

wrapperup:
> > Instead, I'm using `OUT_DIR`, which is only defined if a build script exists. The build script itself doesn't do anything, except signal to Cargo to make a directory and set `OUT_DIR`.
> 
> That seems like a pretty ugly solution to me. Maybe have a look at how something like sqlx does this stuff?

We could go back to `target/askama`. Seems like sqlx will prioritize `SQLX_TMP`, `CARGO_TARGET_DIR`, and then just `target/sqlx`. I guess if it's good enough for them, we can just do that.

askama_derive/Cargo.toml
wrapperup: I think `ahash` will be faster actually, but I think using blake3 or any cryptographic hash function is preferable since we'd like a very stable, collision-free hash for the file names since we can't deal with collisions. `ahash` and others like it are meant more for use in a hashmap, but maybe that's fine?

askama_derive/src/generator.rs
wrapperup: I think it's nice to tell the user that we can't write into their cache directory if they have caching enabled. I'd be fine with making it a warning, but we can't do that yet in stable Rust.

askama_derive/Cargo.toml
djc: Yeah, I think blake3 is the better choice for this kind of thing.

askama_derive/src/generator.rs
```diff
@@ -73,9 +93,84 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
...
+
     Ok(code)
 }
 
+#[cfg(feature = "cache")]
+fn get_or_create_cache_dir() -> Result<PathBuf, CompileError> {
+    use std::{env, fs};
+
+    let dir = if let Ok(dir) = env::var("ASKAMA_CACHE_DIR") {
+        PathBuf::from(dir)
```
djc, line 108: Let's early return here.

askama_derive/src/generator.rs
```diff
@@ -73,9 +93,84 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
...
+    } else {
+        let root = if let Ok(dir) = env::var("CARGO_TARGET_DIR") {
+            PathBuf::from(dir)
+        } else {
+            PathBuf::from("./target")
```
djc, line 113: Does this do the right thing in the face of workspaces? (That is, store `askama` in `<workspace>/target` rather than `<workspace>/<crate>/target`?)

askama_derive/src/generator.rs
```diff
@@ -73,9 +93,84 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, CompileError> {
...
     }
+
+    #[cfg(feature = "cache")]
+    cache_template(&digest, &code)?;
+
     Ok(code)
 }
 
+#[cfg(feature = "cache")]
+fn get_or_create_cache_dir() -> Result<PathBuf, CompileError> {
```
djc, line 104: I'd prefer to just call this `cache_dir()`.

wrapperup:
One glaring issue: this doesn't know if an imported or extended template changed unless we parse the templates (at which point, all the speed benefits will go away).

Maybe someone has a clever-er idea on solving this, but my idea involves making a data format for the hashed files. It will include the files it depends on, and their hashes, and the generated code like before.

Kijewski:
Maybe a `output.d` that contains a list of all used dependencies:

`/home/user/project/template1.html\0/home/user/project/template2.html\/home/user/project/template3.html`

(\n is a valid file name character at least in Linux, but \0 should not be valid anywhere.)
If any of the files has a newer modification date as the cached source, then recompile, otherwise assume that the content is still current.

djc:
Did you ever benchmark which parts of the process take time, anyway? If parsing is slow, maybe there's low-hanging fruit in optimizing it instead of working around the crappy performance with caching?

With a file format where we're going to compare file hashes for included files, this gains like an order of magnitude of complexity, so I'm less sure it's worth it.

wrapperup:
So I made a small mod to `askama_derive` so that I could put it into a binary, and ran it through Superlumial to try and get a much more useful profile than the one in the previous discussion, and found a few things:

1. The template gets parsed twice unnecessarily(?). Once in `find_used_templates`, and again later when we actually read all the tokens. That would speed things up a bit.
2. nom is very slow in debug. Building `askama_derive` in release mode improves the performance a lot, competitive with caching (without cache was about ~500ms, cache was ~300ms, both in release mode). A very actionable thing users can do now is add the snippet below into Cargo.toml for a much faster compile time, and maybe we could make this the default in Askama?
```rust
[profile.dev.package.askama_derive]
opt-level = 3
```

Kijewski:
@wrapperup, could you please have a look at #832? This should fix the 1st problem and templates are only parsed once.

djc:
@wrapperup we merged #833 to avoid reparsing templates, I'd be curious how that improves your experience. Can you share detailed benchmarking information with us somehow? Would be great to get more insight into what the bottlenecks are in your template compilation pipeline.

wrapperup:
> @wrapperup we merged #833 to avoid reparsing templates, I'd be curious how that improves your experience. Can you share detailed benchmarking information with us somehow? Would be great to get more insight into what the bottlenecks are in your template compilation pipeline.

Yep, though it is in Windows. Would also be great if the bulk of `askama_derive` could be split out into a seperate, normal lib crate (perhaps `askama_derive_impl`) using `proc_macro2::TokenStream` in place of `proc_macro::TokenStream`. This will make it easier to embed the parser into a binary so that it can be ran through a profiler.

Regardless, I basically did that to profile it. My test is parsing 100 templates, 10 runs averaged. Results are (WSL):

| Target | Time Avg (Debug) | Time Avg (Release) |
|--------|------------------|--------------------|
| 0.12   | 3.932s           | 182ms              |
| main   | 2.086s           | 126ms              |

Pretty big improvements, debug is almost twice as fast. Thanks @djc and @Kijewski !

For a detailed profile, I made traces in Superlumial and flamegraph, though I'm not too familiar with flamegraph so it may not be as useful as the Superlumial ones.

https://github.com/wrapperup/askama_profiling

It seems to spend a lot of time here on line 131 in `parser\mod.rs` in debug.

image

djc:
Can we use these Superluminal traces without a subscription? I haven't used the tool before, but this seems like a nice opportunity to try it.

wrapperup:
> Can we use these Superluminal traces without a subscription? I haven't used the tool before, but this seems like a nice opportunity to try it.

There is a free trial, but beyond that I'm not sure. I could try profiling with something else if you have an idea.

djc:
@wrapperup on the main branch there's now a separate askama_parser crate with just the parser.

djc:
Going to close this for now, as it seems this adds more complexity than what is needed.

PR #829, opened 2023-06-24
Template node structs & whitespace semantics

couchand:
This is the first set of changes extracted from #782.

These changes focus on clarifying the semantics of the whitespace of each `Node`.  This PR refrains from abstracting out the common whitespace-handling patterns that emerge here, leaving that for a follow-on PR.

We make the following changes here:

- Adding `Node` variant parsing tests
- (Mostly) consistently using one struct per `Node` variant
- Clarifying the semantics of each and every `Ws` value (though since the whitespace patterns aren't yet abstracted this clarification is still incomplete)

There should be no visible effects of these changes, they are purely distilling towards greater insight.  As described in the comments to #782, the general goal is to differentiate between the "inner" whitespace (within nested blocks) and the "outer" whitespace (around every node except lit).

Outer whitespace is the first field of each non-`Lit` variant of `Node`.  Inner whitespace is on each nested-block-type node, stored alongside the `Vec<Node>` value that represents the nested block.  For instance, for the template

```jinja
{%- if test +%}FooBar{%+ endif -%}
```

Would result in the Rust value

```rust
Node::Cond(
    Ws(Some(Whitespace::Suppress), Some(Whitespace::Suppress)),
    vec![
        Cond {
            test: CondTest {
                target: None,
                expr: Expr::Var("test"),
            },
            block: vec![Node::Lit("FooBar")],
            ws: Ws(Some(Whitespace::Preserve), Some(Whitespace::Preserve)),
        }
    ]
)
```

Thanks for your review, your time and effort are appreciated!

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -653,8 +653,10 @@ impl<'a> Generator<'a> {
                 Node::Let(ws, ref var, ref val) => {
                     self.write_let(buf, ws, var, val)?;
                 }
-                Node::Cond(ref conds, ws) => {
-                    size_hint += self.write_cond(ctx, buf, conds, ws)?;
+                Node::Cond(ws, ref conds) => {
+                    self.flush_ws(ws);
+                    size_hint += self.write_cond(ctx, buf, conds)?;
+                    self.prepare_ws(ws);
                 }
                 Node::Match(ws, Match { ref expr, ref arms }) => {
                     self.flush_ws(ws);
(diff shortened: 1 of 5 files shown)


askama_derive/src/generator.rs
```diff
@@ -665,7 +665,14 @@ impl<'a> Generator<'a> {
                 Node::Include(ws, path) => {
                     size_hint += self.handle_include(ctx, buf, ws, path)?;
                 }
-                Node::Call(ws, scope, name, ref args) => {
+                Node::Call(
```
djc: If we're going to do this, I think we should just pass the `&Call` directly to `write_call()`... Not much point in destructing it here. (Same for other node types.)

djc:
#834 extracts a parser crate and improves the parser API a bunch.

PR #830, opened 2023-06-26
book: Add Hyper integration section to Integrations chapter

FrankReh:
Adds an explanation of the `with-hyper` feature to the 'Integrations' chapter. The new section explains the purpose and usage of the `From` and `TryFrom` implementations provided by the feature, for each generated template, as well as how the two implementations differ.

It references the integration's example test as the other integration sections do, but also references the integration's source code for easier navigation for the reader who wants to dig in a little further.

### DIFF
--- a/book/src/integrations.md
+++ b/book/src/integrations.md
@@ -44,6 +44,36 @@ In case of a run-time error occurring during templating, the response will be of
 signature, with a status code of `500 Internal Server Error`, mime `*/*`, and an empty `Body`.
 This preserves the response chain if any custom error handling needs to occur.
 
+## Hyper integration
+
+Enabling the `with-hyper` feature appends `From<Template> for hyper::Response<hyper::Body>` and
+`TryFrom<Template> for hyper::Body` implementations for each template type. These
+provide the ability for Hyper apps to build a response directly from
+a template, or to create a templated hyper body to be used further in building a
+`Response`. See [the example](https://github.com/djc/askama/blob/main/askama_hyper/tests/basic.rs)
+from the Askama test suite for more on how to integrate.
+
+When using `Template::from()` (or `Template::into()`),
+the returned type is `hyper::Response<hyper::Body>`.
+On the success of run-time rendering, the response consists of the status code `200`, the MIME type associated with the
+template, and the body, created from the rendered String.
+On the failure of run-time rendering, the response consists of the status code `500`, and an empty
+body. The returned value can be seen being created in the integration's public `respond` function
+in the
+[integration's source](https://github.com/djc/askama/blob/main/askama_hyper/src/lib.rs).
+
+When using `Template::try_from()` (or `Template::try_into()`),
+the returned type is `Result<hyper::Body, askama::Error>`.
+The generated code for the template does much less in this case, returning either a `hyper::Body`
+when rendering was successful or the `askama::Error` when it was not. This is meant to allow the
+caller more flexibility in how a response is built and in actions taken when rendering fails.
+
+The
+[integration's source](https://github.com/djc/askama/blob/main/askama_hyper/src/lib.rs)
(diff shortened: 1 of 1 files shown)


FrankReh:
In reference  to
- #779

FrankReh:
I should perhaps compare the askama_hyper test with other integration tests but I wanted to try to get this kind of change to the book in first, if you agree. In addition to this change to the book, I think a few well placed comments in the tests/basic.rs could go a long way in helping a new reader understand what's going on in the tests, and specifically why there are essentially two tests.

I think the biggest help I could have gotten by looking at the test code yesterday would have been a comment about how the error from the second test could be picked up and used for a server's logging or something like that where there is no failing rendering apparent when using the first form being tested. 

So the first form is quick and easy to throw into new code but papers over potential problems that can only be encountered during runtime. The second allows detecting runtime problems with template rendering while still helping to build hyper responses, but just less help, leaving more to the caller to have to do.

If someone is using 'hyper', they are probably aware of all this. So how much "help" should come with this feature is debatable.

book/src/integrations.md
```diff
@@ -44,6 +44,36 @@ In case of a run-time error occurring during templating, the response will be of
...
+the returned type is `Result<hyper::Body, askama::Error>`.
+The generated code for the template does much less in this case, returning either a `hyper::Body`
+when rendering was successful or the `askama::Error` when it was not. This is meant to allow the
+caller more flexibility in how a response is built and in actions taken when rendering fails.
+
+The
+[integration's source](https://github.com/djc/askama/blob/main/askama_hyper/src/lib.rs)
+also provides a public `try_respond` function that some may find useful to work with or work from.
+It should not be confused with the `TryFrom` generated impl however; the generated impl
+does not use either of the functions found in that source.
```
Kijewski, line 75: What do you mean with this paragraph? The generated TryFrom implementation does indeed use this function: https://github.com/djc/askama/blob/1856c4b57ca7ab706552e2c47f550c86e8999e7c/askama_derive/src/generator.rs#L461

FrankReh: This what what I thought the generated `TryFrom` function was. Am I misinterpreting something?
```rust
impl < 'a, N > ::core::convert::TryFrom <&HelloTemplate<'a, N>> for ::askama_hyper::hyper::Body
where N: std::fmt::Display,
{
    type Error = ::askama::Error;
    #[inline]
    fn try_from(value : & HelloTemplate < 'a, N >) -> Result < Self, Self :: Error> {
        ::askama::Template::render(value).map(Into::into)
    }
}
```
This looks like it just calls `render`, gets a `String`, and has the compiler convert that to a `hyper::Body`. I didn't think it uses `respond` or `try_respond`. I'll admit the rust compiler still has me baffled sometimes. I thought I had this one figured out though. Apologies. You know the crate better than I. Please suggest mods or make the necessary cleanups.

Kijewski: Ah, yes, the `<Body as TryFrom<Template>>` does not use `try_respond`, only `<Response<Body> as TryFrom<Template>>` does:
`impl From<Template> for hyper::Body` → `askama_hyper::respond()` → `askama_hyper::try_respond()`.

I would call this an implementation detail which may change at any point, though. Do you think it should be highlighted to the user?

FrankReh: Really? First I'm wrong. Then the public API has details that don't need to be documented because they may change at any point? I think something else is going on here. I won't push this further.

Kijewski: I guess you did not express myself properly. I only meant that the user should not need to know (or care) how `TryFrom<Template> for hyper::Body` was implemented. That the functions exists is not an implementation detail.

And I only commented on the third paragraph. I think that documenting the hyper integration is a good addition to the book, not only for new users of Askama, but all users.

PR #832, opened 2023-06-28
WIP: parse sources only once

Kijewski:


Kijewski:
Closes in favor of #833.

PR #833, opened 2023-06-28
Parse templates once

djc:
@Kijewski exactly the same idea (almost the same branch name, too)! You were faster, but this feels cleaner -- presumably your solution comes down to the same solution under the hood. At the cost of one instance of `unsafe`, but I feel pretty safe with this. What do you think?

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1,7 +1,7 @@
 use crate::config::{get_template_source, read_config_file, Config, WhitespaceHandling};
 use crate::heritage::{Context, Heritage};
 use crate::input::{Print, Source, TemplateInput};
-use crate::parser::{parse, Cond, CondTest, Expr, Loop, Node, Target, When, Whitespace, Ws};
+use crate::parser::{Cond, CondTest, Expr, Loop, Node, Target, When, Whitespace, Ws};
 use crate::CompileError;
 
 use proc_macro::TokenStream;
@@ -1047,8 +1047,19 @@ impl<'a> Generator<'a> {
             .input
             .config
             .find_template(path, Some(&self.input.path))?;
-        let src = get_template_source(&path)?;
-        let nodes = parse(&src, self.input.syntax)?;
+
+        enum CowNodes<'a> {
+            Owned(Parsed),
+            Borrowed(&'a [Node<'a>]),
+        }
+
+        let nodes = match self.contexts.get(path.as_path()) {
+            Some(ctx) => CowNodes::Borrowed(&ctx.nodes),
+            None => {
+                let src = get_template_source(&path)?;
+                CowNodes::Owned(Parsed::new(src, self.input.syntax)?)
+            }
+        };
 
         // Make sure the compiler understands that the generated code depends on the template file.
         {
@@ -1065,7 +1076,15 @@ impl<'a> Generator<'a> {
             // Since nodes must not outlive the Generator, we instantiate
             // a nested Generator here to handle the include's nodes.
             let mut gen = self.child();
-            let mut size_hint = gen.handle(ctx, &nodes, buf, AstLevel::Nested)?;
+            let mut size_hint = gen.handle(
+                ctx,
+                match &nodes {
+                    CowNodes::Owned(parsed) => parsed.nodes(),
+                    CowNodes::Borrowed(nodes) => nodes,
+                },
+                buf,
+                AstLevel::Nested,
+            )?;
             size_hint += gen.write_buf_writable(buf)?;
             size_hint
         };
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -322,17 +322,6 @@ impl<'a> Generator<'a> {
         }
     }
 
-    fn child(&mut self) -> Generator<'_> {
-        let locals = MapChain::with_parent(&self.locals);
-        Self::new(
-            self.input,
-            self.contexts,
-            self.heritage,
-            locals,
-            self.whitespace,
-        )
-    }
-
     // Takes a Context and generates the relevant implementations.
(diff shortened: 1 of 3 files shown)


Kijewski:
Your version does not handle `{%include%}`s, i.e. they are still parsed twice. But given that I've never used includes and would prefer `{%imports%}` every time, I daresay that's fine.

The one line of unsafe code is fine. Ouroboros does exactly the same as you did in your code. The fewer dependencies the better.

askama_derive/src/generator.rs
```diff
@@ -237,11 +236,46 @@ fn find_used_templates(
...
 }
 
+mod _parsed {
+    use std::mem;
+
+    use crate::config::Syntax;
+    use crate::parser::{parse, Node};
+    use crate::CompileError;
+
+    pub(super) struct Parsed<'config> {
```
Kijewski, line 251: Does this actually need a named lifetime? I think `Node<'config>` can be `Node<'static>`. Parsed::nodes() reduces the lifetime to `Node<'self>` as it should be.

djc: The compiler was complaining that the lifetime of the syntax has to outlive the `Node`'s lifetime, which seems reasonable?

djc:
I'll see if I can fix up includes next, it seems reasonable to try to fix them. But this seems like an unambiguous improvement, so with your approval I think we can just merge this.

askama_derive/src/generator.rs
Kijewski: That's because of `parse<'a>(src: &'a str, syntax: &'a Syntax<'_>)`. `&'a Syntax<'_>` should be `&Syntax<'_>`: https://github.com/djc/askama/compare/parse-once...Kijewski:askama:parse-once

djc: Nice, thanks!

djc:
Just to randomly express this, collaboration like this makes me grateful to have you as a co-maintainer on this project!

askama_derive/src/generator.rs
```diff
@@ -1037,16 +1040,16 @@ impl<'a> Generator<'a> {
...
-            Borrowed(&'a [Node<'a>]),
-        }
-
         let nodes = match self.contexts.get(path.as_path()) {
-            Some(ctx) => CowNodes::Borrowed(&ctx.nodes),
+            Some(ctx) => ctx.nodes,
             None => {
-                let src = get_template_source(&path)?;
-                CowNodes::Owned(Parsed::new(src, self.input.syntax)?)
+                if !self.includes.contains_key(path.as_path()) {
```
djc, line 1046: I tried to do this with an `Entry` to avoid double-hashing, but couldn't quite get that to work due to the limited lifetime of the `Entry` value. I think the overhead of extra hashing is probably not worth it.

askama_derive/src/generator.rs
```diff
@@ -1061,8 +1064,8 @@ impl<'a> Generator<'a> {
             )?;
         }
 
-        // Since nodes must not outlive the Generator, we instantiate
-        // a nested Generator here to handle the include's nodes.
+        // Since nodes must not outlive the Generator, we instantiate a nested `Generator` here to
```
djc, line 1067: Because we pass `nodes` to `child.handle()` and `nodes` need to outlive the child `Generator`, I don't think we can easily share a map of included ASTs between parents and children. This should already be quite the improvement, as at least includes from the same template don't have to reparse.

askama_derive/src/generator.rs
Kijewski: This should work:

```rust
None => match self.includes.entry(path) {
    Entry::Occupied(entry) => entry.into_mut().nodes(),
    Entry::Vacant(entry) => {
        let src = get_template_source(entry.key())?;
        entry.insert(Parsed::new(src, self.input.syntax)?).nodes()
    }
}
```

Then the "Make sure the compiler understands that the generated code depends on the template file." block needs to come before `let nodes = …`.

askama_derive/src/generator.rs
```diff
@@ -237,18 +236,52 @@ fn find_used_templates(
...
+        source: String,
+        nodes: Vec<Node<'static>>,
+    }
+
+    impl Parsed {
+        pub(super) fn new(source: String, syntax: &Syntax<'_>) -> Result<Self, CompileError> {
+            // Self-referential borrowing: `self` will keep the source alive as `String`,
+            // internally we will transmute it to `&'config str` to satisfy the compiler.
+            // However, we only expose the nodes with a lifetime limited to `self`.
+            let src = unsafe { mem::transmute::<_, &'static str>(source.as_str()) };
```
Kijewski, line 262: Maybe explicitly use `mem::transmute::<&str, &'static str>`, so we can be sure that we are not changing the type, but only its lifetime?

Kijewski:
> Just to randomly express this, collaboration like this makes me grateful to have you as a co-maintainer on this project!

Thank you! :)

The difference this PR does is measurable. I did not chose the most scientific benchmark one could come up with (

```sh
find testing/ -name '*.rs' -exec touch '{}' +
time cargo t
```

) but I got this result:

```text
       before    after  speedup
real   2.469s	2.402s    2.71%
user  15.687s  15.558s    8.22%
sys    3.680s	3.409s    7.36%

```

Kijewski (approved):
Nice!

Thank you @wrapperup for finding this improvement.

wrapperup:
Tried it in my test. Parsing 100 templates, 10 runs averaged. Results are (WSL):

| Target | Time Avg (Debug) | Time Avg (Release) |
|--------|------------------|--------------------|
| 0.12   | 3.932s           | 182ms              |
| main   | 2.086s           | 126ms              |

Pretty big improvements, debug is almost twice as fast. Thanks @djc and @Kijewski !

PR #837, opened 2023-07-05
Fix Rust macro invocations not accepting a path

### LINKED ISSUE #836: Parse error when using macro with expanded module path
I am making a crate that allows people to embed icons from Iconify at compile time, but I ran into a little snag.
Tried on main and 0.12. Only happens when using a macro, normal functions work fine.

The intended usage is:
```jsx
<body>
    <h1>Hello</h1>
    {{ iconify::svg!("mdi:home")|safe }}
</body>
```

But it gives me a parse error:
```
error: problems parsing template source at row 5, column 19 near:
       "!(\"mdi:home\") }}\r\n</body>\r\n</html>\r"
```

I can work around it by importing the macro (below compiles fine), but I'd prefer being able to keep the module path so I don't need to.
```jsx
{{ svg!("mdi:home")|safe }}
```

wrapperup:
Fixes #836 

This just changes the RustMacro expression to accept a path, and runs it before `expr_path`, pretty simple small change. I did try making it a part of `expr_suffix`, but the parser would need to accept `!(...)` patterns, which seems to be much harder to parse correctly. This way also seems to be closer to how `rustc` parses macro invocations too.

### DIFF
--- a/askama_derive/src/parser/expr.rs
+++ b/askama_derive/src/parser/expr.rs
@@ -30,7 +30,7 @@ pub(crate) enum Expr<'a> {
     Group(Box<Expr<'a>>),
     Tuple(Vec<Expr<'a>>),
     Call(Box<Expr<'a>>, Vec<Expr<'a>>),
-    RustMacro(&'a str, &'a str),
+    RustMacro(Vec<&'a str>, &'a str),
     Try(Box<Expr<'a>>),
 }
 
@@ -182,8 +182,8 @@ fn expr_single(i: &str) -> IResult<&str, Expr<'_>> {
         expr_num_lit,
         expr_str_lit,
         expr_char_lit,
-        expr_path,
         expr_rust_macro,
+        expr_path,
         expr_array_lit,
         expr_var,
         expr_group,
@@ -274,8 +274,12 @@ fn macro_arguments(i: &str) -> IResult<&str, &str> {
 }
 
 fn expr_rust_macro(i: &str) -> IResult<&str, Expr<'_>> {
-    let (i, (mname, _, args)) = tuple((identifier, char('!'), macro_arguments))(i)?;
-    Ok((i, Expr::RustMacro(mname, args)))
+    let (i, (path_vec, _, args)) = tuple((
+        alt((path, map(identifier, |id| vec![id]))),
+        char('!'),
+        macro_arguments,
+    ))(i)?;
+    Ok((i, Expr::RustMacro(path_vec, args)))
 }
 
 macro_rules! expr_prec_layer {
--- a/askama_derive/src/parser/expr.rs
+++ b/askama_derive/src/parser/expr.rs
@@ -4,9 +4,10 @@ use nom::branch::alt;
 use nom::bytes::complete::{tag, take_till};
 use nom::character::complete::char;
 use nom::combinator::{cut, map, not, opt, peek, recognize};
+use nom::error::ErrorKind;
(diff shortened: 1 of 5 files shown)


Kijewski:
> This way also seems to be closer to how rustc parses macro invocations too.

Yes, you're right. Something like `{{ a.b.c!() }}` should fail to parse.

djc (approved):
Thanks!

djc:
@Kijewski do you want to take a look at this?

askama_derive/src/parser/expr.rs
```diff
@@ -182,8 +182,8 @@ fn expr_single(i: &str) -> IResult<&str, Expr<'_>> {
         expr_num_lit,
         expr_str_lit,
         expr_char_lit,
-        expr_path,
         expr_rust_macro,
+        expr_path,
```
Kijewski, line 186: I actually dislike this reordering, because it means that the stream is parsed three times for the average case, i.e. one is not invoking a macro, and one is utilizing an identifier/variable instead of a path. I think it would be best to implement parsing macros inside of `expr_suffix` anyway. We just need to make sure that the left-hand side is a path or a macro.

Kijewski (changes requested):

Kijewski:
Side question: Do we want to allow bracket as delimiters? I only ever saw `syn` even mentioning them, and only for [`Token![]`](https://docs.rs/syn/2.0.23/syn/macro.Token.html).

djc:
Sounds like the kind of thing that would need two separate people complaining about it before we implement it.

askama_derive/src/parser/expr.rs
wrapperup: It might get messy. Something has to accept `path::macro!` and `macro!` or `!(..)` as an expression to parse down the line. I did try adding the capability to identifiers and paths to accept a macro, but it was very fragile. Maybe the better idea is to fix the ordering instead?

wrapperup: It does this now 👍

Kijewski:
@wrapperup, did you see https://github.com/wrapperup/askama/pull/1?

wrapperup:
> @wrapperup, did you see [wrapperup#1](https://github.com/wrapperup/askama/pull/1)?

I didn't! It looks very similar to what I did though, so I'll see if there's changes I can merge in that I might've missed.

wrapperup:
Looks like a new clippy lint is causing the build to fail unrelated to these changes, but everything else works fine.

Kijewski (approved):
Nice! LGTM, thank you!

askama_derive/src/parser/expr.rs
```diff
@@ -194,6 +194,7 @@ enum Suffix<'a> {
     Attr(&'a str),
     Index(Expr<'a>),
     Call(Vec<Expr<'a>>),
+    MacroCall(&'a str),
```
djc, line 197: What is the value of the unit variant here representing? Seems strange that it isn't a `Vec` of something like for `Call`.

Kijewski: It's the whole content, everything in `(…)`. Proc macros can access the input verbatim including any white spaces, so I don't think we should not attempt to understand what's going on in a macro invocation, just paste it verbatim.

wrapperup: It's representing the inner body/args of the macro call. It's possible Vec of Expr would work there, but given that it can accept potentially any kind of syntax (valid rust or not), it's probably the safer bet to just pass the args as is.

PR #839, opened 2023-07-14
Fix `clippy::items-after-test-module` error

wrapperup:
A new clippy lint (items_after_test_module) was added in rust 1.71, which is causing builds to fail in the CI.


Kijewski (approved):
LGTM! Thank you!

Kijewski:
Clippy, my old friend, sometimes you are overdoing it. :)

PR #840, opened 2023-07-14
Parse paths and identifiers only once

Kijewski:
In the old implementation each variable in an expression would be parsed up to three times:

* Try to parse a path because it contains a leading double colon, or infix double colons.
* Try to parse it as path again by scanning for an identifier that contains an upper case character.
* Fall back to scanning for any identifier.

This PR turns all three steps into one, without the need for backtracking.

### DIFF
--- a/askama_parser/src/expr.rs
+++ b/askama_parser/src/expr.rs
@@ -9,7 +9,10 @@ use nom::multi::{fold_many0, many0, separated_list0};
 use nom::sequence::{pair, preceded, terminated, tuple};
 use nom::{error_position, IResult};
 
-use super::{bool_lit, char_lit, identifier, not_ws, num_lit, path, str_lit, ws};
+use super::{
+    bool_lit, char_lit, identifier, not_ws, num_lit, path_or_identifier, str_lit, ws,
+    PathOrIdentifier,
+};
 
 macro_rules! expr_prec_layer {
     ( $name:ident, $inner:ident, $op:expr ) => {
@@ -141,9 +144,8 @@ impl<'a> Expr<'a> {
             Self::num,
             Self::str,
             Self::char,
-            Self::path,
+            Self::path_or_var,
             Self::array,
-            Self::var,
             Self::group,
         ))(i)
     }
@@ -186,13 +188,11 @@ impl<'a> Expr<'a> {
         )(i)
     }
 
-    fn path(i: &'a str) -> IResult<&'a str, Self> {
-        let (i, path) = path(i)?;
-        Ok((i, Self::Path(path)))
-    }
-
-    fn var(i: &'a str) -> IResult<&'a str, Self> {
-        map(identifier, Self::Var)(i)
+    fn path_or_var(i: &'a str) -> IResult<&'a str, Self> {
+        map(path_or_identifier, |v| match v {
+            PathOrIdentifier::Path(v) => Self::Path(v),
+            PathOrIdentifier::Identifier(v) => Self::Var(v),
+        })(i)
     }
 
     fn str(i: &'a str) -> IResult<&'a str, Self> {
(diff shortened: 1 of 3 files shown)


askama_derive/src/parser/mod.rs
```diff
@@ -240,31 +240,38 @@ fn nested_parenthesis(i: &str) -> IResult<&str, ()> {
...
-                // - Doesn't contain any lowercase characters, i.e. it's a constant
-                // In short, if it contains any uppercase characters it's a path.
-                if name.contains(char::is_uppercase) {
-                    return Ok((i, vec![name]));
-                }
-            }
-
-            // If `identifier()` fails then just return the original error
-            Err(err)
+        (None, name, []) if !name.contains(char::is_uppercase) => {
```
djc, line 267: What tests break if we leave out this arm? Maybe move the comment to be directly before this arm, which seems to be the main thing that's affected?

askama_derive/src/parser/expr.rs
```diff
@@ -139,13 +135,14 @@ fn expr_char_lit(i: &str) -> IResult<&str, Expr<'_>> {
...
-fn expr_var(i: &str) -> IResult<&str, Expr<'_>> {
-    map(identifier, Expr::Var)(i)
-}
-
-fn expr_path(i: &str) -> IResult<&str, Expr<'_>> {
-    let (i, path) = path(i)?;
-    Ok((i, Expr::Path(path)))
+fn expr_path_or_var_or_bool_lit(i: &str) -> IResult<&str, Expr<'_>> {
+    map(path_or_identifier, |v| match v {
+        super::PathOrIdentifier::Path(v) => Expr::Path(v),
```
djc, line 140: Would prefer to import PathOrIdentifier here.

askama_derive/src/parser/expr.rs
```diff
@@ -139,13 +135,14 @@ fn expr_char_lit(i: &str) -> IResult<&str, Expr<'_>> {
...
+        super::PathOrIdentifier::Identifier(v) => match v {
```
djc, line 141: I think we could outline this into three separate top-levels arms instead of a nested match? That seems nicer to me.

djc (approved):
Nice! Did you benchmark the effect?

askama_derive/src/parser/mod.rs
Kijewski: This arm tells variable names (no upper case characters) and struct names (at least one upper case character) apart. I guess every test will fail without it. :)

Kijewski:
As benchmark I used `cd testing`; `find -name '*.rs' -exec touch '{}' '+'`; `time cargo b  --tests` (test out of three runs):

```
       real     user    sys
old  1.623s  10.442s  2.235s
new  1.473s  10.483s  2.272s
```

I cannot say why the real time got so much better, if the user time got worse? All CPU cores only have around 33°C, so they should all be utilized in my test without thermal throttling. I guess I should try to find a better benchmark, but I guess without #834 we cannot properly benchmark the parser. :-/ I'll try to review your PR this weekend!

djc (approved):

djc:
(Feel free to just merge this after rebasing.)

Kijewski:
Not a huge difference, but it's a little better:

```text
100 bytes               time:   [11.475 µs 11.594 µs 11.694 µs]
                        change: [-6.8066% -4.5471% -2.2145%] (p = 0.00 < 0.05)
                        Performance has improved.

1000 bytes              time:   [62.269 µs 62.421 µs 62.555 µs]
                        change: [-4.8183% -4.1854% -3.5724%] (p = 0.00 < 0.05)
                        Performance has improved.

10000 bytes             time:   [536.05 µs 536.83 µs 537.69 µs]
                        change: [-3.2005% -2.9935% -2.7835%] (p = 0.00 < 0.05)
                        Performance has improved.

100000 bytes            time:   [4.9944 ms 4.9996 ms 5.0053 ms]
                        change: [-2.8939% -2.7114% -2.5330%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #700, opened 2022-06-17
I18n

89Q12:
This PR is based on the work of [jhoobergs](https://github.com/jhoobergs) and PR #434.
I mainly just updated the code to work with the changes that have been made since the initial PR, and I split it into its own feature "localization" which can be renamed to maybe "with-localization". 
Also I added askama::Locale as suggested in  #434.

Pending:
- Docs
-  Improve the feature, mainly the use of  #[cfg(feature = "localization")] since I'm new to rust/cargo

All tests pass.
Looking at the old PR the TODO: "Fix the test for invalid text_id's" isn't really needed since we make use of a function from fluent that checks for a fallback language which is why the test doesn't panic, in fact it will never panic as seen [here](https://docs.rs/fluent-templates/0.7.1/src/fluent_templates/loader/static_loader.rs.html#78)

PR #841, opened 2023-07-16
i18n

TTWNO:
A PR with the same basic features as #700 but up to date with current main.


djc:
My feedback from #700 still stands, I guess:

> I guess the main thing that could help resolving this deadlock would be to represent the changes here as a series of a smaller PR that have one logical change per PR, so that I can review each of them quickly and thus we can make progress.

PR #843, opened 2023-07-21
Fix parsing calls

Kijewski:
This change:

* adds a cut when the leading `(` was encountered, and
* fixed the interaction between call expressions and boolean OR.


djc (approved):

djc:
This will need to be rebased now, hope it's not too painful!

Kijewski:
No problem at all. :)

djc (approved):

PR #844, opened 2023-07-30
Derive i18n support

TTWNO:
An smaller PR, with basic functionality of #841 , but with some important differences:

1. Only sections of code that use the optional `i18n` dependencies are feature gated. The `Expr` type will always contain a `Localization` variant, but the developer will still be given an error if they attempt to use `i18n` within a template without the feature enabled.
2. Only the basic, underlying derivations are implemented in this PR. This does not connect `i18n` into other parts of the system. This will be for a future PR, since djc wants smaller, more manageable PRs for this large feature.
3. Removes the use of the `Unlazy` type, implemented with `parking_lot` and an `unsafe { transmute ... } `; whether or not its supposed "unsafely" is really an issue is not my main concern. Rather, creating a new type, just to implement one singular function seems excessive. Perhaps I'm wrong about this, and I'm happy to revert that change upon request.

Over half the code, 393 lines, is in the `src/i18n.rs` file, all other modifications are fairly small and simply introduce the ability to handle localization expressions.

This is the biggest PR by far, since all functions within `src/i18n.rs` are required to get the expressions working correctly.


djc:
Thanks for splitting this up! Unfortunately I just merged #834, which will require some changes here.

GuillaumeGomez:
Ping @TTWNO. Do you think you will have time to come back to this? It'd be a shame that this great feature wouldn't be merged. :)

TTWNO:
Ah! Totally forget.

Let me see what I can do this afternoon.

TTWNO:
I... think I fixed the conflicts and still integrated it properly.

Let me know if this needs some additional work. I realize making `askama_parse` optionally dependent on `fluent-*` is not ideal.

TTWNO:
I fixed the formatting. The warnings are mostly a big bunch of functions that are not run. IIUC, I'm not actually calling the `exp_localization` anywhere, or if I am, it can not be triggered.

Anyways, this is really close, and I would appreciate a review on what I need to fix to get this over the finish line.

Some dependency depends on a later version of `tinystr` than the MSRV for the crate. It's two versions ahead `1.67 vs 1.65`. I'd like some direction on whether explicitly downgrading the dependency is better, or if the MSRV of askama should be raised.

TTWNO:
I am still committed to getting this merged. My biggest issue is that I do not often have time to work on this in big blocks (20 minutes here, an hour there) and so it ends up languishing in "I'll get to it" hell.

I'll emphasize again that a bit of TLC would be super helpful in getting this wrapped up. If anyone is willing to review and give this some help, I will get on at least once a week to read and make changes as necessary.

I am still using my old, branched version in an activate project to have internationalization support. It would be super helpful to start using a mainline version of askama :)

GuillaumeGomez:
That's still 90 commits. Can you squash a bit please? ^^'

testing/tests/i18n.rs
```diff
@@ -0,0 +1,58 @@
+#![cfg(feature = "i18n")]
+
+use askama::i18n::{langid, Locale};
+use askama::Template;
+
+askama::i18n::load!(LOCALES);
```
djc, line 6: I think this should look like `static LOCALES: _ = askama::i18n::load!();` -- what is preventing us from doing something more like that?

askama_parser/src/i18n.rs
```diff
@@ -0,0 +1,425 @@
+use fluent_templates::LanguageIdentifier;
```
djc, line 1: We should not duplicate all this stuff across the derive and parser crates.

askama_parser/src/expr.rs
```diff
@@ -300,6 +303,74 @@ impl<'a> Expr<'a> {
     fn char(i: &'a str) -> ParseResult<'a, Self> {
         map(char_lit, Self::CharLit)(i)
     }
+
+    #[cfg(not(feature = "i18n"))]
+    fn localize(i: &'a str) -> ParseResult<'a, Self> {
+        let (i, _) = pair(tag("localize"), ws(tag("(")))(i)?;
+        eprintln!(r#"Activate the "i18n" feature to use {{ localize() }}."#);
```
djc, line 310: This should not be an `eprintln!()`, but some form of error that is yielded to the caller. (Similar for the other error conditions here.)

askama_parser/Cargo.toml
```diff
@@ -13,5 +13,14 @@ readme = "README.md"
...
+i18n = ["fluent-syntax", "fluent-templates", "serde", "basic-toml"]
+
 [dependencies]
 nom = { version = "7", default-features = false, features = ["alloc"] }
+fluent-syntax = { version = "0.11.0", default-features = false, optional = true }
+fluent-templates = { version = "0.8.0", default-features = false, optional = true }
+serde = { version = "1.0.193", default-features = false, features = ["derive"], optional = true }
+basic-toml = { version = "0.1.7", default-features = false, optional = true }
+syn = { version = "2.0.43", default-features = false }
+proc-macro2 = { version = "1.0.71", default-features = false }
```
djc, line 26: These should be ordered alphabetically -- but also, why do we need all this extra stuff in the parser crate? Maybe this was misrebased?

askama_derive/src/i18n.rs
```diff
@@ -0,0 +1,387 @@
...
+        use_isolating: fluent.use_isolating.unwrap_or(false),
+        core_locales,
+        locales,
+        fallbacks,
+        assets_dir,
+    })
+}
+
+fn get_i18n_config() -> Result<&'static Configuration, CompileError> {
+    lazy_static! {
```
djc, line 228: We should not use `lazy_static!()`, instead use a `OnceLock` or `once_cell::Lazy`.

askama_derive/src/i18n.rs
```diff
@@ -0,0 +1,387 @@
...
+    }
+    if resources.is_empty() {
+        return Ok(None);
+    }
+
+    resources.sort_by(|(l, _), (r, _)| Path::cmp(l, r));
+    Ok(Some((language, resources)))
+}
+
+fn read_configuration() -> Result<Configuration, String> {
```
djc, line 143: Could this configuration be part of `askama.toml`? I don't see a good reason why it should be a separate file?

askama/src/i18n.rs
```diff
@@ -0,0 +1,129 @@
...
+
+/// Similar to OnceCell, but it has an additional take() function, which can only be used once,
+/// and only if the instance was never dereferenced.
+///
+/// The struct is only meant to be used by the [`i18n_load!()`] macro.
+/// Concurrent access will deliberately panic.
+///
+/// Rationale: StaticLoader cannot be cloned.
+#[doc(hidden)]
+pub struct Unlazy<T>(parking_lot::Mutex<UnlazyEnum<T>>);
```
djc, line 85: Why do we need this? The stated rationale is that `StaticLoader` cannot be cloned, but why would it need to be?

Cargo.toml
```diff
@@ -23,3 +22,5 @@ default-members = [
     "askama_parser",
     "testing",
 ]
+
+resolver = "2"
```
djc, line 26: Please revert this change.

djc (commented):
No need to squash this for now, let's just review it as one big commit because it looks like that might make sense for this. The test case looks pretty good and a lot of the other stuff does, too, but there are still seems to be a bunch of unnecessary complexity.

askama_derive/src/i18n.rs
Ben-PH: standby for pr update:

```diff
@@ -1,4 +1,4 @@
-use fluent_templates::LanguageIdentifier;
+use fluent_templates::{once_cell, LanguageIdentifier };
 use std::borrow::Cow;
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
@@ -259,12 +259,9 @@ fn read_configuration() -> Result<Configuration, String> {
     })
 }

-use fluent_templates::lazy_static::lazy_static;

 fn get_i18n_config() -> Result<&'static Configuration, CompileError> {
-    lazy_static! {
-        static ref CONFIGURATION: Result<Configuration, String> = read_configuration();
-    }
+    static CONFIGURATION: once_cell::sync::Lazy<Result<Configuration, String>> = once_cell::sync::Lazy::new(read_configuration);
```

testing/tests/i18n.rs
Ben-PH: this is the expansion of line 6:

```rust
static LOCALES: ::askama::i18n::fluent_templates::once_cell::sync::Lazy<
    ::askama::i18n::fluent_templates::StaticLoader,
> = ::askama::i18n::fluent_templates::once_cell::sync::Lazy::new(|| {
    mod fluent_templates {
        pub use ::askama::i18n::fluent_templates::*;
        pub mod once_cell {
            pub mod sync {
                pub use ::askama::i18n::Unlazy as Lazy;
            }
        }
    }
    pub static LOCALES: fluent_templates::once_cell::sync::Lazy<
        fluent_templates::StaticLoader,
    > = fluent_templates::once_cell::sync::Lazy::new(|| {
        static CORE_RESOURCE: fluent_templates::once_cell::sync::Lazy<
            Option<fluent_templates::fluent_bundle::FluentResource>,
        > = fluent_templates::once_cell::sync::Lazy::new(|| { None });
....
```
...and so on

...so the expansion handles the variable setting. Because the type can't be elided ([it being static](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=45450a94c14613c85fa16a2a13db81da)),  that would be requiring the end user to write like so:

```rust
static LOCALES: askama::18n::fluent_templates::once_cell::Lazy<askama::i18n::fluent_templates::StaticLoader> = askama::i18n::load!();

// with sane re-exports
static LOCALES: askama::i18n::LazyStaticLoader = askama::i18n::load!();
```



OP is being consistent, in a sense, with the pattern used by the wrapped `fluent_templates::static_loader` macro, which is documented like so:


```rust
//! ### Basic Example
//! ```
//! fluent_templates::static_loader! {
//!     // Declare our `StaticLoader` named `LOCALES`.
//!     static LOCALES = {
//!         // The directory of localisations and fluent resources.
//!         locales: "./tests/locales",
//!         // The language to falback on if something is not present.
//!         fallback_language: "en-US",
//!         // Optional: A fluent resource that is shared with every locale.
//!         core_locales: "./tests/locales/core.ftl",
//!     };
//! }
//! # fn main() {}
//! ```
```

...in our case, though, the wrapper handles the locales and other settings by means of reading the i18n.toml file, if i'm reading things correctly

I think it would be worth some follow-up after this PR is merged to see if it would be nice to "de-abstract" a touch the underlying fluent engine. The scenario I have in mind is that an "actix-web + askama" project should be internationalised by turning it into an "actix-web + askama + fluent" and not "actix-web + fluent + askama-with-i18n". That probably doesn't make sense, but this conversation is a bit out of scope, so I'll clarify when it comes time to put on the polish.

askama_parser/src/expr.rs
Ben-PH: like so?

```diff
     #[cfg(not(feature = "i18n"))]
     fn localize(i: &'a str) -> ParseResult<'a, Self> {
+        // Happy path: no i18n _should_ fail to parse "localize ("
         let (i, _) = pair(tag("localize"), ws(tag("(")))(i)?;
-        eprintln!(r#"Activate the "i18n" feature to use {{ localize() }}."#);
-        Err(nom::Err::Failure(error_position!(i, ErrorKind::Tag)))
+
+        Err(nom::Err::Failure(ErrorContext {
+            input: i,
+            message: Some(Cow::Owned(format!(
+                r#"Activate the "i18n" feature to use {{ localize() }}."#
+            )))
+        }))
     }
```

djc: Yup, that looks reasonable to me!

testing/tests/i18n.rs
djc: I think `static LOCALES: askama::i18n::LazyStaticLoader = askama::i18n::load!();` looks better, I prefer it over the more opaque/comprehensive wrapper macro.

askama/src/i18n.rs
```diff
@@ -0,0 +1,129 @@
...
+pub use askama_derive::i18n_load as load;
+
+pub use fluent_templates::{self, fluent_bundle::FluentValue, fs::langid, LanguageIdentifier};
+use fluent_templates::{Loader, StaticLoader};
+use parking_lot::const_mutex;
+
+pub struct Locale<'a> {
+    loader: &'a StaticLoader,
+    language: LanguageIdentifier,
+}
```
Restioson, line 56: Would it be possible to also support non-static loaders (e.g ArcLoader)? Could be a dyn type or parameterised by the type. Wondering as I have a use-case for this

djc: What's the use case?

djc: Okay, I think that all sounds plausible, but I'd probably like to declare it out of scope for the initial PR.

PR #845, opened 2023-07-30
Askama base crate i18n support

TTWNO:
Depends on #844 :

* Adds `Locale` struct behind feature flag,
* Does not include unsafe code, unlike the original #700 PR
* This also drops a dependency on `parking_lot`, since it was used as part of the unsafe code
* 113 line diff between this PR and #844

### DIFF
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -33,6 +33,7 @@ with-mendes = ["askama_derive/with-mendes"]
 with-rocket = ["askama_derive/with-rocket"]
 with-tide = ["askama_derive/with-tide"]
 with-warp = ["askama_derive/with-warp"]
+i18n = ["askama_derive/i18n", "fluent-templates", "parking_lot"]
 
 # deprecated
 mime = []
@@ -48,6 +49,8 @@ percent-encoding = { version = "2.1.0", optional = true }
 serde = { version = "1.0", optional = true, features = ["derive"] }
 serde_json = { version = "1.0", optional = true }
 serde_yaml = { version = "0.9", optional = true }
+fluent-templates = { version = "0.8.0", optional = true }
+parking_lot = { version = "0.12.1", optional = true }
 
 [package.metadata.docs.rs]
 features = ["config", "humansize", "num-traits", "serde-json", "serde-yaml"]
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -33,7 +33,7 @@ with-mendes = ["askama_derive/with-mendes"]
 with-rocket = ["askama_derive/with-rocket"]
 with-tide = ["askama_derive/with-tide"]
 with-warp = ["askama_derive/with-warp"]
-i18n = ["askama_derive/i18n", "fluent-templates", "parking_lot"]
+i18n = ["askama_derive/i18n", "fluent-templates"]
 
 # deprecated
 mime = []
@@ -50,7 +50,6 @@ serde = { version = "1.0", optional = true, features = ["derive"] }
 serde_json = { version = "1.0", optional = true }
 serde_yaml = { version = "0.9", optional = true }
 fluent-templates = { version = "0.8.0", optional = true }
-parking_lot = { version = "0.12.1", optional = true }
 
 [package.metadata.docs.rs]
 features = ["config", "humansize", "num-traits", "serde-json", "serde-yaml"]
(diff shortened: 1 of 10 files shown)


askama/Cargo.toml
```diff
@@ -33,6 +33,7 @@ with-mendes = ["askama_derive/with-mendes"]
 with-rocket = ["askama_derive/with-rocket"]
 with-tide = ["askama_derive/with-tide"]
 with-warp = ["askama_derive/with-warp"]
+i18n = ["askama_derive/i18n", "fluent-templates"]
```
Ben-PH, line 36: as this is a `fluid` backed i18n, would it be appropriate to solf-document that in the feature name? e.g. `i18n_fluent`?

askama_derive/src/generator.rs
```diff
@@ -1376,9 +1376,40 @@ impl<'a> Generator<'a> {
...
+            "self.{}.translate(",
+            normalize_identifier(localizer)
+        ));
+        self.visit_expr(buf, msg_id)?;
+        buf.writeln(", [")?;
+        buf.indent();
+        for (k, v) in args {
+            buf.write(&format!("({:?}, ::askama::i18n::FluentValue::from(", k));
+            self.visit_expr(buf, v)?;
+            buf.writeln(")),")?;
```
Ben-PH, line 1404: personal preference for `key` and `val`

askama_derive/src/lib.rs
```diff
@@ -10,14 +10,28 @@ use proc_macro2::Span;
...
+#[proc_macro]
+pub fn i18n_load(_input: TokenStream) -> TokenStream {
+    #[cfg(feature = "i18n")]
+    match i18n::load(_input) {
+        Ok(ts) => ts,
+        Err(err) => err.into_compile_error(),
+    }
+
+    #[cfg(not(feature = "i18n"))]
+    CompileError::from(r#"Activate the "i18n" feature to use i18n_load!()."#).into_compile_error()
```
Ben-PH, line 32: consider using cfg-if

askama_derive/src/parser/expr.rs
```diff
@@ -33,6 +33,8 @@ pub(crate) enum Expr<'a> {
     Call(Box<Expr<'a>>, Vec<Expr<'a>>),
     RustMacro(Vec<&'a str>, &'a str),
     Try(Box<Expr<'a>>),
+    #[allow(dead_code)]
```
Ben-PH, line 36: Is it feasible to feature gate this instead?

askama_derive/src/parser/expr.rs
```diff
@@ -331,9 +336,72 @@ expr_prec_layer!(expr_compare, expr_bor, "==", "!=", ">=", ">", "<=", "<");
...
 expr_prec_layer!(expr_or, expr_and, "||");
 
+#[cfg(not(feature = "i18n"))]
+fn expr_localize(i: &str) -> IResult<&str, Expr<'_>> {
+    let (i, _) = pair(tag("localize"), ws(tag("(")))(i)?;
+    eprintln!(r#"Activate the "i18n" feature to use {{ localize() }}."#);
+    Err(nom::Err::Failure(error_position!(i, ErrorKind::Tag)))
+}
+
+#[cfg(feature = "i18n")]
```
Ben-PH, line 346: consider using cfg-if crate

askama/src/i18n.rs
```diff
@@ -0,0 +1,74 @@
...
+pub struct Locale<'a> {
+    loader: &'a StaticLoader,
+    language: LanguageIdentifier,
+}
+
+impl Locale<'_> {
+    pub fn new(language: LanguageIdentifier, loader: &'static StaticLoader) -> Self {
+        Self { loader, language }
+    }
+
```
Ben-PH, line 61: Doc-comment here would be handy.

Ben-PH (commented):
I'm by no means a heavy-hitter. I'm just passing through in my search of "the good way" to add i18n to my askama project, and thought I'd try to do my little part.

nothing in this review should be considered blocking. I didn't see any issues, but that by no means implies I didn't miss any. I hope this helps. I really look forward to askama + fluent being able to Just Work^tm

askama_derive/src/lib.rs
djc: If you mean the crate, I'd prefer not to add more dependencies unless we really need them.

PR #846, opened 2023-07-30
I18n testing

TTWNO:
Final PR for i18n support, depends on #845 

* Adds basic i18n testing in `testing` directory
* 129 lines diff between this and #845


AleVul:
Any chance of this being reviewed in near term?

TTWNO:
This depends on other changes being merged, notably, #845, which depends on #844.

844 still needs some work, but at least does not conflict with the current release. I could use some help getting it over the finish line.

PR #848, opened 2023-08-01
parser: some fixes and QoL improvements for `Parsed`

Kijewski:


askama_parser/src/lib.rs
```diff
@@ -54,12 +54,12 @@ mod _parsed {
 pub use _parsed::Parsed;
 
 #[derive(Debug)]
-pub struct Ast<'a> {
+struct Ast<'a> {
```
djc, line 71: Hmm, I'd prefer to go the other way on this. I think we should add a `pub fn nodes(&self) -> &[Node<'_>]` accessor and keep `Parsed` as an optional addition.

askama_parser/src/lib.rs
```diff
@@ -57,6 +57,12 @@ mod _parsed {
                 .finish_non_exhaustive()
         }
     }
+
+    impl std::cmp::PartialEq for Parsed {
```
djc, line 61: Style nit: would prefer to import `PartialEq` directly.

djc (approved):

PR #849, opened 2023-08-01
parser: add types for `Node`'s variants

Kijewski:
To make `askama_parser` usable outside of the generator, the meaning of the fields of `Node`'s variants should be obvious.


askama_parser/src/node.rs
```diff
@@ -353,15 +360,13 @@ impl<'a> Node<'a> {
...
-            Self::Macro(
-                Macro {
-                    ws1: Ws(pws1, nws1),
-                    name,
-                    args: params,
-                    nodes: contents,
-                    ws2: Ws(pws2, nws2),
-                },
-            ),
+            Self::Macro(Macro {
```
djc: This belongs in the previous commit.

askama_parser/src/node.rs
```diff
@@ -311,8 +311,15 @@ impl<'a> Node<'a> {
                 cut(pair(ws(identifier), opt(Whitespace::parse))),
             ))),
         ));
-        let (i, (pws, _, (name, _, (scope, nws)))) = p(i)?;
-        Ok((i, Self::Import(Ws(pws, nws), name, scope)))
+        let (i, (pws, _, (path, _, (scope, nws)))) = p(i)?;
```
djc, line 346: If we're going to do this I think we should also move the parser into an `Import::parse()` constructor, with `Node::import()` only wrapping the result of that in `Self::Import`.

askama_parser/src/node.rs
```diff
@@ -156,14 +156,20 @@ impl<'a> Node<'a> {
         ));
 
         let (i, (pws1, cond, (nws1, _, (nodes, elifs, (_, pws2, _, nws2))))) = p(i)?;
-        let mut res = vec![Cond {
+        let mut branches: Vec<Cond<'_>> = vec![Cond {
```
djc, line 159: Don't think we need the type annotation here?

askama_parser/src/node.rs
```diff
@@ -753,7 +753,7 @@ impl From<char> for Whitespace {
 }
 
 #[derive(Debug, PartialEq)]
-pub struct Loop<'a> {
+pub struct NodeLoop<'a> {
```
djc, line 756: Not a fan of this. I suppose you think conflicts will become an issue because you want to give the same treatment to `Expr`? I feel like maybe we should leave all the single-value `Expr` variants alone, adding wrappers there seems mostly pointless to me -- even here I'm not really convinced about the `Extends` case.

Instead, we could make `expr` and `node` public modules and then callers can reference `expr::Macro` and `node::Macro`, which seems nicer to me.

djc (commented):
Thanks for working on this!

askama_parser/src/node.rs
Kijewski: I was inspired by [syn](https://docs.rs/syn/latest/syn/), which does it this way. Yes, you're right, I'll make the module pub.

askama_parser/src/node.rs
```diff
@@ -802,6 +779,46 @@ impl<'a> BlockDef<'a> {
...
+                // {block,comment,expr}_start follows immediately.
+                return Err(nom::Err::Error(error_position!(i, ErrorKind::TakeUntil)));
+            }
+            Some(content) => (i, content),
+            None => ("", i), // there is no {block,comment,expr}_start: take everything
+        };
+        Ok((i, Self::split_ws_parts(content)))
+    }
+
+    pub(crate) fn split_ws_parts(s: &'a str) -> Self {
```
djc, line 677: Ohhh, this is nice.

askama_parser/src/lib.rs
```diff
@@ -14,21 +14,18 @@ use nom::multi::separated_list1;
...
 
-pub use self::expr::Expr;
-pub use self::node::{
-    BlockDef, Call, Cond, CondTest, Extends, If, Import, Include, Let, Lit, Loop, Macro, Match,
-    Node, Raw, Target, When, Whitespace, Ws,
-};
-
-mod expr;
-mod node;
+use self::node::Node;
```
djc, line 17: This should be spelled as `use node::Node;`, and please move it directly underneath the `pub mod node;` declaration.

I'm inclined to `pub use node::Node;` and `pub use expr::Expr;` which are arguably higher-level types, and leave the inner types as things that should be imported through the module name, what do you think?

askama_parser/src/tests.rs
```diff
@@ -498,109 +498,109 @@ fn test_odd_calls() {
...
+                Node::Comment(comment) => comment.ws,
+                val => panic!("expected comment, but parsed={:?}", val),
+            }
+        }
+    }
+
     assert_eq!(
-        Ast::from_str("{##}", s).unwrap().nodes,
-        vec![Node::Comment(Ws(None, None))],
+        Ast::from_str("{##}", s).unwrap().nodes.one_comment_ws(),
```
djc, line 516: I feel like `one_comment_ws()` should just be a function that takes a `&str`, so it can subsume the whole thing starting from `Ast::from_str()`. Any reason not to do so?

djc (commented):
Great stuff!

Kijewski:
> Great stuff!

Thanks! You did the difficult work in #834. That made this smaller refactoring easy. :)

Should `Node::Loop` maybe box its content? It's 208 bytes. The next biggest is only half its size, `Node::Let` with 112 bytes.

djc:
> Should `Node::Loop` maybe box its content? It's 208 bytes. The next biggest variant is only half its size, `Node::Let` with 112 bytes.

If clippy is not complaining, I'm inclined to add the extra allocation.

askama_parser/src/lib.rs
```diff
@@ -14,21 +14,18 @@ use nom::multi::separated_list1;
...
+pub use node::Node;
 #[cfg(test)]
 mod tests;
 
 mod _parsed {
     use std::mem;
 
-    use super::{Ast, Node, ParseError, Syntax};
+    use super::node::Node;
+    use super::{Ast, ParseError, Syntax};
```
djc, line 28: We can now import `Node` directly from `super`.

askama_derive/src/generator.rs
```diff
@@ -2,10 +2,12 @@ use crate::config::{get_template_source, read_config_file, Config, WhitespaceHan
...
 use crate::CompileError;
-use parser::{
-    Call, CondTest, Expr, If, Include, Let, Lit, Loop, Match, Node, Parsed, Target, Whitespace, Ws,
-};
 
+use parser::expr::Expr;
+use parser::node::{
+    Call, CondTest, If, Include, Let, Lit, Loop, Match, Node, Target, Whitespace, Ws,
+};
+use parser::Parsed;
```
djc, line 10: Let's import `Expr` directly from `Parser` here?

djc (approved):

djc (approved):

Kijewski:
Thank you for your comments and reviewing! :)

PR #852, opened 2023-08-02
[WIP] parser benchmark & fuzzing

Kijewski:
Fuzzer:
```sh
cargo run --bin fuzz -r -- --progress
```

Benchmark:
```sh
cargo bench --bench min-size
```

WIP:
* not every node is implemented
* there are some errors, either in the parser or the random source generator


parser_benchmark/src/bin/parse.rs
```diff
@@ -0,0 +1,118 @@
+use std::fs::OpenOptions;
+use std::io::{stdin, stdout, Read, Write};
+use std::path::PathBuf;
+
+use gumdrop::Options;
+use parser::{Ast, ParseError, Syntax};
+
+#[derive(Debug, Options)]
+struct Args {
```
djc, line 9: I'd prefer to use clap, and move `Args` and `Error` types below `main`.

parser_benchmark/benches/min-size.rs
```diff
@@ -0,0 +1,50 @@
+use std::env::temp_dir;
```
djc, line 1: Can this just be in `askama_parser/benches` instead?

parser_benchmark/src/bin/fuzz.rs
```diff
@@ -0,0 +1,135 @@
+use std::env::temp_dir;
```
djc, line 1: I guess I'd prefer to use cargo-fuzz for this, at the top level of the repo. Any reason you didn't go that way?

Kijewski:
I could only get honggfuzz to run properly. If I activate instrumentation, libfuzzer always tells me

```
ERROR: The size of coverage PC tables does not match the
number of instrumented PCs. This might be a compiler bug,
please contact the libFuzzer developers.
Also check https://bugs.llvm.org/show_bug.cgi?id=34636
for possible workarounds (tl;dr: don't use the old GNU ld)
```

and I'm out of ideas. My normal link rustflags are `-Clinker=clang++-18 -Clink-arg=-fuse-ld=lld-18`, which might be too new. But downgrading to -17 or using `gold` did not help either.

I liked my homespun fuzzer better. :)

parser_benchmark/benches/min-size.rs
Kijewski: Done.

djc:
I think it's great to have a fuzzer but this is a lot of code to maintain. IMO we should have (a) a coverage-guided fuzzer and (b) it should work by taking random strings and trying to parse them, rather than painstakingly generating a whole bunch of combinations.

Kijewski:
The code is mostly meant as a benchmarking tool. I found a few errors in the parser when I ran the benchmark, so I thought it might be useful for fuzzing, too. Sorry, I know too little about the topic and I have too much on my plate already to be able to read deeper into it. IMO the fuzz folder could be rm'd, and I'd scale the PR down to only benchmarking.

djc:
Yes, I think that would probably make more sense.

djc:
Sorry, I guess I misunderstood how this PR was put together. The current state of the PR still retains all the code to randomly generate templates, which is what I was objecting to in a previous comment, so I don't want to merge this as-is.

(Sorry to be dismissive of all the work you've done on this PR.)

Kijewski:
Abandoned in favor of #1043.

Issue #860, opened 2023-09-08
Fuzzing askama_parser results in panic


manunio:
Hi, fuzzing `askama_parser` resulted in panic at following line.
https://github.com/djc/askama/blob/43e92aa3b6b9cd967a70bd0fd54d1f087d6ed76b/askama_parser/src/expr.rs#L290
I suppose it happens because crash input contains Cyrillic letters which are multi-byte and we need exact byte indices to avoid such panic in `[..]` notation.
```rust
#[test]
fn testing() {
    let d = "{eeuuu{b&{!!&{!!11{{
        0!(!1q҄א!)!!!!!!n!";
    if let Ok(_) = Ast::from_str(d, &Syntax::default()) {}
}
```
```
running 1 test
thread 'tests::testing' panicked at 'byte index 6 is not a char boundary; it is inside 'א' (bytes 5..7) of `!1q҄א!)!!!!!!n!`', askama_parser/src/expr.rs:290:22
stack backtrace:
```

djc:
Hah, figured that would be the next question. Yes, sounds good!

PR #862, opened 2023-09-11
fuzz: Add fuzz for askama_parser

manunio:
Hi, I would like to help integrate this project into [OSS-Fuzz](https://github.com/google/oss-fuzz).

- As an initial step for integration I have created this PR: https://github.com/google/oss-fuzz/pull/10935, it contains necessary logic from an OSS-Fuzz perspective to integrate askama.

- OSS-Fuzz is a free service run by Google that performs continuous fuzzing of important open source projects, you can find more details in our [FAQ](https://google.github.io/oss-fuzz/faq/).

Note:
 - This fuzz_target was able to find bug https://github.com/djc/askama/issues/860
 - I have limited input len to 500 as anything greater than that was resulting in stack-overflow which was crashing fuzz_target repeatedly.

### DIFF
--- a/askama_parser/fuzz/fuzz_targets/fuzz_parser.rs
+++ b/askama_parser/fuzz/fuzz_targets/fuzz_parser.rs
@@ -1,0 +1,13 @@
+#![no_main]
+use askama_parser::*;
+use libfuzzer_sys::fuzz_target;
+use std::str;
+
+fuzz_target!(|data: &[u8]| {
+    // fuzzed code goes here
+    if data.len() < 500 {
+        if let Ok(data) = str::from_utf8(data) {
+            if let Ok(_) = Ast::from_str(data, &Syntax::default()) {}
+        }
+    }
+});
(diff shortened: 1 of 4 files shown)


askama_parser/fuzz/fuzz_targets/fuzz_parser.rs
```diff
@@ -0,0 +1,13 @@
+#![no_main]
+use askama_parser::*;
+use libfuzzer_sys::fuzz_target;
+use std::str;
+
+fuzz_target!(|data: &[u8]| {
+    // fuzzed code goes here
+    if data.len() < 500 {
```
djc, line 8: How did you decide on the 500-byte boundary? I think it probably makes sense...

djc (approved):

askama_parser/fuzz/fuzz_targets/fuzz_parser.rs
djc: Ah sorry, you mentioned there was a stack overflow. That seems suspicious, maybe I'll have a look at that.

djc: I'll see what happens if I increase the limit locally.

PR #863, opened 2023-09-18
fuzz: remove input limit

manunio:
While working on https://github.com/djc/askama/pull/862 fuzz_parser was crashing and failing oss-fuzz build_checks. so a limit of 500 was placed, this pr removes that.


djc (commented):
Was there some work elsewhere done to enable this limit removal?

djc (approved):

djc:
I'm confused why the fuzzer is not causing stack overflows anymore, as you said it did while initially integrating (in #862).

djc:
I'm confused. Why are you proposing to remove the limit if doing so causes stack overflows?

PR #866, opened 2023-09-28
Limit expression nesting level to avoid stack overflows

djc:
This is triggered by the fuzzer, see also https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=62759.

### DIFF
--- a/askama_parser/src/expr.rs
+++ b/askama_parser/src/expr.rs
@@ -15,11 +15,12 @@ use super::{
 
 macro_rules! expr_prec_layer {
     ( $name:ident, $inner:ident, $op:expr ) => {
-        fn $name(i: &'a str) -> IResult<&'a str, Self> {
-            let (i, left) = Self::$inner(i)?;
+        fn $name(i: &'a str, level: Level) -> IResult<&'a str, Self> {
+            let level = level.nest(i)?;
+            let (i, left) = Self::$inner(i, level)?;
             let (i, right) = many0(pair(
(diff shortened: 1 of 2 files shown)


Kijewski:
Wouldn't same be needed for Nodes, too?

```jinja2
{%if true%}{%if true%}{%if true%}{%if true%}{%if true%}{%if true%}{%if true%}{%if true%}...
```

djc:
Probably? Was going to wait until the fuzzer complains.

askama_parser/src/expr.rs
```diff
@@ -15,11 +15,12 @@ use super::{
...
     ( $name:ident, $inner:ident, $op:expr ) => {
-        fn $name(i: &'a str) -> IResult<&'a str, Self> {
-            let (i, left) = Self::$inner(i)?;
+        fn $name(i: &'a str, level: Level) -> IResult<&'a str, Self> {
+            let level = level.nest(i)?;
+            let (i, left) = Self::$inner(i, level)?;
             let (i, right) = many0(pair(
                 ws(tag($op)),
-                Self::$inner,
+                |nested| Self::$inner(nested, level),
```
Kijewski, line 23: I was confused for a moment what `nested` was. I think the identifier `i` would be better. We use it throughout all the code and it is used in all of nom's examples for the input string.

djc: Fair enough, renamed them.

Kijewski (approved):

PR #868, opened 2023-09-28
Limit nesting in parser nodes, too

djc:
As expected. 👍


Kijewski:
For nodes the level could also be passed around in the state.

djc:
I guess? I'm not really motivated enough to make it happen, but open to if you want to take a whack at it.

Kijewski:
Implemented in #869.

PR #870, opened 2023-09-29
Clean up CI workflow & prepare for releases

djc:


djc:
Will also publish the first askama_parser release.

djc:
Published.

PR #875, opened 2023-10-04
Upgrade to comrak 0.19

djc:


djc:
Hmm, I'm just now noticing that the `markdown` filter's option argument makes comrak effectively part of our public API, such that we would technically need to give the askama crate a semver-incompatible bump every time we bump comrak. Might be an argument to just remove the options argument... which probably gets used fairly rarely?

Kijewski:
Hm, I did not think about that. Do you think it would be enough to re-export Comrak's Options and supporting structs?

djc:
No, that doesn't change anything because the versions of those types still change (that is, in one version of Askama you can pass a comrak 0.18 `Options` and in the next version it requires comrak 0.19 `Options`).

djc:
@Kijewski ping? I think we should just merge this, and then #920 is already lining up comrak 0.20.

Kijewski (approved):

PR #880, opened 2023-10-12
Allow macros to take `self` as an argument

PizzasBear:
Fixes the error ``no field `self` on type `&Templ` `` in the following example

```rust
#[derive(askama::Template)]
#[template(
    ext = "html",
    source = "
        {% macro m(x) %}
            {{ x.name }}
        {% endmacro %}

        {# this doesn't work #}
        {% call m(self) %}
    "
)]
struct Templ {
    name: String,
}

fn main() {
    println!(
        "{}",
        Templ {
            name: "Templ".into()
        }
    );
}
```

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -952,7 +952,7 @@ impl<'a> Generator<'a> {
                 // If `expr` is already a form of variable then
                 // don't reintroduce a new variable. This is
                 // to avoid moving non-copyable values.
-                Expr::Var(name) => {
+                &Expr::Var(name) if name != "self" => {
                     let var = self.locals.resolve_or_self(name);
                     self.locals.insert(arg, LocalMeta::with_ref(var));
                 }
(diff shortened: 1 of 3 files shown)


PizzasBear:
A temporary fix is to wrap `self` in parentheses, e.g.

```rust
#[derive(askama::Template)]
#[template(
    ext = "html",
    source = "
        {% macro m(x) %}
            {{ x.name }}
        {% endmacro %}

        {# this does work #}
        {% call m((self)) %}
    "
)]
struct Templ {
    name: String,
}

fn main() {
    println!(
        "{}",
        Templ {
            name: "Templ".into()
        }
    );
}
```

djc (approved):
Seems okay -- since we know `self` is a reference, we know we won't be moving a non-`Copy` value.

Could you maybe add a test for this?

PizzasBear:
I'm pretty new to pull requests and this repo, so I don't know how to add a test...

djc:
Basically add your example from the PR description in `testing/tests/macro.rs` and add a test function to make sure it has the expected output.

PizzasBear:
I've added the test

djc:
Thanks!

PR #886, opened 2023-10-23
Round angles around `endblock` and `endmacro`

GuillaumeGomez:
The documentation didn't mention it was possible and I didn't find many tests to enforce this so I decided to add them.

As a side-note, I'm planning to improve the reported error because it's really not great currently.


testing/tests/named_end.rs
```diff
@@ -0,0 +1,13 @@
+use askama::Template;
```
djc, line 1: Let's just put this in `inheritance.rs` rather than adding yet another file?

djc (approved):
LGTM, though I don't quite understand what "round angles" have to do with these changes!

GuillaumeGomez:
> LGTM, though I don't quite understand what "round angles" have to do with these changes!

That wasn't clear indeed. I couldn't find if it was possible to use it by looking at the tests so I had to look into the code. So "round angles" here meant "adding tests and docs".

testing/tests/named_end.rs
GuillaumeGomez: Don't you prefer to have a specific test for it? I'm not against it, just want to be sure before making the change.

GuillaumeGomez: Ok!

djc (approved):

PR #887, opened 2023-10-23
Improve error message for incompatible names in `endmacro` and `endblock`

GuillaumeGomez:
To achieve this, I needed to completely change how errors were handled and to create a new type (`ParsingError`). Thanks to this, we can now add custom error messages. I'm planning to iterate on this base to provide better error messages overall and potentially even suggestion (like when the parser encounters `elif` so it can suggest to use `else if` instead).

### DIFF
--- a/askama_parser/src/lib.rs
+++ b/askama_parser/src/lib.rs
@@ -101,7 +101,7 @@ impl<'a> Ast<'a> {
         let msg = format!(
             "{}problems parsing template source at row {}, column {} near:\n{}",
             if let Some(message) = message {
-                format!("error: {message}\n")
+                format!("{message}\n")
             } else {
                 String::new()
             },
--- a/askama_parser/src/lib.rs
+++ b/askama_parser/src/lib.rs
@@ -1,6 +1,7 @@
 #![deny(unreachable_pub)]
 #![deny(elided_lifetimes_in_paths)]
 
+use std::borrow::Cow;
 use std::cell::Cell;
 use std::{fmt, str};
 
@@ -9,7 +10,7 @@ use nom::bytes::complete::{escaped, is_not, tag, take_till};
 use nom::character::complete::char;
 use nom::character::complete::{anychar, digit1};
 use nom::combinator::{cut, eof, map, opt, recognize};
-use nom::error::ErrorKind;
+use nom::error::{Error, ErrorKind, FromExternalError};
 use nom::multi::many1;
 use nom::sequence::{delimited, pair, preceded, terminated, tuple};
 use nom::{error_position, AsChar, IResult, InputTakeAtPosition};
@@ -76,14 +77,16 @@ pub struct Ast<'a> {
 impl<'a> Ast<'a> {
     pub fn from_str(src: &'a str, syntax: &Syntax<'_>) -> Result<Self, ParseError> {
         let parse = |i: &'a str| Node::many(i, &State::new(syntax));
-        let err = match terminated(parse, cut(eof))(src) {
+        let (input, message) = match terminated(parse, cut(eof))(src) {
             Ok(("", nodes)) => return Ok(Self { nodes }),
             Ok(_) => unreachable!("eof() is not eof?"),
-            Err(nom::Err::Error(err)) | Err(nom::Err::Failure(err)) => err,
+            Err(
+                nom::Err::Error(ErrorContext { input, message, .. })
+                | nom::Err::Failure(ErrorContext { input, message, .. }),
+            ) => (input, message),
             Err(nom::Err::Incomplete(_)) => return Err(ParseError("parsing incomplete".into())),
         };
 
-        let nom::error::Error { input, .. } = err;
         let offset = src.len() - input.len();
         let (source_before, source_after) = src.split_at(offset);
 
@@ -96,7 +99,12 @@ impl<'a> Ast<'a> {
         let column = last_line.chars().count();
 
         let msg = format!(
-            "problems parsing template source at row {}, column {} near:\n{}",
+            "{}problems parsing template source at row {}, column {} near:\n{}",
+            if let Some(message) = message {
+                format!("error: {message}\n")
+            } else {
+                String::new()
+            },
             row + 1,
             column,
             source_after,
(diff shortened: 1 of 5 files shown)


askama_parser/src/error.rs
```diff
@@ -0,0 +1,46 @@
...
+    fn from_external_error(input: I, _kind: ErrorKind, e: E) -> Self {
+        Self {
+            input,
+            message: Some(Cow::Owned(e.to_string())),
+        }
+    }
+}
+
+impl<I> ParsingError<I> {
+    pub(crate) fn from_err(error: nom::Err<Error<I>>) -> nom::Err<Self> {
```
djc, line 33: Can this be a `From` or `Into` impl instead, or does that run into coherence issues?

askama_parser/src/error.rs
```diff
@@ -0,0 +1,46 @@
+use nom::error::{Error, ErrorKind, FromExternalError, ParseError};
+use std::borrow::Cow;
+
+#[derive(Debug)]
+pub(crate) struct ParsingError<I> {
```
djc, line 5: We have `ParseError` already. How does this relate to that, why do we need to keep both?

askama_parser/src/node.rs
```diff
@@ -635,6 +636,30 @@ pub struct BlockDef<'a> {
...
+                        "unexpected name `{end_name}` in `endblock` tag for unnamed `block`",
+                    ))),
+                }));
+            } else if !name.is_empty() && !end_name.is_empty() && name != end_name {
+                return Err(nom::Err::Failure(ParsingError {
+                    input: before,
+                    message: Some(Cow::Owned(format!(
+                        "expected name `{name}` in `endblock` tag, found `{end_name}`",
+                    ))),
+                }));
```
djc: I think we can deduplicate a bunch of wrapping here?

testing/tests/ui/name_mismatch_endblock.stderr
```diff
@@ -1,4 +1,5 @@
-error: problems parsing template source at row 1, column 27 near:
+error: error: expected name `foo` in `endblock` tag, found `not_foo`
```
djc, line 1: Don't love the `error: error:` here, can we avoid that?

askama_parser/src/node.rs
```diff
@@ -626,6 +632,31 @@ impl<'a> Match<'a> {
     }
 }
 
+fn check_end_name<'a>(
```
djc, line 635: Moving this should be in a separate commit from making changes to it to accomodate the second use case. IMO if you already know you're going to take it out of the parent function, better to keep it at the top-level in the introducing commit.

Please keep it ordered below the highest user.

djc (commented):
Great that you're working on better errors! I like the direction, here are some nits about the implementation.

testing/tests/ui/name_mismatch_endblock.stderr
GuillaumeGomez: Ah yes, completely missed that. I'll remove the extra `error:` I added.

askama_parser/src/error.rs
GuillaumeGomez: `ParserError` is the "general" error which always contains a message and doesn't contain any `nom` information. Well, I don't see any reason why not merging both so I'll modify this type and merge them.

askama_parser/src/error.rs
GuillaumeGomez: `From` forces to return `Self`, which is not the case here, hence why I didn't implement it.

askama_parser/src/node.rs
GuillaumeGomez: Ok, I'll move it out directly then.

askama_parser/src/error.rs
djc: Ah, that makes sense.

askama_parser/src/error.rs
djc: If we can merge them I think that would be good thing. I think maybe we could just name it `Error`. I think it should just live in `lib.rs` -- it's too small to get its own file for my taste.

askama_parser/src/node.rs
GuillaumeGomez: Indeed, good catch!

askama_parser/src/error.rs
GuillaumeGomez: So merging both types would actually not be a great idea because of the generic argument, which have no use for users but that they would have to "live with". I can move the type to `lib.rs` though if you prefer.

GuillaumeGomez:
Should be ready now.

askama_parser/src/error.rs
djc: Hmm, okay, so what's the difference between both of these types?

GuillaumeGomez: The public one only (and always) contains a string and is meant to be used for displaying an error whereas the private on contains `nom` information that we need at the end to generate `ParseError`'s message.

GuillaumeGomez: I'll add a comment on the type to make it clear.

GuillaumeGomez:
I moved `ParsingError` into `lib.rs` and added documentation on it.

askama_parser/src/lib.rs
```diff
@@ -131,6 +129,55 @@ impl fmt::Display for ParseError {
     }
 }
 
+/// This type is used to handle `nom` errors and in particular to add custom error messages.
+/// It used to generate `ParserError`.
+///
+/// It cannot be used to replace `ParseError` because it expects a generic, which would make
+/// `askama`'s users experience less good (since this generic is only needed for `nom`).
+#[derive(Debug)]
+pub(crate) struct ParsingError<I> {
```
djc, line 138: I'd like this commit squashed into the first commit (that introduces it), and I still think it should be renamed? I'm thinking maybe `ErrorContext` as I think this actually represents Askama-specific context that's spliced into nom errors in order to propagate some extra state across nom parser combinators.

askama_parser/src/node.rs
```diff
@@ -667,6 +674,26 @@ impl<'a> BlockDef<'a> {
...
+    before: &'a str,
+    after: &'a str,
+    name: &'a str,
+    end_name: &'a str,
+    kind: &str,
+) -> IResult<&'a str, &'a str, ParsingError<&'a str>> {
+    let message = if name.is_empty() && !end_name.is_empty() {
```
djc: I think we should move the simple case up front:

```
if name == end_name {
    return Ok((after, end_name));
}
```

Then we can simplify the subsequent `if` because we've already yeeted out the case where both `name` and `end_name` are empty.

Also, please ditch the trailing commas in single-line `format!()` calls.

askama_parser/src/node.rs
```diff
@@ -510,6 +516,26 @@ impl<'a> Macro<'a> {
     }
 }
 
+fn check_end_name<'a>(
```
djc, line 519: This shouldn't move in the second commit.

djc (commented):
Some nits left, thanks for working on this!

askama_parser/src/lib.rs
GuillaumeGomez: `ErrorContext` sounds good to me.

askama_parser/src/node.rs
GuillaumeGomez: You told me to put it just after the highest user. Since the highest user changed, I moved it. I'm gonna revert this move then.

GuillaumeGomez:
Updated!

askama_parser/src/node.rs
GuillaumeGomez: Looks better indeed, good catch!

djc (approved):

askama_parser/src/node.rs
djc: You're strictly correct -- I'd have added it in the first commit in the place where it would be after the second user introduced it, balancing the concerns of (a) clean commits that don't introduce unnecessary changes within a PR, and (b) a clean ordering at the end of the PR.

djc:
I wonder if we should bring in a type alias for `IResult<&'a str, Self, ErrorContext<&'a str>>`...

GuillaumeGomez:
That's exactly what we're talking about in #889. :)

Issue #488, opened 2021-05-25
[feature] Add "embed" block


YannikSc:
Having an "embed" block like [twig has it](https://twig.symfony.com/doc/2.x/tags/embed.html), which is basically an "extend" but without an extra file, would be an awesome addition to the library I think.

Is it in the current implementiation possible to get something like this working (or is there maybe even something like this already)?

This would be very useful if a macro is just to small or you want to get more markup replaced in a general structure (like a modal).

vallentin:
> which is basically an "extend" but without an extra file

What do you mean by "without an extra file"? Unless I'm misinterpreting something from the docs you linked, then you still need all your (partial) templates in separate files. So with `embed` you're able to `include` a separate template and without the `embed` scope you're able to override the `block`s.

I like it and can personally see many uses. Just want to make sure I didn't miss anything.

djc:
I'm open to it in principle as long as it doesn't add too much complexity in the code generator; if it can reuse most of the existing infrastructure for extends/block, I'm fine with adding another interface to it.

YannikSc:
"without an extra file" means, that you don't need an extra file to extend the base template. Normally when you want to extend something, you use 3 files. For the parent and child template, and the template, where you include it. With this embed you only have 2. The parent and the including template which is doing what the child template is doing but inside of the embed.

I'm just describing it badly. The linked documentation however exaclty describes, what I'm trying to explain.

vallentin:
I'm still not sure I understand the "without an extra file", maybe I'm tired, I don't know. :)

If we go with your original example, you have a `page` template and a `modal` template. If the `modal` template contains a macro, then you need to `import` and `call`. If it instead is a "sub-page", then we instead `embed` it. But overall, we still have 2 files, right?

Even if we're talking about complex template inheritance. Say you have a `base` template, a `page` template, then an `error` template. They all extend each other, and you finally have a `not-found` and `internal-server-error` templates, that both extend `error`. Maybe instead of extending `error` you could extend `page`, and then `embed` `error`. But once again, that doesn't decrease the amount of files?

I'm clearly missing something. :)

Regardless, I still think `embed` is nice and useful, in relation to replacing e.g. `macro`s with `embed`, as I have quite a few of those myself. :)

YannikSc:
Okay then; here the use case I had:

Building a `modal.html` with the structure for a model: header, body and footer, controls etc.
[...]

PR #888, opened 2023-10-24
feature: Add embed node

YannikSc:
In memory of my old issue/feature request in https://github.com/djc/askama/issues/488, here is finally the implementation of it. Tests are included of cause.

~~I have (not yet) updated the documentation for it. If this feature and implementation is fine, I'm open for updating the documentation to describe the new possibilities.~~

Edit: I did update the documentation now, as I think it might be easier to understand, what it's actually doing.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -4,7 +4,7 @@ use crate::input::{Print, Source, TemplateInput};
 use crate::CompileError;
 
 use parser::node::{
-    Call, Comment, CondTest, If, Include, Let, Lit, Loop, Match, Target, Whitespace, Ws,
+    Call, Comment, CondTest, Embed, If, Include, Let, Lit, Loop, Match, Target, Whitespace, Ws,
 };
 use parser::{Expr, Node, Parsed};
 use proc_macro::TokenStream;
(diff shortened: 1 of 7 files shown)


askama_derive/src/generator.rs
```diff
@@ -1053,6 +1074,33 @@ impl<'a> Generator<'a> {
...
+        embed: &'a Embed<'_>,
+    ) -> Result<usize, CompileError> {
+        self.flush_ws(embed.ws1);
+        self.write_buf_writable(buf)?;
+        let embed_path = self
+            .input
+            .config
+            .find_template(embed.path, Some(&self.input.path))?;
+        let mut embedded_context =
+            Context::new(self.input.config, &self.input.path, embed.nodes.as_slice())?;
```
wrapperup, line 1089: Probably makes sense to add an additional parameter to set whether or not this context starts at the top level, just so we get an error.

Adding an `{% extends ... %}` (or any top-level only node) inside the embed node is currently valid, even though it doesn't end up doing anything.

wrapperup (changes requested):
This is a really cool (and useful!) feature, it's sort of like components in a way. Had a few thoughts and nits.

askama_derive/src/generator.rs
wrapperup, line 1089: There's missing errors for doing anything outside block tags. Some of these don't actually do anything (extends for example gets overwritten), but you're allowed to write macro and include nodes here, which seems weird.

askama_derive/src/generator.rs
```diff
@@ -1053,6 +1074,33 @@ impl<'a> Generator<'a> {
...
+
+        let mut generator = Generator::new(
+            self.input,
+            self.contexts,
+            Some(&heritage),
+            MapChain::new(),
+            self.whitespace,
+        );
+        self.prepare_ws(embed.ws2);
+        generator.handle(heritage.root, heritage.root.nodes, buf, AstLevel::Top)
```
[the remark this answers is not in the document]
wrapperup: Basically this would become valid
```django
{{ my_context_value }}
{% let x = 10 %}

{% embed "embed.html" %}
      {% block content %}
            {{ my_context_value }}
            {{ x }}
      {% endblock %}
{% endembed %}
```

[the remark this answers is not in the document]
YannikSc: This sounds very cool. I did not intend to drop the parent context here, so I will apply this to the code!

askama_derive/src/generator.rs
YannikSc: Yes, I see why this is problematic, but it's consistent. I think this is something, that should be fixed/improved in general.

Another real problem, that I mentioned in the issue (#488) is, that the compiler is not including the file when the `embed` is inside a `block` in an `extend`ed file. This problem however also exists for `include`s and I'm not sure how to approach this.

PR #889, opened 2023-10-25
Improve `elif` error

GuillaumeGomez:
It's based on https://github.com/djc/askama/pull/887 (only the two last commits are from this PR).

When switching from another jinja-like framework, it's frequent to forget keywords like `elif`, so instead of just throwing a generic error, it also suggests `else if`.

### DIFF
--- a/askama_parser/src/node.rs
+++ b/askama_parser/src/node.rs
@@ -299,7 +299,15 @@ impl<'a> Cond<'a> {
         let mut p = tuple((
             |i| s.tag_block_start(i),
             opt(Whitespace::parse),
-            ws(keyword("else")),
+            ws(alt((keyword("else"), |i| {
+                let _ = keyword("elif")(i)?;
+                Err(nom::Err::Failure(ErrorContext {
+                    input: i,
+                    message: Some(Cow::Borrowed(
+                        "unknown `elif` keyword; did you mean `else if`?",
+                    )),
+                }))
+            }))),
             cut(tuple((
                 opt(|i| CondTest::parse(i, s)),
                 opt(Whitespace::parse),
--- a/testing/tests/ui/elif.rs
+++ b/testing/tests/ui/elif.rs
@@ -1,0 +1,8 @@
+use askama::Template;
+
+#[derive(Template)]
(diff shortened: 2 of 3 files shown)


askama_parser/src/node.rs
```diff
@@ -299,7 +299,15 @@ impl<'a> Cond<'a> {
         let mut p = tuple((
             |i| s.tag_block_start(i),
             opt(Whitespace::parse),
-            ws(keyword("else")),
+            ws(alt((keyword("else"), |i| {
+                let _ = keyword("elif")(i)?;
+                Err(nom::Err::Failure(ParsingError {
+                    input: i,
+                    message: Some(Cow::Borrowed(
+                        "Unknown `elif` keyword. Did you mean `else if`?",
```
djc, line 307: I think we should not capitalize `Unknown` here, since this is preceded by `error: ` anyway. Also don't love the `.` here, maybe `;` or `,`?

GuillaumeGomez: Let's go with `;`.

GuillaumeGomez:
Updated the message (and the first commits too).

Kijewski:
This is great, thank you for working on making errors more readable!

`I` in `ErrorContext<I>` is always `&'a str`, isn't it? Maybe you could replace the generic argument with a simple lifetime, then?

Maybe `IResult<&'a str, T, ErrorContext<'a>>` could then be type aliased to

```rust
/// Parser result
PResult<'a, T = &'a str> = IResult<&'a str, T, ErrorContext<'a>>;
```

GuillaumeGomez:
I don't know what the future of `askama` is going to be so I'd rather keep using `T` and not specialize it. However I can create a type alias if you want?

djc:
Ha, just posted very similar ideas about a type alias in #887 before I saw the discussion here. Great minds...

I think we're quite unlikely to change the type from `&'a str` so I think specializing would make sense.

GuillaumeGomez:
Do you want me to do it in this PR or should I open a new one?

djc (approved):

PR #890, opened 2023-10-25
Create a type alias for all `IResult` to simplify code reading

GuillaumeGomez:
As discussed in #889.

### DIFF
--- a/askama_parser/src/lib.rs
+++ b/askama_parser/src/lib.rs
@@ -13,7 +13,7 @@ use nom::combinator::{cut, eof, map, opt, recognize};
 use nom::error::{Error, ErrorKind, FromExternalError};
 use nom::multi::many1;
 use nom::sequence::{delimited, pair, preceded, terminated, tuple};
-use nom::{error_position, AsChar, IResult, InputTakeAtPosition};
+use nom::{error_position, AsChar, InputTakeAtPosition};
 
 pub mod expr;
 pub use expr::Expr;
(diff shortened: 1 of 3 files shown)


askama_parser/src/lib.rs
```diff
@@ -178,6 +178,8 @@ impl<I> ErrorContext<I> {
     }
 }
 
+pub(crate) type PResult<'a, T = &'a str> = IResult<&'a str, T, ErrorContext<'a>>;
```
djc, line 181: I'd like this defined before the `ErrorContext` type (since it relies on it), renamed to `ParseResult`, and inline the `nom::IResult` alias (that is, `ParseResult` should directly alias to something like `Result<(&'a str, T), nom::Err<ErrorContext<'a>>>`).

djc (commented):
This is great, thanks!

askama_parser/src/lib.rs
GuillaumeGomez: So just to sure I understood: you want `PResult` to be renamed to `ParseResult` and un-alias `IResult` and move it above `ErrorContext` (as I'm writing it, I finally were able to understand your message haha).

djc: I want `pub(crate) type ParseResult<'a, T = &'a str> = Result<(&'a str, T), nom::Err<ErrorContext<'a>>;` to be defined before the `struct ErrorContext` (restating because I'm not sure I understand "un-alias `IResult`").

GuillaumeGomez: Yes that's what I meant, replace the `IResult` alias by the aliased content (hence "un-alias", not sure if the word exists though ^^').

GuillaumeGomez:
Applied suggestions.

djc (approved):

Issue #885, opened 2023-10-21
Allow to have named arguments in macro


GuillaumeGomez:
It'd pretty nice to be able to have named arguments. For example:

```
{% macro menu_link(href, text, target) %}
...
{% endmacro menu_link %}
```

Then to be able to call it like this:

```
{% menu_link(href="/something", text="link", target="_blank") %}
```

It makes the code much better when reading it.

As a first step, only allowing it for macros would be nice, but potentially extending it to filters would be nice too.

djc:
I worry that this would be a little foreign to users -- that is, I'm not sure it passes the design test of "it should be obvious how this gets compiled to Rust code".

GuillaumeGomez:
Does it really matter? If this is a big enough concern, we can add a new entry in the askama book explaining how it is generated.

djc:
It matters to me -- I think an important design aspect of the templating language for Askama is that people can reason about how their template code gets translated into Rust code. Why don't you write the paragraph that explains this first and we can see how that looks?

GuillaumeGomez:
Sure. I'll do that (hopefully) soon.

GuillaumeGomez:
So now that #893 is merged, let's come back to this. :laughing: 

So I had in mind to keep a `Vec<String>` for each macro where each element is a macro argument name. Then when we encounter a macro with named arguments (should we allw to have a mix between named and not named?), we store the expression into a variable in the same order as declared in the template (so the "computation order" remains the same) and then pass the variables in the correct order to the macro.

Does this sound like an acceptable approach?

PR #893, opened 2023-10-26
Add new chapter in the askama book about template expansion

GuillaumeGomez:
As discussed in https://github.com/djc/askama/issues/885.

### DIFF
--- a/book/src/template_expansion.md
+++ b/book/src/template_expansion.md
@@ -1,0 +1,485 @@
+# Template Expansion
+
+This chapter will explain how the different parts of the templates are
+translated into Rust code.
+
+⚠️ Please note that the generated code might change in the future so the
+following examples might not be up-to-date.
+
+## Basic explanations
+
+Whwn you add `#[derive(Template)]` and `#[template(...)]` on your type, the
+`Template` derive proc-macro will then generate an implementation of the
+`askama::Template` trait which will be a Rust version of the template.
+
+It will also implement the `std::fmt::Display` trait on your type which will
+internally call the `askama::Template` trait.
+
+Let's take a small example:
+
+```rust
+#[derive(Template)]
+#[template(source = "{% set x = 12 %}", ext = "html")]
+struct Mine;
+```
+
+will generate:
+
+```rust
+impl ::askama::Template for YourType {
+    fn render_into(
+        &self,
+        writer: &mut (impl ::std::fmt::Write + ?Sized),
+    ) -> ::askama::Result<()> {
+        let x = 12;
+        ::askama::Result::Ok(())
+    }
+    const EXTENSION: ::std::option::Option<&'static ::std::primitive::str> = Some(
+        "html",
+    );
+    const SIZE_HINT: ::std::primitive::usize = 0;
+    const MIME_TYPE: &'static ::std::primitive::str = "text/html; charset=utf-8";
+}
+
+impl ::std::fmt::Display for YourType {
+    #[inline]
+    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+        ::askama::Template::render_into(self, f).map_err(|_| ::std::fmt::Error {})
+    }
+}
+```
+
+For simplicity, we will only keep the content of the `askama::Template::render_into`
+function from now on.
+
+## Text content
+
+If you have "text content" (for example HTML) in your template:
+
+```html
+<h1>{{ title }}</h1>
+```
+
+It will generate it like this:
+
+```rust
+writer
+    .write_fmt(
+        format_args!(
+            "<h1>{0}</h1>",
+            &::askama::MarkupDisplay::new_unsafe(&(self.title), ::askama::Html),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
+About `MarkupDisplay`: we need to use this type in order to prevent generating
+invalid HTML. Let's take an example: if `title` is `"<a>"` and we display it as
+is, in the generated HTML, you won't see `<a>` but instead a new HTML element
+will be created. To prevent this, we need to escape some characters.
+
+In this example, `<a>` will become `&lt;a&gt;`. And this is why there is the
+`safe` builtin filter, in case you want it to be displayed as is.
+
+## Variables
+
+### Variables creation
+
+If you create a variable in your template, it will be created in the generated
+Rust code as well. For example:
+
+```jinja
+{% set x = 12 %}
+{% let y = x + 1 %}
+```
+
+will generate:
+
+```rust
+let x = 12;
+let y = x + 1;
+::askama::Result::Ok(())
+```
+
+### Variables usage
+
+By default, variables will reference a field from the type on which the `askama::Template`
+trait is implemented:
+
+```jinja
+{{ y }}
+```
+
+This template will expand as follows:
+
+```rust
+writer
+    .write_fmt(
+        format_args!(
+            "{0}",
+            &::askama::MarkupDisplay::new_unsafe(&(self.y), ::askama::Html),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
+This is why if the variable is undefined, it won't work with Askama and why
+we can't check if a variable is defined or not.
+
+You can still access constants and statics by using paths. Let's say you have in
+your Rust code:
+
+```rust
+const FOO: u32 = 0;
+```
+
+Then you can use them in your template by referring to them with a path:
+
+```jinja
+{{ crate::FOO }}{{ super::FOO }}{{ self::FOO }}
+```
+
+It will generate:
+
+```rust
+writer
+    .write_fmt(
+        format_args!(
+            "{0}{1}{2}",
+            &::askama::MarkupDisplay::new_unsafe(&(crate::FOO), ::askama::Html),
+            &::askama::MarkupDisplay::new_unsafe(&(super::FOO), ::askama::Html),
+            &::askama::MarkupDisplay::new_unsafe(&(self::FOO), ::askama::Html),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
+(Note: `crate::` is to get an item at the root level of the crate, `super::` is
+to get an item in the parent module and `self::` is to get an item in the
+current module.)
+
+You can also access items from the type that implements `Template` as well using
+as `Self::`, it'll use the same logic.
+
+## Control blocks
+
+### if/else
+
+The generated code can be more complex than expected, as seen with `if`/`else`
+conditions:
+
+```jinja
+{% if x == "a" %}
+gateau
+{% else %}
+tarte
+{% endif %}
+```
+
+It will generate:
+
+```rust
+if *(&(self.x == "a") as &bool) {
+    writer.write_str("gateau")?;
+} else {
+    writer.write_str("tarte")?;
+}
+::askama::Result::Ok(())
+```
+
+Very much as expected except for the `&(self.x == "a") as &bool`. Now about why
+the `as &bool` is needed:
+
+The following syntax `*(&(...) as &bool)` is used to  trigger Rust's automatic
+dereferencing, to coerce e.g. `&&&&&bool` to `bool`. First `&(...) as &bool`
+coerces e.g. `&&&bool` to `&bool`. Then `*(&bool)` finally dereferences it to
+`bool`.
+
+In short, it allows to fallback to a boolean as much as possible, but it also
+explains why you can't do:
+
+```jinja
+{% set x = "a" %}
+{% if x %}
+    {{ x }}
+{% endif %}
+```
+
+Because it fail to compile because:
+
+```console
+error[E0605]: non-primitive cast: `&&str` as `&bool`
+```
+
+### if let
+
+```jinja
+{% if let Some(x) = x %}
+    {{ x }}
+{% endif %}
+```
+
+will generate:
+
+```rust
+if let Some(x) = &(self.x) {
+    writer
+        .write_fmt(
+            format_args!(
+                "{0}",
+                &::askama::MarkupDisplay::new_unsafe(&(x), ::askama::Html),
+            ),
+        )?;
+}
+```
+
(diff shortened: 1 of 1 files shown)


book/src/template_expansion.md
```diff
@@ -0,0 +1,387 @@
...
+```rust
+impl ::askama::Template for YourType {
+    fn render_into(
+        &self,
+        writer: &mut (impl ::std::fmt::Write + ?Sized),
+    ) -> ::askama::Result<()> {
+        let x = 12;
+        ::askama::Result::Ok(())
+    }
+    // I skipped some internal `askama` impl items to keep this as simple
```
djc, line 34: I would probably retain these extra items at least once to give people a better sense of what's going on. It's not all that much, right?

book/src/template_expansion.md
```diff
@@ -0,0 +1,387 @@
...
+writer
+    .write_fmt(
+        format_args!(
+            "<h1>{0}</h1>",
+            &::askama::MarkupDisplay::new_unsafe(&(self.title), ::askama::Html),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
```
djc, line 69: Maybe this can add like one sentence to suggest why the `MarkupDisplay` dance is useful/necessary?

book/src/template_expansion.md
```diff
@@ -0,0 +1,387 @@
...
+        format_args!(
+            "{0}{1}",
+            &::askama::MarkupDisplay::new_unsafe(&(crate::FOO), ::askama::Html),
+            &::askama::MarkupDisplay::new_unsafe(&(super::FOO), ::askama::Html),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
+You can also access items from your current type as well using `Self::`, it'll
```
djc, line 142: What about `self::FOO` for a `const` in the same module?

book/src/template_expansion.md
```diff
@@ -0,0 +1,387 @@
...
+```rust
+if *(&(self.x == "a") as &bool) {
+    writer.write_str("gateau")?;
+} else {
+    writer.write_str("tarte")?;
+}
+::askama::Result::Ok(())
+```
+
+Very much as expected except for the `&(self.x == "a") as &bool`. It allows to
```
djc, line 170: It would be nice to provide a little explanation of why this is generated like this... I don't remember but `git blame` could probably tell us.

book/src/template_expansion.md
```diff
@@ -0,0 +1,387 @@
...
+                    &::askama::MarkupDisplay::new_unsafe(&(user), ::askama::Html),
+                ),
+            )?;
+    }
+    if !_did_loop {}
+}
+::askama::Result::Ok(())
+```
+
+Interestingly enough, there is a `_did_loop` variable. It is used in case your
```
djc, line 238: This seems to conflate two different concepts:

* `if`-based loop filters
* The `else` branch, which should presumably also work for unfiltered iterators that are empty?

(We should arguably not generate the `if !did_loop {}` -- and maybe even if the `did_loop` variable -- if the `for` block didn't have an `else` clause?)

book/src/template_expansion.md
```diff
@@ -0,0 +1,387 @@
...
+            &::askama::MarkupDisplay::new_unsafe(
+                &(::askama::filters::abs(-2)?),
+                ::askama::Html,
+            ),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
+The filter is called with `-2` as first argument. If you add arguments to the
```
djc, line 309: The "If" in "If you add" doesn't really lead anywhere? Maybe rewrite as "You can add further arguments to the call like this:".

book/src/template_expansion.md
```diff
@@ -0,0 +1,387 @@
...
+                    &(::askama::filters::indent("a", 4)?),
+                )?),
+                ::askama::Html,
+            ),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
+As expected, `capitalize` first argument is the value returned by the `indent` call.
```
djc, line 356: This is missing a possessive something, maybe "`capitalize`'s"?

djc (commented):
I didn't actualy mean for you to write this, just the addition that would explain the new addition -- but it's great to have this! There should probably be a disclaimer about a lack of guarantees about the exact output because we might not keep this exactly up to date.

book/src/template_expansion.md
GuillaumeGomez: It's only some associated constants, but I don't think they bring much value. I can put them back though if you think they're useful to see.

book/src/template_expansion.md
GuillaumeGomez: Good idea.

book/src/template_expansion.md
GuillaumeGomez: I assumed that people would understand it but I can add it. :)

book/src/template_expansion.md
GuillaumeGomez: I was planning to send a follow-up to clean things up. :)

book/src/template_expansion.md
GuillaumeGomez: Good catch!

book/src/template_expansion.md
GuillaumeGomez: Outch indeed.

book/src/template_expansion.md
GuillaumeGomez: Gonna look around.

book/src/template_expansion.md
djc: Yes, I think it's good to give the full output once for the most minimal template.

book/src/template_expansion.md
GuillaumeGomez: > The `else` branch, which should presumably also work for unfiltered iterators that are empty?

Do you have an example by any chance?

GuillaumeGomez: > > The `else` branch, which should presumably also work for unfiltered iterators that are empty?
> 
> Do you have an example by any chance?

Ping @djc :)

djc: You wrote "It is used in case your `for` loop has a `if` condition" (I think that needs to be "an `if` condition), but I expect that we would hit the `else` branch even if no `if` condition is attached to the loop, but the iterator is just empty (like `&[]`).

GuillaumeGomez: Oh I didn't realize that could have a `else` branch without the `if`. Interesting. I will also need to update the `_did_loop` condition since it changed since #896.

djc: I'm actually not sure if you can, but I would expect them to be orthogonal.

GuillaumeGomez: I confirm it works with just the `else`. It's quite crazy how many things I didn't know about askama. ^^'

GuillaumeGomez:
Updated!

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+impl ::std::fmt::Display for YourType {
+    #[inline]
+    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+        ::askama::Template::render_into(self, f).map_err(|_| ::std::fmt::Error {})
+    }
+}
+```
+
+For simplicity, we will now only keep the content of the `askama::Template::render_into`
+function from now on.
```
djc, line 53: "now only .. from now on" feels like duplicated content, do you mind fixing that?

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+    .write_fmt(
+        format_args!(
+            "<h1>{0}</h1>",
+            &::askama::MarkupDisplay::new_unsafe(&(self.title), ::askama::Html),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
+About `MarkupDisplay`: we need to use this trait in order to prevent generating
```
djc, line 76: `MarkupDisplay` is actually a type, not a trait.

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+
+```rust
+let x = 12;
+let y = x + 1;
+::askama::Result::Ok(())
+```
+
+### Variables usage
+
+By default, variables will be fields from the type on which the `askama::Template`
```
djc, line 106: "will be fields from the type" -> "will reference a field from the type"

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+trait is implemented:
+
+```jinja
+{{ y }}
+```
+
+will generate:
```
djc, line 113: This reads a little awkwardly, you have "implemented: [[ example ]] will generate: [[ generated code ]]". I'd probably expand the "will generate" to something more like "This template will expand as follows:".

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+    .write_fmt(
+        format_args!(
+            "{0}",
+            &::askama::MarkupDisplay::new_unsafe(&(self.y), ::askama::Html),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
+This is why if the variable is undefined, it won't work with askama and why
```
djc, line 126: Please capitalize "Askama".

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+we can't check if a variable is defined or not.
+
+However, you can still access constants and statics by using paths. Let's say
```
djc, line 129: I'm not sure what "However" is trying to convey here -- there doesn't seem to be much of a juxtaposition between the variable handling and `const`/`static` support?

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+            "{0}{1}{2}",
+            &::askama::MarkupDisplay::new_unsafe(&(crate::FOO), ::askama::Html),
+            &::askama::MarkupDisplay::new_unsafe(&(super::FOO), ::askama::Html),
+            &::askama::MarkupDisplay::new_unsafe(&(self::FOO), ::askama::Html),
+        ),
+    )?;
+::askama::Result::Ok(())
+```
+
+Quite reminder: `crate::` is to get an item at the root level of the crate,
```
djc, line 157: "Quite" is probably not the right word here? I'd replace "Quite reminder:" with "Note:" and parenthesize this sentence.

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+`super::` is to get an item in the parent module and `self::` is to get an item
+in the current module.
+
+You can also access items from your current type as well using `Self::`, it'll
```
djc, line 161: Don't love "your current type", I like to call them the "context type", maybe you can introduce that term earlier when you talk about "type that implements `Template`"?

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+use the same logic.
+
+## Control blocks
+
+### if/else
+
+There is an interesting twist for `if/else` conditions. Let's check this code:
```
djc, line 168: "interesting twist" is a little vague. I'd either skip that and be more generic to say "Let's look at code generation for `if`/`else` conditions:" or say something like "The generated code can be more complex than expected, as seen with `if`/`else` conditions:".

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+
+Very much as expected except for the `&(self.x == "a") as &bool`. Now about why
+the `as &bool` is needed:
+
+The following syntax `*(&(...) as &bool)` is used to  trigger Rust's automatic
+dereferencing, to coerce e.g. `&&&&&bool` to `bool`. First `&(...) as &bool`
+coerces e.g. `&&&bool` to `&bool`. Then `*(&bool)` finally dereferences it to
+`bool`.
+
+In short, tt allows to fallback to a boolean as much as possible, but it also
```
djc, line 197: "tt" should probably just be "it"?

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+
+```jinja
+{% for user in users if users.len() > 2 %}
+    {{ user }}
+{% else %}
+    {{ x }}
+{% endfor %}
+```
+
+Which generates:
```
djc, line 317: For consistency, should probably avoid capitalizing "Which" here.

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+                ),
+            )?;
+    }
+}
+::askama::Result::Ok(())
+```
+
+## Filters
+
+Example by using the `abs` built-in filter:
```
djc, line 384: "Example by" -> "Example of".

book/src/template_expansion.md
```diff
@@ -0,0 +1,484 @@
...
+
+```html
+{% macro heading(arg) %}
+<h1>{{arg}}</h1>
+{% endmacro %}
+
+{% call heading("title") %}
+```
+
+Generates:
```
djc, line 467: Lowercase "Generates"?

djc (commented):
I still have a bunch of small nits...

book/src/template_expansion.md
GuillaumeGomez: :+1:

book/src/template_expansion.md
GuillaumeGomez: Oh indeed. Fixing that!

book/src/template_expansion.md
GuillaumeGomez: Removing it then.

book/src/template_expansion.md
GuillaumeGomez: Big typo, I meant to write "Quick". I'll apply your suggestion.

book/src/template_expansion.md
GuillaumeGomez: typo ^^'

GuillaumeGomez:
Updated!

djc (approved):

PR #896, opened 2023-10-30
Remove unused `_did_loop` condition if `else` block is empty

GuillaumeGomez:
As we saw in https://github.com/djc/askama/pull/893, there was room for improvement so I removed the `_did_loop` variable if the `else` block is empty. (which means I'll also need to update #893 ^^')


djc (approved):
Thanks!

PR #898, opened 2023-11-04
Add documentation for function syntax in templates

Regaez:
I was wanting to use functions in my own project templates and was happy to discover that PR https://github.com/djc/askama/pull/304 added support for them. I noticed that the documentation for this template syntax was still missing, so I added some details to the book to make it more accessible for others in the future.

This is based on the PR description of functionality added in https://github.com/djc/askama/pull/304 and my own understanding of how it all works after using Askama for my own projects. Please correct me if I have written anything inaccurate.


djc (approved):

djc:
This is great, thanks!

PR #901, opened 2023-11-07
Template expansion

GuillaumeGomez:
Follow-up of #893.

I was so focused on the content that I actually forgot to make `mdbook` link to it... Sorry about that.


djc (approved):

djc:
No problem, thanks for catching it!

PR #905, opened 2023-11-13
Derive Clone for Expr and Target

couchand:
When working with an `askama` AST I frequently find myself wishing for these `Clone` implementations.

c.f. the Rust API guidelines [C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits)


djc (approved):

djc:
Thanks!

Issue #882, opened 2023-10-17
`num as {integer}` cast does not work?


realGWM:
Hi, I have this MRE:

```
use askama::Template;

#[derive(Template)]
#[template(ext = "txt", source = r#"
    {% let a = 51 %} // works
    {% let a = 51 as i64 %} // does not work
"#)]
struct HelloTemplate {}

fn main() {
    let a = 51 as i64; // sanity check that it works in normal Rust code
    let hello = HelloTemplate {};
    println!("{}", hello.render().unwrap());
}
```

and it doesn't compile because of the line with casting `as i64`.
Is there some alternative syntax for it that I'm missing?

djc:
It's possible we don't support `as` casts. Maybe `51_i64` is supported? I forget...

djc:
Okay, well, if you want to try hacking it in, the code for binary expressions is around here:

https://github.com/djc/askama/blob/main/askama_parser/src/expr.rs#L107

If you can submit a patch with a basic test I'm happy to review it and/or provide more guidance.

Kijewski:
In the meantime using `i64::from(51)` should work.

GuillaumeGomez:
You can now declare floats and integers with their suffix (since #908).

PR #908, opened 2023-11-17
Add better support for rust-like number literals

PizzasBear:
Added support for underscore separators, scientific notation and suffixes (e.g. `usize` or `f64`) for numbers.
Alleviates issue #882 for number literals.

### DIFF
--- a/askama_parser/src/lib.rs
+++ b/askama_parser/src/lib.rs
@@ -248,26 +248,8 @@ fn bool_lit(i: &str) -> ParseResult<'_> {
     alt((keyword("false"), keyword("true")))(i)
 }
 
-/// Underscore separateed digits of the given base, unless `must_start_with_digit` this may start
-/// with an underscore.
-fn separated_digits(radix: u32, must_start_with_digit: bool) -> impl Fn(&str) -> ParseResult<'_> {
-    move |i| {
-        recognize(tuple((
-            |i| {
-                if must_start_with_digit {
-                    Ok((i, 0))
-                } else {
-                    many0_count(char('_'))(i)
-                }
-            },
-            satisfy(|ch| ch.is_digit(radix)),
-            many0_count(satisfy(|ch| ch == '_' || ch.is_digit(radix))),
-        )))(i)
-    }
-}
-
 fn num_lit(i: &str) -> ParseResult<'_> {
-    let integer_suffix_tags = |i| {
+    let integer_suffix = |i| {
         alt((
             tag("i8"),
             tag("i16"),
@@ -283,7 +265,7 @@ fn num_lit(i: &str) -> ParseResult<'_> {
             tag("usize"),
         ))(i)
     };
-    let float_suffix_tags = |i| alt((tag("f32"), tag("f64")))(i);
+    let float_suffix = |i| alt((tag("f32"), tag("f64")))(i);
 
     recognize(tuple((
         opt(char('-')),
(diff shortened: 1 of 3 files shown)


askama_parser/src/lib.rs
```diff
@@ -252,8 +251,88 @@ fn bool_lit(i: &str) -> ParseResult<'_> {
 fn num_lit(i: &str) -> ParseResult<'_> {
     recognize(tuple((
         opt(char('-')),
-        digit1,
-        opt(pair(char('.'), digit1)),
+        // digit1,
```
djc, line 254: Looks like there's a bunch of duplication here, which I'd like to avoid. I'm guessing this will require some named parsers, which would probably aid in making this code easier to understand. Also please remove the commented out version of the old code.

PizzasBear:
I've tried to reduce the duplication, and removed the commented-out code.

askama_parser/src/lib.rs
```diff
@@ -249,11 +248,75 @@ fn bool_lit(i: &str) -> ParseResult<'_> {
     alt((keyword("false"), keyword("true")))(i)
 }
 
+/// Underscore separateed digits of the given base, unless `must_start_with_digit` this may start
+/// with an underscore.
+fn separated_digits(radix: u32, must_start_with_digit: bool) -> impl Fn(&str) -> ParseResult<'_> {
```
djc, line 253: Please move this below `num_lit()` and rename `must_start_with_digit` to just `start`.

askama_parser/src/lib.rs
```diff
@@ -249,11 +248,75 @@ fn bool_lit(i: &str) -> ParseResult<'_> {
...
+    move |i| {
+        recognize(tuple((
+            |i| {
+                if must_start_with_digit {
+                    Ok((i, 0))
+                } else {
+                    many0_count(char('_'))(i)
+                }
+            },
```
djc, line 262: Please rewrite this `if` as a `match`:

```rust
match start {
    true => Ok((i, 0)),
    false => many0_count(char('_))(i)
}
```

askama_parser/src/lib.rs
```diff
@@ -249,11 +248,75 @@ fn bool_lit(i: &str) -> ParseResult<'_> {
...
+            satisfy(|ch| ch.is_digit(radix)),
+            many0_count(satisfy(|ch| ch == '_' || ch.is_digit(radix))),
+        )))(i)
+    }
+}
+
 fn num_lit(i: &str) -> ParseResult<'_> {
+    let integer_suffix_tags = |i| {
```
djc, line 270: Please remove the `_tags` suffixes from this (and `float_suffix_tags`).

djc (commented):
Thanks, this is close!

askama_parser/src/lib.rs
PizzasBear: done

askama_parser/src/lib.rs
PizzasBear: done

askama_parser/src/lib.rs
PizzasBear: done

askama_parser/src/lib.rs
```diff
@@ -250,13 +249,74 @@ fn bool_lit(i: &str) -> ParseResult<'_> {
...
+                        separated_digits(10, true),
+                        opt(float_suffix),
+                    ))),
+                ))),
+            ))),
+        )),
     )))(i)
 }
 
+/// Underscore separateed digits of the given base, unless `start` is true this may start
```
djc, line 305: Nit: "separateed" -> "separated".

djc (approved):

PR #909, opened 2023-11-17
Improve output error for non-matching number of arguments for macros

GuillaumeGomez:
Instead of checking for each argument if it doesn't go over the size, I added the check before the loop, allowing to iterate without needing to check anything.

While working on this, I also realized that it was ok to call the macro without passing all arguments. Should I maintain this behaviour or not? If you want me to keep this behaviour, I'll simply update the check to:

```rust
if args.len() > def.args.len() {
```

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -719,11 +719,18 @@ impl<'a> Generator<'a> {
         let mut names = Buffer::new(0);
         let mut values = Buffer::new(0);
         let mut is_first_variable = true;
+        if args.len() != def.args.len() {
+            return Err(CompileError::from(format!(
+                "macro {name:?} expected {} argument{}, found {}",
+                def.args.len(),
+                if def.args.len() != 1 { "s" } else { "" },
+                args.len()
+            )));
+        }
         for (i, arg) in def.args.iter().enumerate() {
             let expr = args.get(i).ok_or_else(|| {
                 CompileError::from(format!("macro {name:?} takes more than {i} arguments"))
             })?;
-
             match expr {
                 // If `expr` is already a form of variable then
                 // don't reintroduce a new variable. This is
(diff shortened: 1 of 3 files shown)


testing/tests/ui/macro.stderr
```diff
@@ -0,0 +1,23 @@
...
+
+error: macro "thrice" expected 2 arguments, found 0
+  --> tests/ui/macro.rs:11:10
+   |
+11 | #[derive(Template)]
+   |          ^^^^^^^^
+   |
+   = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error: macro "thrice" expected 0 argument, found 2
```
djc, line 17: Should write "arguments" for the 0 case in addition to the 1 case. :)

askama_derive/src/generator.rs
```diff
@@ -719,11 +719,15 @@ impl<'a> Generator<'a> {
...
-
+        if args.len() != def.args.len() {
+            return Err(CompileError::from(format!(
+                "macro {name:?} expected {} argument{}, found {}",
+                def.args.len(),
+                if def.args.len() > 1 { "s" } else { "" },
+                args.len()
+            )));
+        }
+        for (expr, arg) in std::iter::zip(args, &def.args) {
```
djc, line 730: Would you mind extracting the `enumerate()` + `.get()` -> `zip()` change into a separate commit, please?

djc (approved):
Looks nice!

testing/tests/ui/macro.stderr
GuillaumeGomez: I always wonder for the 0 case. Updating then. :)

askama_derive/src/generator.rs
GuillaumeGomez: Sure.

GuillaumeGomez:
Applied suggestions!

djc (approved):

PR #910, opened 2023-11-17
Allow to pass named arguments to macro calls

### LINKED ISSUE #885: Allow to have named arguments in macro

GuillaumeGomez:
Fixes #885.

Based on https://github.com/djc/askama/pull/909 (so only the last 3 commits are relevant to this PR).

I made a very permissive implementation which allows to mix named and non-named arguments. The only thing we need to check outside of the parser is that the named arguments actually have an equivalent in the called macro. For the rest, everything is checked in the parser directly, which allows nicer error messages.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -727,7 +727,50 @@ impl<'a> Generator<'a> {
                 args.len()
             )));
         }
-        for (expr, arg) in std::iter::zip(args, &def.args) {
+        let mut named_arguments = HashMap::new();
+        // Since named arguments can only be passed last, we only need to check if the last argument
+        // is a named one.
+        if let Some(Expr::NamedArgument(_, _)) = args.last() {
+            // First we check that all named arguments actually exist in the called item.
+            for arg in args.iter().rev() {
+                let Expr::NamedArgument(arg_name, _) = arg else {
+                    break;
+                };
+                if !def.args.iter().any(|arg| arg == arg_name) {
+                    return Err(CompileError::from(format!(
+                        "no argument named `{arg_name}` in macro {name:?}"
+                    )));
+                }
+                named_arguments.insert(arg_name, arg);
+            }
+        }
+
+        // Handling both named and unnamed arguments requires to be careful of the named arguments
+        // order. To do so, we iterate through the macro defined arguments and then check if we have
+        // a named argument with this name:
+        //
+        // * If there is one, we add it and move to the next argument.
+        // * If there isn't one, then we pick the next argument (we can do it without checking
+        //   anything since named arguments are always last).
+        let mut allow_positional = true;
+        for (index, arg) in def.args.iter().enumerate() {
+            let expr = match named_arguments.get(&arg) {
+                Some(expr) => {
+                    allow_positional = false;
+                    expr
+                }
+                None => {
+                    if !allow_positional {
+                        // If there is already at least one named argument, then it's not allowed
+                        // to use unnamed ones at this point anymore.
+                        return Err(CompileError::from(format!(
+                            "cannot have unnamed argument (`{arg}`) after named argument in macro \
+                             {name:?}"
+                        )));
+                    }
+                    &args[index]
+                }
+            };
             match expr {
                 // If `expr` is already a form of variable then
                 // don't reintroduce a new variable. This is
@@ -1104,6 +1147,7 @@ impl<'a> Generator<'a> {
             Expr::RustMacro(ref path, args) => self.visit_rust_macro(buf, path, args),
             Expr::Try(ref expr) => self.visit_try(buf, expr.as_ref())?,
             Expr::Tuple(ref exprs) => self.visit_tuple(buf, exprs)?,
+            Expr::NamedArgument(_, ref expr) => self.visit_named_argument(buf, expr)?,
         })
     }
 
@@ -1501,6 +1545,15 @@ impl<'a> Generator<'a> {
             buf.write(",");
         }
         buf.write(")");
+        Ok(DisplayWrap::Unwrapped)
+    }
+
+    fn visit_named_argument(
+        &mut self,
+        buf: &mut Buffer,
+        expr: &Expr<'_>,
+    ) -> Result<DisplayWrap, CompileError> {
+        self.visit_expr(buf, expr)?;
         Ok(DisplayWrap::Unwrapped)
     }
 
@@ -1913,6 +1966,7 @@ pub(crate) fn is_cacheable(expr: &Expr<'_>) -> bool {
         }
         Expr::Group(arg) => is_cacheable(arg),
         Expr::Tuple(args) => args.iter().all(is_cacheable),
+        Expr::NamedArgument(_, expr) => is_cacheable(expr),
         // We have too little information to tell if the expression is pure:
         Expr::Call(_, _) => false,
         Expr::RustMacro(_, _) => false,
--- a/askama_parser/src/expr.rs
+++ b/askama_parser/src/expr.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+use std::collections::HashSet;
 use std::str;
 
 use nom::branch::alt;
@@ -12,7 +14,7 @@ use nom::sequence::{pair, preceded, terminated, tuple};
 use super::{
     char_lit, identifier, not_ws, num_lit, path_or_identifier, str_lit, ws, Level, PathOrIdentifier,
 };
-use crate::ParseResult;
+use crate::{ErrorContext, ParseResult};
 
 macro_rules! expr_prec_layer {
     ( $name:ident, $inner:ident, $op:expr ) => {
@@ -61,6 +63,7 @@ pub enum Expr<'a> {
     Attr(Box<Expr<'a>>, &'a str),
     Index(Box<Expr<'a>>, Box<Expr<'a>>),
     Filter(&'a str, Vec<Expr<'a>>),
+    NamedArgument(&'a str, Box<Expr<'a>>),
     Unary(&'a str, Box<Expr<'a>>),
     BinOp(&'a str, Box<Expr<'a>>, Box<Expr<'a>>),
     Range(&'a str, Option<Box<Expr<'a>>>, Option<Box<Expr<'a>>>),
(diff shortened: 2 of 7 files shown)


GuillaumeGomez:
Rebased.

askama_derive/src/generator.rs
```diff
@@ -727,6 +727,42 @@ impl<'a> Generator<'a> {
...
             )));
         }
+        let mut output_args = Vec::new();
+        let args = if args
+            .iter()
+            .any(|arg| matches!(arg, Expr::NamedArgument(_, _)))
+        {
+            // First we check that all named arguments actually exist in the called item.
+            for arg in args.iter() {
+                if let Expr::NamedArgument(arg_name, _) = arg {
```
djc, line 737: Let's do `let .. else` for the outer branch here?

askama_derive/src/generator.rs
```diff
@@ -727,6 +727,42 @@ impl<'a> Generator<'a> {
                 args.len()
             )));
         }
+        let mut output_args = Vec::new();
```
djc, line 730: This seems pretty messy. I'd like to have a comment here that explains the high-level logic for handling invocations that have both named and unnamed arguments, which seems like the tricky case here. Then, I think there should be one pass over the arguments that both splits the arguments in `unnamed_args` and `named_args` (I think we should be able to do this without cloning?) and checks whether all the names are acceptable.

askama_parser/src/expr.rs
```diff
@@ -74,15 +77,45 @@ pub enum Expr<'a> {
 impl<'a> Expr<'a> {
     pub(super) fn arguments(i: &'a str, level: Level) -> ParseResult<'a, Vec<Self>> {
         let (_, level) = level.nest(i)?;
+        let mut named_arguments = HashSet::new();
```
djc, line 80: I don't think this logic should live in `Expr::arguments()`, which is not just used for macros.

book/src/template_syntax.md
```diff
@@ -587,3 +587,49 @@ You can optionally specify the name of the macro in `endmacro`:
...
+You can also specify arguments by their name (as defined in the macro):
+
+```
+{% macro heading(arg, bold) %}
+
+<h1>{{arg}} <b>{{bold}}</b></h1>
+
+{% endmacro %}
+
+{% call heading(bold="something", arg = "title") %}
```
djc, line 600: I think the documentation should be consistent in whether it uses spacing around `=` even though the parser should cope with both.

askama_derive/src/generator.rs
GuillaumeGomez: I don't really see how it would improve the code since there is no `else` clause here?

askama_parser/src/expr.rs
GuillaumeGomez: It allows to greatly improve the parser error, but as you prefer.

GuillaumeGomez: I updated the code to not always allow to have named arguments. It'll fail earlier in this case.

However, I still think we should keep the `named_arguments` set checks for duplicated names as it makes the errors much better (otherwise we don't get the origin of the error since we don't have the template text anymore when generating the code).

book/src/template_syntax.md
GuillaumeGomez: Updated! I also precised that we can have whitespace characters around `=`.

GuillaumeGomez:
Is there anything else to be done here?

askama_parser/src/expr.rs
```diff
@@ -72,17 +75,74 @@ pub enum Expr<'a> {
...
+                            is_jinja_macro,
+                        )
+                    }),
+                ),
                 char(')'),
             )),
         )(i)
     }
 
+    fn arguments_parsing(
```
djc, line 107: I'd prefer to inline this. I wonder if we can move the `if` inside the `alt` somehow?

```rust
alt((
    match is_jinja_macro {
        true => move |i| Self::named_argument(i, level, named_arguments, start),
        false => move |i| nom::err::Error::something(), // TODO?
    },
    move |i| Self::parse(i, level),
))(i)
```

askama_parser/src/expr.rs
```diff
@@ -72,17 +75,74 @@ pub enum Expr<'a> {
 }
 
 impl<'a> Expr<'a> {
-    pub(super) fn arguments(i: &'a str, level: Level) -> ParseResult<'a, Vec<Self>> {
+    pub(super) fn arguments(
+        i: &'a str,
+        level: Level,
+        is_jinja_macro: bool,
```
djc, line 81: Prefer to rename this to `is_template_macro`, since this isn't Jinja.

askama_parser/src/expr.rs
```diff
@@ -72,17 +75,74 @@ pub enum Expr<'a> {
...
+    fn named_argument(
+        i: &'a str,
+        level: Level,
+        named_arguments: &mut HashSet<&'a str>,
+        start: &'a str,
+    ) -> ParseResult<'a, Self> {
+        let (_, level) = level.nest(i)?;
+        let (i, (argument, _, value)) =
+            tuple((identifier, ws(char('=')), move |i| Self::parse(i, level)))(i)?;
+        if !named_arguments.insert(argument) {
```
djc, line 134: Please invert this `if` with an early return so that we can avoid rightward drift for the error case.

askama_derive/src/generator.rs
djc: This is still iterating `args` twice: once for `args.iter().any()` and once for `args.iter().partition()`. I'm pretty sure this isn't actually necessary?

askama_derive/src/generator.rs
```diff
@@ -727,6 +727,52 @@ impl<'a> Generator<'a> {
                 args.len()
             )));
         }
+        // Handling both named and unnamed arguments requires to be careful of the arguments order.
```
djc, line 730: I think I'd prefer a more restrictive way of combining named and unnamed arguments, which is that we only allow named arguments after positional arguments.

askama_derive/src/generator.rs
GuillaumeGomez: I can simplify it even further indeed. I'll do that.

GuillaumeGomez: My idea couldn't work actually (sorry for the noise). The problem is that I first need to **ensure** that all named arguments actually have an equivalent in the macro, otherwise they'll be passed as non-named arguments and the `remove` call would fail. However, a first improvement that could be done would be to move this loop after we split both kind of arguments so that we only iterate on named arguments.

askama_parser/src/expr.rs
GuillaumeGomez: :+1:

askama_parser/src/expr.rs
GuillaumeGomez: We can actually do even better and move the `is_macro` check directly into `named_argument` to make it return an error if it's not the case. Much simpler than currently.

askama_derive/src/generator.rs
GuillaumeGomez: Since named arguments are always last, it allowed to remove the first loop and to instead just check if the last element is a named argument.

GuillaumeGomez:
Updated! I applied your suggestions and also enforced that named arguments should always be the last arguments. As mentioned above, it allowed to remove a loop since we can check if we have named arguments by checking if the last argument is a named one.

askama_derive/src/generator.rs
```diff
@@ -759,6 +749,18 @@ impl<'a> Generator<'a> {
             let (mut named_arguments, mut non_named_arguments): (Vec<_>, Vec<_>) = args
                 .iter()
                 .partition(|item| matches!(item, Expr::NamedArgument(_, _)));
+
```
djc, line 755: These changes should be inlined into the original commits.

GuillaumeGomez:
Squashed commits so named arguments coming last are now part of the first commit.

askama_derive/src/generator.rs
djc: Please update this comment to be in line with the updated logic.

askama_derive/src/generator.rs
```diff
@@ -727,6 +727,57 @@ impl<'a> Generator<'a> {
...
+        // named argument with this name:
+        //
+        // * If there is one, we add it and move to the next argument.
+        // * If there isn't one, we then need to pick the next unnamed argument (as we consider they
+        //   are passed in the right order). To ensure we won't accidentally pass a named argument,
+        //   we split named and unnamed arguments in two vectors before starting.
+        let mut output_args = Vec::new();
+        // Since named arguments can only be passed last, we only need to check if the last argument
+        // is a named one.
+        let args = if args
```
djc, line 741: I think this can be simplified to `if let Some(Expr::NamedArgument(_, _)) = args.last()`?

askama_derive/src/generator.rs
```diff
@@ -727,6 +727,57 @@ impl<'a> Generator<'a> {
...
+            //
+            // To make things easier, we split them in two different vectors so we won't need to
+            // check each time we need to push an unnamed argument if it isn't, in fact, a named
+            // one.
+            let (mut named_arguments, mut non_named_arguments): (Vec<_>, Vec<_>) = args
+                .iter()
+                .partition(|item| matches!(item, Expr::NamedArgument(_, _)));
+
+            // First we check that all named arguments actually exist in the called item.
+            for arg in named_arguments.iter() {
```
djc, line 757: I think the right logic here is:

- Set up a `named_args` `HashMap<&str, &Expr>`
- For each argument, in reverse order:
  - If the argument is not named, break out of the loop
  - If it is named, insert into `named_args`
- Then, change the `iter::zip()` loop to add `.enumerate()`
  - If an argument is named, use the `named_args` map to lookup the actual argument

askama_parser/src/expr.rs
djc: I still think we should inline `arguments_parsing()`.

askama_derive/src/generator.rs
GuillaumeGomez: Much better indeed!

askama_parser/src/expr.rs
GuillaumeGomez: I find the code a bit less nice to follow but if you prefer it inlined, let's do it! :)

GuillaumeGomez:
The idea of using a `HashMap` was really good. I simplified the code quite a lot: I loop through the template arguments, for each of them, I check if there is a named argument with this name in the `HashMap` and if so use the associated `Expr`.

With this, we avoid allocating other `Vec`s and remove a few loops.

I updated the code comments as well.

askama_derive/src/generator.rs
```diff
@@ -727,7 +727,41 @@ impl<'a> Generator<'a> {
...
+        }
+
+        // Handling both named and unnamed arguments requires to be careful of the named arguments
+        // order. To do so, we iterate through the macro defined arguments and then check if we have
+        // a named argument with this name:
+        //
+        // * If there is one, we add it and move to the next argument.
+        // * If there isn't one, then we pick the next argument (we can do it without checking
+        //   anything since named arguments are always last).
+        let mut index = 0;
```
GuillaumeGomez, line 756: Instead of this index, I could use an iterator. As you prefer.

askama_derive/src/generator.rs
djc: It reduces rightward drift for the further check, so I would still like to do this.

askama_derive/src/generator.rs
djc: This seems fine.

book/src/template_syntax.md
```diff
@@ -587,3 +587,57 @@ You can optionally specify the name of the macro in `endmacro`:
...
+However please note than named arguments must always come **last**.
+
+In this case, the named arguments will be passed to the related parameter and then the other
+arguments will be placed in what remains. Let's explain it in code:
+
+```
+{% macro heading(arg1, arg2, arg3, arg4) %}
+{% endmacro %}
+
+{% call heading("something", "b", arg4="ah", arg2="title") %}
```
djc, line 624: Ahh, sorry, these aren't quite the semantics I was going for. The positional arguments should always retain their position, so it should not be allowed to specify `arg2` by name here since the call also passes two positional arguments.

GuillaumeGomez: That however seems a bit too much of a restriction imo. We can't mix named and unnamed arguments, so I would at least want to be able to specify some arguments by name (because it's not obvious what they are just from the value) whether or not it is one of the first arguments without having to specify all arguments as named ones. If that makes sense to you?

askama_derive/src/generator.rs
GuillaumeGomez: Now there is a good reason for it (the `break` which was added only recently) so adding it!

askama_derive/src/generator.rs
```diff
@@ -727,7 +727,50 @@ impl<'a> Generator<'a> {
...
+        }
+
+        // Handling both named and unnamed arguments requires to be careful of the named arguments
+        // order. To do so, we iterate through the macro defined arguments and then check if we have
+        // a named argument with this name:
+        //
+        // * If there is one, we add it and move to the next argument.
+        // * If there isn't one, then we pick the next argument (we can do it without checking
+        //   anything since named arguments are always last).
+        let mut args_iterator = args
```
djc, line 755: The `args_iterator` approach is unnecessarily complex. We can just have an `allow_positional` that is initialized to `true` which is changed to `false` on the first occurrence of a named argument.

GuillaumeGomez:
Updated!

askama_derive/src/generator.rs
djc: How less good? Do you have an example?

GuillaumeGomez: Before I could say `it's bad to use named argument "x" here`. Now I simply provide the argument name of the template macro. I could maybe provide the index too.

But I have to admit, before it was a bit strange too:

> is taking the position of an unnamed argument

Not very clear.

djc (approved):
Okay, thanks for sticking with this, I think this is in a pretty good shape now!

GuillaumeGomez:
Thanks for all the review rounds! I'll continue my docs.rs migration and come back here if anything else is missing. :)

djc:
Sounds great!

Issue #572, opened 2021-12-07
Allow inlined extending within included templates.


genusistimelord:
So here is the issue, in the current Jinja i can use a card.html which include something like this

```
<div class="uk-card uk-card-default">
    <div class="uk-card-header">
        <span class="uk-card-title">
            {% block card_title %}{% endblock card_title %}
        </span>
        {% block card_badge %}{% endblock card_badge %}
    </div>
    <div class="uk-card-body">
        {% block card_body %}{% endblock card_body %}
    </div>
    {% block card_footer %}{% endblock card_footer %}
</div>
```

so then when i go to setup certain things lets say I want to make a Article Card that uses the card.html to extend itself 

```
{% extends "components/card.html" %}

{% block card %}
    {% block card_title %}
            Registration Agreement
    {% endblock card_title %}
    {% block card_body %}
                    <p>This is a random Article</p>
                    <p><b>Last update: August 9, 2018</b></p>
                    <p><b>By continuing you agree to the above privacy policy.</b></p>
    {% endblock card_body %}
{% endblock card %}
```

This would in jinja load the single extension, Place the changes inside it, then send the changed version down the line to be used in the parent. However Askama does not support this. Would it be Possible to add this or to add it as an inline_extend type? As this is useful for reducing HTML that may get modified to change theme looks ETC without having to go thru 1000 files to do so.

djc:
Sorry, it's not really clear to me what you mean by this. Can you do a more complete example of what you're trying to achieve in terms of output and how reuse should facilitate that?

genusistimelord:
I think the major issue is you only allow parents to Extend so in askamas Case it would only allow
<- means to extend -> means to include

base<-forum->category-card

for this to work id need to be able to 
base<-forum->card<-category-card

djc:
Ahhh, so you want includes to support inheritance? I don't see any fundamental reason why that should not work, might just be an oversight in the parser or code generator. Do you get an error message when you try this?

genusistimelord:
yeah inheritance would be nice. Currently when i try this it tells me extend only works for the parent. Though we could probably make it more apparent that it is an inheritance rather than an extend. Maybe by making it {% inherit "components/category-card.html" %} rather than {% extend "components/category-card.html" %}.  If this makes it easier to introduce and fix. Otherwise if it is a easy fix then I would not worry about the rename. =3.

djc:
[...]

PR #915, opened 2023-11-23
Allow `extends`, `macro` and `import` inside an included template

PizzasBear:
This makes `include` work [more like in Jinja2 ](https://jinja.palletsprojects.com/en/3.1.x/templates/#include), and closes (#572).
This also improves caching of included templates by storing them inside the `contexts` map.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1,8 +1,8 @@
 use std::collections::hash_map::{Entry, HashMap};
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::{cmp, hash, mem, str};
 
-use crate::config::{get_template_source, WhitespaceHandling};
+use crate::config::WhitespaceHandling;
 use crate::heritage::{Context, Heritage};
 use crate::input::{Source, TemplateInput};
 use crate::CompileError;
@@ -10,7 +10,7 @@ use crate::CompileError;
 use parser::node::{
     Call, Comment, CondTest, If, Include, Let, Lit, Loop, Match, Target, Whitespace, Ws,
 };
-use parser::{Expr, Node, Parsed};
+use parser::{Expr, Node};
 use quote::quote;
 
 pub(crate) struct Generator<'a> {
@@ -20,8 +20,6 @@ pub(crate) struct Generator<'a> {
     contexts: &'a HashMap<&'a Path, Context<'a>>,
     // The heritage contains references to blocks and their ancestry
     heritage: Option<&'a Heritage<'a>>,
-    // Cache ASTs for included templates
-    includes: HashMap<PathBuf, Parsed>,
     // Variables accessible directly from the current scope (not redirected to context)
     locals: MapChain<'a, &'a str, LocalMeta>,
     // Suffix whitespace from the previous literal. Will be flushed to the
@@ -50,7 +48,6 @@ impl<'a> Generator<'a> {
             input,
             contexts,
             heritage,
-            includes: HashMap::default(),
             locals,
             next_ws: None,
             skip_ws: WhitespaceHandling::Preserve,
@@ -846,24 +843,35 @@ impl<'a> Generator<'a> {
             )?;
         }
 
-        // Since nodes must not outlive the Generator, we instantiate a nested `Generator` here to
-        // handle the include's nodes. Unfortunately we can't easily share the `includes` cache.
+        // We clone the context of the child in order to preserve their macros and imports.
+        // But also add all the imports and macros from this template that don't override the
+        // child's ones to preserve this template's context.
+        let child_ctx = &mut self.contexts[path.as_path()].clone();
+        for (name, mac) in &ctx.macros {
+            child_ctx.macros.entry(name).or_insert(mac);
+        }
+        for (name, import) in &ctx.imports {
+            child_ctx
+                .imports
+                .entry(name)
+                .or_insert_with(|| import.clone());
+        }
 
-        let locals = MapChain::with_parent(&self.locals);
-        let mut child = Self::new(self.input, self.contexts, self.heritage, locals);
-
-        let nodes = match self.contexts.get(path.as_path()) {
-            Some(ctx) => ctx.nodes,
-            None => match self.includes.entry(path) {
-                Entry::Occupied(entry) => entry.into_mut().nodes(),
-                Entry::Vacant(entry) => {
-                    let src = get_template_source(entry.key())?;
-                    entry.insert(Parsed::new(src, self.input.syntax)?).nodes()
-                }
-            },
+        // Create a new generator for the child, and call it like in `impl_template` as if it were
+        // a full template, while preserving the context.
+        let heritage = if !child_ctx.blocks.is_empty() || child_ctx.extends.is_some() {
+            Some(Heritage::new(child_ctx, self.contexts))
+        } else {
+            None
         };
 
-        let mut size_hint = child.handle(ctx, nodes, buf, AstLevel::Nested)?;
+        let handle_ctx = match &heritage {
+            Some(heritage) => heritage.root,
+            None => child_ctx,
+        };
+        let locals = MapChain::with_parent(&self.locals);
+        let mut child = Self::new(self.input, self.contexts, heritage.as_ref(), locals);
+        let mut size_hint = child.handle(handle_ctx, handle_ctx.nodes, buf, AstLevel::Top)?;
         size_hint += child.write_buf_writable(buf)?;
         self.prepare_ws(i.ws);
 
--- a/askama_derive/src/input.rs
+++ b/askama_derive/src/input.rs
@@ -119,8 +119,13 @@ impl TemplateInput<'_> {
             while let Some(nodes) = nested.pop() {
                 for n in nodes {
                     let mut add_to_check = |path: PathBuf| -> Result<(), CompileError> {
-                        let source = get_template_source(&path)?;
-                        check.push((path, source));
+                        if !map.contains_key(&path) {
+                            // Add a dummy entry to `map` in order to prevent adding `path`
+                            // multiple times to `check`.
+                            map.insert(path.clone(), Parsed::default());
+                            let source = get_template_source(&path)?;
+                            check.push((path, source));
+                        }
                         Ok(())
                     };
 
--- a/askama_derive/src/input.rs
+++ b/askama_derive/src/input.rs
@@ -118,6 +118,12 @@ impl TemplateInput<'_> {
             let mut nested = vec![parsed.nodes()];
             while let Some(nodes) = nested.pop() {
                 for n in nodes {
+                    let mut add_to_check = |path: PathBuf| -> Result<(), CompileError> {
+                        let source = get_template_source(&path)?;
+                        check.push((path, source));
+                        Ok(())
+                    };
+
                     use Node::*;
                     match n {
                         Extends(extends) if top => {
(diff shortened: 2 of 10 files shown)


djc:
What is the use case? Why do you want this?

PizzasBear:
This allows me to use macros fully inside an included template without having to create another macro file that is associated with the included template and importing it using the same name everywhere I include this template.

GuillaumeGomez:
Nice! I'm working on something similar: allowing imports and macro declarations everywhere. So big :+1: for this from me. :)

Do you have some numbers for the impact on compilation performance by any chance?

PizzasBear:
No, I didn't measure the performance impact. Though I suspect it wouldn't get much worse because this caches included templates better, but it does clone the `Context` struct for every include.

GuillaumeGomez:
Let's see if there is a change already and if there one big enough, we can start optimizing (if @djc is ok with this feature of course).

PizzasBear:
I don't know how to measure the performance impact, can you explain me?

GuillaumeGomez:
Very ad-hoc way to do it (if nothing comes out from this, I think it's mostly ok): write a file with like 10 templates of different sizes. Then run `cargo build --timings` with and without this PR a few times and compare the median time for both.

PizzasBear:
With big files full of `if` statements there doesn't seem to be any difference from `main`.  And including a bunch of big files full of `if`s doesn't seem to show any differences either.

GuillaumeGomez:
Can you provide us the code you used to test so we can also confirm please?

PizzasBear:
I used the following to generate the templates.
```python
from datetime import datetime


def generate_big_if(num: int, level: int = 0) -> str:
    indent = "    " * level
    if num <= 1:
        return f"{indent}hello {datetime.now()}\n"
    right = num // 2
    left = num - right
    return (
        indent
        + "{% if true %}\n"
        + generate_big_if(left, level + 1)
        + indent
        + "{% else %}\n"
        + generate_big_if(right, level + 1)
        + indent
        + "{% endif %}\n"
    )


def main():
    for i in range(9):
        with open(f"templates/{i}.html", "w+") as f:
            f.write(generate_big_if(int(1.75 ** (i + 3))))


if __name__ == "__main__":
    main()
```
And had this file as well:
```jinja
{% include "1.html" %}
{% include "2.html" %}
{% include "3.html" %}
{% include "4.html" %}
{% include "5.html" %}
{% include "6.html" %}
{% include "7.html" %}
{% include "8.html" %}
{% include "1.html" %}
{% include "2.html" %}
{% include "3.html" %}
{% include "4.html" %}
{% include "5.html" %}
{% include "6.html" %}
{% include "7.html" %}
{% include "8.html" %}
```

And compiled:
```rust
use askama::Template;

#[derive(Template)]
#[template(path = "0.html")]
struct T0;

#[derive(Template)]
#[template(path = "1.html")]
struct T1;

#[derive(Template)]
#[template(path = "2.html")]
struct T2;

#[derive(Template)]
#[template(path = "3.html")]
struct T3;

#[derive(Template)]
#[template(path = "4.html")]
struct T4;

#[derive(Template)]
#[template(path = "5.html")]
struct T5;

#[derive(Template)]
#[template(path = "6.html")]
struct T6;

#[derive(Template)]
#[template(path = "7.html")]
struct T7;

#[derive(Template)]
#[template(path = "8.html")]
struct T8;

#[derive(Template)]
#[template(path = "9.html")]
struct T9;

fn main() {
    println!("Hello, world!");
}
```

GuillaumeGomez:
Thanks! Confirming the impact on my side.

askama_derive/src/generator.rs
```diff
@@ -802,9 +803,18 @@ impl<'a> Generator<'a> {
         // Since nodes must not outlive the Generator, we instantiate a nested `Generator` here to
         // handle the include's nodes. Unfortunately we can't easily share the `includes` cache.
 
-        let ctx = &self.contexts[path.as_path()];
-        let heritage = if !ctx.blocks.is_empty() || ctx.extends.is_some() {
-            Some(Heritage::new(ctx, self.contexts))
+        let child_ctx = &mut self.contexts[path.as_path()].clone();
```
djc, line 806: These changes should be squashed into the first commit. I review on a commit-by-commit basis, so PRs should have a clean commit history.

Please add some comments about what this code is trying to do/why it is necessary.

askama_derive/src/input.rs
```diff
@@ -113,32 +113,63 @@ impl TemplateInput<'_> {
...
+                        Node::Loop(l) => {
+                            nested.push(&l.body);
+                            nested.push(&l.else_nodes);
+                        }
+                        Node::Match(m) => {
+                            for arm in &m.arms {
+                                nested.push(&arm.nodes);
+                            }
+                        }
+                        _ => {}
```
djc, line 178: Since this is now listing many `Node` variants, I'd prefer to explicitly list variants to be ignored here instead of allowing the rest to silently match. That way, we'll explicitly have to consider whether a new variant needs to be handled here.

askama_derive/src/input.rs
```diff
@@ -113,32 +113,63 @@ impl TemplateInput<'_> {
...
+                                    "cyclic dependency in graph {:#?}",
+                                    dependency_graph
+                                        .iter()
+                                        .map(|e| format!("{:#?} --> {:#?}", e.0, e.1))
+                                        .collect::<Vec<String>>()
+                                )
+                                .into());
+                            }
+                            dependency_graph.push(dependency_path);
+                            let source = get_template_source(&extends)?;
```
djc: Before this commit, I'd like to see a commit that abstracts over the `get_template_source()`,  then `check.push()` pattern -- the abstraction should work for the later optimization, too.

askama_derive/src/input.rs
```diff
@@ -113,32 +113,63 @@ impl TemplateInput<'_> {
...
-                                dependency_graph
-                                    .iter()
-                                    .map(|e| format!("{:#?} --> {:#?}", e.0, e.1))
-                                    .collect::<Vec<String>>()
-                            )
-                            .into());
+
+            let mut top = true;
+            let mut nested = vec![parsed.nodes()];
+            while let Some(nodes) = nested.pop() {
```
djc, line 119: Why do we move from a `while` to a `for` here?

PizzasBear: We need to DFS through all the nodes to find the `include` statements just like in `Context::new`.

askama_derive/src/generator.rs
djc: Sure, use `git rebase --interactive main`, reorder the lines and replace the first word with `squash` (or `s`).

askama_derive/src/input.rs
djc: Why doesn't the `check` `Vec` suffice for doing that?

askama_derive/src/input.rs
PizzasBear: The ignored node list is quite long but okay

djc: I suggest doing a `use Node::*` before the `match` to avoid the repetition.

PizzasBear: done

askama_derive/src/generator.rs
PizzasBear: done

askama_derive/src/input.rs
PizzasBear: I added a closure that does that

askama_derive/src/input.rs
```diff
@@ -118,8 +118,17 @@ impl TemplateInput<'_> {
...
                 for n in nodes {
+                    let mut add_to_check = |path: PathBuf| -> Result<(), CompileError> {
+                        if !map.contains_key(&path) {
+                            map.insert(path.clone(), Parsed::default());
+                            let source = get_template_source(&path)?;
+                            check.push((path, source));
+                        }
+                        Ok(())
+                    };
+                    use Node::*;
```
djc, line 129: Please introduce the `use Node::*` in a separate commit at the start of the PR.

askama_derive/src/input.rs
djc: I think you missed the part that requested this in a separate commit, before the other commits you have so far. Please closely read and follow-up with my feedback.

djc: No:

* One commit to abstract over the `check` additions
* One commit to improve performance for the `check` additions
* One commit for the other things

askama_derive/src/input.rs
PizzasBear: done

askama_derive/src/generator.rs
```diff
@@ -843,8 +843,9 @@ impl<'a> Generator<'a> {
             )?;
         }
 
-        // Since nodes must not outlive the Generator, we instantiate a nested `Generator` here to
-        // handle the include's nodes. Unfortunately we can't easily share the `includes` cache.
+        // We clone the context of the child in order to preserve their macros and imports.
```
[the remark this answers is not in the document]
PizzasBear: Merged them into the first commit

askama_derive/src/input.rs
PizzasBear: Are the new commits okay?

askama_derive/src/input.rs
PizzasBear: I don't understand, do you mean to put this before the main commit?

askama_derive/src/input.rs
```diff
@@ -119,8 +119,11 @@ impl TemplateInput<'_> {
             while let Some(nodes) = nested.pop() {
                 for n in nodes {
                     let mut add_to_check = |path: PathBuf| -> Result<(), CompileError> {
-                        let source = get_template_source(&path)?;
-                        check.push((path, source));
+                        if !map.contains_key(&path) {
+                            map.insert(path.clone(), Parsed::default());
```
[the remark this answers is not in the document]
PizzasBear: The `parsed` variable has to be owned and not mutably borrowed from `map` because we are going to insert new entries into it. Inserting a dummy entry prevents re-adding the same file multiple times to `check` before we process it.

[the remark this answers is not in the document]
GuillaumeGomez: Worth adding a code comment to explain it then. ;)

askama_derive/src/input.rs
djc: Yes.

askama_derive/src/input.rs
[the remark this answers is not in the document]
PizzasBear: Should I add it to the commit: "[Improve performance of find_used_templates](https://github.com/djc/askama/pull/915/commits/f752cc5396b311afcb8ed104cb76ccd3408f28d2)", or make a separate one?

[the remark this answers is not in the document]
PizzasBear: I added the comment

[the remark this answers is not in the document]
GuillaumeGomez: :+1:

askama_derive/src/input.rs
PizzasBear: redone

GuillaumeGomez:
Does it look good to you @djc?

GuillaumeGomez:
Ping @djc.

askama_derive/src/generator.rs
```diff
@@ -846,24 +843,35 @@ impl<'a> Generator<'a> {
...
         };
 
-        let mut size_hint = child.handle(ctx, nodes, buf, AstLevel::Nested)?;
+        let locals = MapChain::with_parent(&self.locals);
+        let mut child = Self::new(self.input, self.contexts, heritage.as_ref(), locals);
+        let mut size_hint = if let Some(heritage) = &heritage {
+            child.handle(heritage.root, heritage.root.nodes, buf, AstLevel::Top)
+        } else {
+            child.handle(child_ctx, child_ctx.nodes, buf, AstLevel::Top)
+        }?;
```
djc, line 874: Style nit: I'd prefer to write this like this:

```rust
let ctx = match &heritage {
     Some(heritage) => &heritage.root,
     None => &child_ctx,
};

let mut size_hint = child.handle(ctx, ctx.nodes, buf, AstLevel::Top);
```

djc (approved):
Okay, sorry for the long delay -- one more nit.

askama_derive/src/generator.rs
PizzasBear: done

djc:
Please squash that change into the originating commit.

PizzasBear:
I squashed it

MeirKriheli:
Thanks for this!

Is it going into a release soon? It's very useful with partials for htmx.

djc:
I wrote up some thoughts on release planning in #1035.

PR #916, opened 2023-11-25
Add support for `{% if var is [not] defined %}` statements

### LINKED ISSUE #257: Tests support like: {% if foo is defined %}     {{ foo }} {% endif %}
```jinja
{% if foo is defined %}
    {{ foo }}
{% endif %}
```

This doesn't work with askama. It works in tera.
Issue [#46](https://github.com/djc/askama/issues/46) is closed with "This has long been done." but it's either not the case or it's only the case for the match syntax from the same issue.

In tera/jijna2 lingo _defined_ is a _test_:
[tera tests](https://tera.netlify.com/docs/templates/#built-in-tests)
[jinja 2 tests](http://jinja.pocoo.org/docs/2.10/templates/#builtin-tests).
I couldn't find any tests in askama, don't know where to put this.

Code wise my _defined_ is rather simple:
```rust
pub fn defined<T>(o: &Option<T>) -> Result<bool>
{
    if o.is_some() {
        return Ok(true);
    }
    Ok(false)
}
```
The reason for using an Option is that in the template input struct everything a template refers to must always be defined anyway, so to satisfy a template with optional data, using the rust Option type makes the most sense.

Results are different in that the Err can also hold a value, so that's not something to use with defined anyway.

PizzasBear:
Implements compile time checking of whether a variable is defined in an `if` statement, which can be useful for `include` statements. Closes #257.

```jinja
{%- if name is defined -%}
    Hello {{ name }}!
{%- else if other_name is not defined -%}
    Both names aren't defined
{%- else -%}
    Aloha {{ other_name }}!
{%- endif -%}
```

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -480,7 +480,8 @@ impl<'a> Generator<'a> {
                 syn::Data::Struct(s) => s
                     .fields
                     .iter()
-                    .any(|field| field.ident.as_ref().is_some_and(|f| f == name)),
+                    .filter_map(|f| f.ident.as_ref())
+                    .any(|f| f == name),
                 _ => false,
             }
         }
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -472,6 +472,20 @@ impl<'a> Generator<'a> {
         Ok(size_hint)
     }
 
+    fn has_local(&self, name: &str) -> bool {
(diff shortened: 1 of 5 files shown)


djc:
I think I'd rather stay away from this feature. I worry that actually inspecting the fields passed in the context type will constrain different approaches to code generation later on (for example, using a procedural macro).

PR #919, opened 2023-11-27
Add checking for declaration of reserved variable names (`writer` and `self`)

PizzasBear:
Check for declaration of variables named `writer` or `self` when parsing a target, and error if found.


djc:
Did you actually run into this? What's your use case, anyway?

PizzasBear:
I didn't run into it, but I just thought it'd be useful for someone who does.

djc:
Okay, I think think this complexity has diminishing returns so I'd rather avoid this.

Kijewski:
IMO that's a good addition. I don't think people will try to use `self` as an argument name, but I can see people stumble over the name `writer`.

djc (approved):

djc:
Alright, let's do it.

PR #922, opened 2023-12-04
Add automatic borrowing to `let` statements

### LINKED ISSUE #330: Add support for reference operator (i.e. `{% let x = &y %}`)
Sometimes while working with deeply nested structures, it would be useful to create an alias to a field, like say `{% let h = &parent.child.hash %}`

At the moment the `&` unary operator is not supported.

PizzasBear:
Adds automatic borrowing to `let` statements if value isn't "copyable". Eases #330.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -916,7 +916,14 @@ impl<'a> Generator<'a> {
         };
 
         let mut expr_buf = Buffer::new(0);
+        let borrow_val = !is_copyable(val);
+        if borrow_val {
+            expr_buf.write("&(");
+        }
         self.visit_expr(&mut expr_buf, val)?;
+        if borrow_val {
+            expr_buf.write(")");
+        }
 
         let shadowed = self.is_shadowing_variable(&l.var)?;
         if shadowed {
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1945,7 +1945,7 @@ fn is_copyable_within_op(expr: &Expr<'_>, within_op: bool) -> bool {
 pub(crate) fn is_attr_self(expr: &Expr<'_>) -> bool {
     match expr {
         Expr::Attr(obj, _) if matches!(obj.as_ref(), Expr::Var("self")) => true,
-        Expr::Attr(obj, _) if matches!(obj.as_ref(), Expr::Attr(..)) => is_attr_self(expr),
+        Expr::Attr(obj, _) if matches!(obj.as_ref(), Expr::Attr(..)) => is_attr_self(obj),
         _ => false,
     }
 }
--- a/testing/tests/calls.rs
+++ b/testing/tests/calls.rs
@@ -80,3 +80,22 @@ fn test_one_func_binop() {
     };
     assert_eq!(t.render().unwrap(), "246");
 }
+
+fn double_attr_arg_helper(x: u32) -> u32 {
+    x * x + x
+}
+
+#[derive(askama::Template)]
+#[template(
+    source = "{{ self::double_attr_arg_helper(self.x.0 + 2) }}",
+    ext = "txt"
+)]
+struct DoubleAttrArg {
+    x: (u32,),
+}
+
+#[test]
+fn test_double_attr_arg() {
+    let t = DoubleAttrArg { x: (10,) };
+    assert_eq!(t.render().unwrap(), "156");
+}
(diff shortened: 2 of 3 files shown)


GuillaumeGomez:
Can you show what the generated code looks like with this change (with `cargo expand`) please? (Both before and after)

PizzasBear:
Before:
```rust
#[template(source = "{% let word = s %}{{ word }}", ext = "html")]
struct LetBorrow {
    s: String,
}
impl ::askama::Template for LetBorrow {
    fn render_into(
        &self,
        writer: &mut (impl ::std::fmt::Write + ?Sized),
    ) -> ::askama::Result<()> {
        let word = self.s; // THIS LINE IS CHANGED
        writer
            .write_fmt(
                format_args!(
                    "{0}",
                    &::askama::MarkupDisplay::new_unsafe(&(word), ::askama::Html),
                ),
            )?;
        ::askama::Result::Ok(())
    }
    const EXTENSION: ::std::option::Option<&'static ::std::primitive::str> = Some(
        "html",
    );
    const SIZE_HINT: ::std::primitive::usize = 3;
    const MIME_TYPE: &'static ::std::primitive::str = "text/html; charset=utf-8";
}
impl ::std::fmt::Display for LetBorrow {
    #[inline]
    }
}
```
After:
```rust
#[template(source = "{% let word = s %}{{ word }}", ext = "html")]
struct LetBorrow {
    s: String,
}
impl ::askama::Template for LetBorrow {
    fn render_into(
        &self,
        writer: &mut (impl ::std::fmt::Write + ?Sized),
    ) -> ::askama::Result<()> {
        let word = &(self.s); // THIS LINE IS CHANGED
        writer
            .write_fmt(
                format_args!(
                    "{0}",
                    &::askama::MarkupDisplay::new_unsafe(&(word), ::askama::Html),
                ),
            )?;
        ::askama::Result::Ok(())
    }
    const EXTENSION: ::std::option::Option<&'static ::std::primitive::str> = Some(
        "html",
    );
    const SIZE_HINT: ::std::primitive::usize = 3;
    const MIME_TYPE: &'static ::std::primitive::str = "text/html; charset=utf-8";
}
impl ::std::fmt::Display for LetBorrow {
    #[inline]
    }
}
```

GuillaumeGomez:
I tried to think of case where having a reference to a field of a struct would prevent another case to work but since `render_into` takes a non-mutable reference, I think it's actually not possible... Looks good as far as I can see.

askama_derive/src/generator.rs
```diff
@@ -873,7 +873,14 @@ impl<'a> Generator<'a> {
         };
 
         let mut expr_buf = Buffer::new(0);
+        let borrow_val = !is_copyable(val);
+        if borrow_val {
+            expr_buf.write("&(");
```
GuillaumeGomez, line 878: Would it be worth it to not generate the parens if the expression is a literal or a field?

GuillaumeGomez (changes requested):
Please also add the stack overflow test you showed in your first comment:

```rust
fn f(x: u32) -> u32 {
    x * x + x
}

#[derive(askama::Template)]
#[template(source = "{{ self::f(self.x.0 + 2) }}", ext = "txt")]
struct CallFunction {
    x: (u32,),
}
```

askama_derive/src/generator.rs
PizzasBear: I don't think it'd be worth it. This isn't done in `Generator::_visit_args`

PizzasBear: Also literals are explicitly rejected in `is_copyable`.

askama_derive/src/generator.rs
```diff
@@ -1892,7 +1899,7 @@ fn is_copyable_within_op(expr: &Expr<'_>, within_op: bool) -> bool {
 pub(crate) fn is_attr_self(expr: &Expr<'_>) -> bool {
     match expr {
         Expr::Attr(obj, _) if matches!(obj.as_ref(), Expr::Var("self")) => true,
-        Expr::Attr(obj, _) if matches!(obj.as_ref(), Expr::Attr(..)) => is_attr_self(expr),
+        Expr::Attr(obj, _) if matches!(obj.as_ref(), Expr::Attr(..)) => is_attr_self(obj),
```
djc, line 1902: This seems unrelated? I'd like to have this in a separate commit with a test for it.

djc (approved):
This seems reasonable to me.

askama_derive/src/generator.rs
PizzasBear: I've already added the test as suggested by @GuillaumeGomez.

PizzasBear: I've splitted & squashed the commits.

djc:
In the future, please avoid commit messages like "Bugfix in" and describe the actual change.

PR #937, opened 2023-12-19
Only generate reference wrapping if there is no function/macro/method call

### LINKED ISSUE #935: Regression in askama_parser 0.2.0 update
Just updated askama_derive to 0.12.4 from 0.12.2 and askama_parser to 0.2.0 from 0.1.1 and now my project fails to compile.

```
error[E0277]: the trait bound `&std::string::String: From<&str>` is not satisfied
   --> askama_templates/src/templates.rs:202:10
    |
202 | #[derive(Template)]
    |          ^^^^^^^^
    |          |
    |          the trait `From<&str>` is not implemented for `&std::string::String`
    |          in this derive macro expansion
    |
   ::: .../.cargo/registry/src/index.crates.io-6f17d22bba15001f/askama_derive-0.12.4/src/lib.rs:22:1
    |
22  | pub fn derive_template(input: TokenStream) -> TokenStream {
    | --------------------------------------------------------- in this expansion of `#[derive(Template)]`
    |
    = help: the following other types implement trait `From<T>`:
              <std::string::String as From<char>>
              <std::string::String as From<bytestring::ByteString>>
              <std::string::String as From<smol_str::SmolStr>>
              <std::string::String as From<Box<str>>>
              <std::string::String as From<uuid::Uuid>>
              <std::string::String as From<Cow<'a, str>>>
              <std::string::String as From<url::Url>>
              <std::string::String as From<sqlx_postgres::types::citext::PgCiText>>
            and 4 others
    = note: required for `&str` to implement `Into<&std::string::String>`
```

My template struct:

```
#[derive(Template)]
#[template(path = "report.html")]
pub struct ReportTemplate<'a> {
    pub report_type: &'a str,
    pub report_date: &'a str,
}
```

GuillaumeGomez:
Fixes #935.

I added tests for: function call, method call, macro and filter. Having more regression tests would be nice I think. If you have other ideas that are currently failing, it'd be really nice to add them here.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -916,7 +916,7 @@ impl<'a> Generator<'a> {
         };
 
         let mut expr_buf = Buffer::new(0);
-        let borrow_val = !is_copyable(val);
+        let borrow_val = !is_copyable(val) && !has_call(val);
         if borrow_val {
             expr_buf.write("&(");
         }
@@ -1918,6 +1918,18 @@ impl<'a, K: Eq + hash::Hash, V> Default for MapChain<'a, K, V> {
             parent: None,
             scopes: vec![HashMap::new()],
         }
+    }
+}
+
+/// Returns `true` if an element of this `Expr` is a function/method call.
+fn has_call(expr: &Expr<'_>) -> bool {
+    use Expr::*;
+    match expr {
+        Filter(_, _) | Call(_, _) | RustMacro(_, _) => true,
+        Unary(_, expr) | Group(expr) | Try(expr) => has_call(expr),
+        Tuple(exprs) => exprs.iter().any(|expr| has_call(expr)),
+        BinOp(_, left_expr, right_expr) => has_call(left_expr) || has_call(right_expr),
+        _ => false,
     }
 }
 
--- a/testing/tests/vars.rs
+++ b/testing/tests/vars.rs
@@ -131,3 +131,31 @@ fn test_decl_assign_range() {
     let t = DeclAssignRange;
     assert_eq!(t.render().unwrap(), "1");
 }
+
+// This ensures that we do not wrap any call in a reference (which would prevent this template to
+// compile).
+#[derive(Template)]
+#[template(
+    source = r#"{% let my_string -%}
+{% if a == 1 -%}
+{% let my_string = format!("testing {}", true) -%}
+{% else if a == 2 -%}
+{% let my_string = "testing {}"|format(a) -%}
+{% else if a == 3 -%}
+{% let my_string = String::from("yop yop") -%}
+{% else -%}
+{% let my_string = "something else".into() -%}
+{% endif %}
+
+{{- my_string }}"#,
+    ext = "html"
+)]
+struct LetWithoutRef {
+    a: u32,
+}
+
+#[test]
+fn test_let_without_ref() {
+    let t = LetWithoutRef { a: 1 };
+    assert_eq!(t.render().unwrap(), "testing true");
+}


askama_derive/src/generator.rs
```diff
@@ -1921,6 +1921,18 @@ impl<'a, K: Eq + hash::Hash, V> Default for MapChain<'a, K, V> {
     }
 }
 
+/// Returns `true` if an element of this `Expr` is a function/method call.
+fn has_call(expr: &Expr<'_>) -> bool {
+    use Expr::*;
+    match expr {
+        Filter(_, _) | Call(_, _) | RustMacro(_, _) => true,
```
djc, line 1928: For each of these arms, can you briefly summarize your thinking on why you choose this outcome?

GuillaumeGomez: In a comment in the code or here? Let's start by only here and you'll tell me if you want it as a comment in the code:

For `Filter`, it basically calls a function (the filter), potentially on multiple levels (each filter becomes an argument to the next one). So with the reference, it'd be the same as doing `let x = &filter(blabla);`. Hence a reference to a temporary value.
For `Call`, it's a function call, so same as `Filter`.
For `RustMacro`, we have no idea what will actually end up being generated so better not try to wrap it in a reference either because the last macro element could very much be a function/method call (then back to argument for `Filter`).

Does my logic make sense?

c1wren:
I've tested this PR and I've found a case where this fails.

Here's a test case that doesn't work.
```
#[derive(Template)]
#[template(
    source = r#"{% let my_string -%}
{% if a == 1 -%}
{% let my_string = format!("testing {}", true) -%}
{% else if a == 2 -%}
{% let my_string = "testing {}"|format(a) -%}
{% else if a == 3 -%}
{% let my_string = String::from("yop yop") -%}
{% else if a == 4 -%}
{% let my_string = "something else".into() -%}
{% else -%}
{% let my_string = my_str -%}
{% endif %}
{{- my_string }}"#,
    ext = "html"
)]
struct LetWithoutRef {
    a: u32,
    my_str: String,
}
```

Last case is `&String` and all other cases are simply` String`. That's a type mismatch and will not compile.

I'm not sure that this is going to be a solvable problem without things getting much much more complicated.

GuillaumeGomez:
Agreed. Seems like we'll need to revert #922. Maybe allowing `&` in front of an expr could solve this. What do you think about that @djc?

djc:
@c1wren was that actually part of your project, or did you cook that up as a failure case?

c1wren:
@djc It is legitimately part of my project. I simply changed askama in my project to reference this PR to test it and it failed to compile again. So I then expanded the code and found that while the initial issues were resolved with this PR, it created other new issues. Some of my templates are absolutely insane in what I have to do. 

Also, I don't not like the implication that I'm personally trying to make this PR fail. I do not have the time nor the energy to do so.

GuillaumeGomez:
@djc I really think that this PR is taking the wrong turn here since it makes it more complicated to guess what the generated code looks like. Adding a reference by default (in #922) was a bad idea. Like I suggested, adding support to `&`  would allow to give more control to users over how their data is managed and wouldn't be too complicated to support (I think).

djc:
Okay, let's revert #922. I still think we should not support adding `&` in templates, though.

GuillaumeGomez:
> I still think we should not support adding & in templates, though.

A reason for that?

c1wren:
> I don't think I've asked before -- can you briefly describe what you're project is doing with Askama? Always good to understand how downstream code is using Askama, to help guide future design decisions.

@djc 

We use askama for doing two things: 1) creating emails and 2) generating html which is then turned into PDF's. The PDF html generation is where things get unruly due to the number of different settings permutations which control what has to be included. Some of our askama template files approach 1000 lines long and do funky things in order to achieve what is necessary.

djc:
> Some of our askama template files approach 1000 lines long and do funky things in order to achieve what is necessary.

Would be keen to hear about some of the uglier things you have to do to make things work, or in general if you have any suggestions on how to improve Askama!

djc:
> > I still think we should not support adding & in templates, though.
> 
> A reason for that?

IMO conceptually template code shouldn't be dealing with the complexities of referencing/dereferencing.

GuillaumeGomez:
Closing in favour of #938.

c1wren:
> > Some of our askama template files approach 1000 lines long and do funky things in order to achieve what is necessary.
> 
> Would be keen to hear about some of the uglier things you have to do to make things work, or in general if you have any suggestions on how to improve Askama!

@djc 

The most painful part is dealing with debugging templates when they won't compile and expanding the derive fails. When that happens, it's a painful game of removing random sections of template that you think might be the problem until you find the part that is the actual problem.

The lack of auto formatting also bothers me, but I don't think it's a problem you should feel like you need to solve. I know there are jinja template plugins and they somewhat help, but they aren't aware of nested ifs or matches.

djc:
> The most painful part is dealing with debugging templates when they won't compile and expanding the derive fails. When that happens, it's a painful game of removing random sections of template that you think might be the problem until you find the part that is the actual problem.

What editor are you using? I've found the "Expand macro recursively" action quite useful (from Rust-Analyzer via VS Code actions). Are you aware of the `print = "code"` option in the `#[template]` attribute?

c1wren:
> > The most painful part is dealing with debugging templates when they won't compile and expanding the derive fails. When that happens, it's a painful game of removing random sections of template that you think might be the problem until you find the part that is the actual problem.
> 
> What editor are you using? I've found the "Expand macro recursively" action quite useful (from Rust-Analyzer via VS Code actions). Are you aware of the `print = "code"` option in the `#[template]` attribute?

I mostly use VSCode and use the Expand macro recursively action quite a bit (when it decides to work). 

I have used `print = "code"` but that often doesn't work when there's a problem with the template. It's been a while since I've had to pound my head against the wall in debugging a template, so it's possible things have improved (either with rust or askama) without me knowing about it.

djc:
Okay, then I don't have any good ideas on how to further improve. Anyway, please submit any feedback in our issue tracker in the future!

PR #938, opened 2024-01-03
Revert "Add automatic borrowing to `let` statements"

GuillaumeGomez:
Reverts djc/askama#922

As discussed in https://github.com/djc/askama/pull/937, this was actually a pretty bad idea to do this. We'll need to discuss about potential ways to go around this limit but in the meantime, better revert it.

@PizzasBear: Can't you write a filter `as_ref` which takes a ref and returns a ref in the meantime to go around the limitation in your case?

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -916,14 +916,7 @@ impl<'a> Generator<'a> {
         };
 
         let mut expr_buf = Buffer::new(0);
-        let borrow_val = !is_copyable(val);
-        if borrow_val {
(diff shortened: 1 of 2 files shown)


PizzasBear:
I could use an `as_ref` filter instead, but I don't think you should revert the bugfix.

djc:
@GuillaumeGomez when you have a chance, could you remove the second commit here?

GuillaumeGomez:
Updated!

djc (approved):

Issue #719, opened 2022-09-02
askama with markdown filter using String: cannot move out of `self.name` which is behind a shared reference


schaschko:
I cloned the repo and then run this test:

https://github.com/djc/askama/blob/5748c357d435b24848d1571df010d777859fede9/askama_axum/tests/basic.rs

This works, then I modify basic.rs to (`String` instead of `&str`):
```
use askama::Template;
use axum::{
    body::Body,
    http::{Request, StatusCode},
    routing::get,
    Router,
};
use tower::util::ServiceExt;

#[derive(Template)]
#[template(path = "hello.html")]
struct HelloTemplate {
    name: String,
}

async fn hello() -> HelloTemplate {
    HelloTemplate { name: String::from("world") }
}

#[tokio::test]
async fn template_to_response() {
    let app = Router::new().route("/", get(hello));

    let res = app
        .oneshot(Request::builder().uri("/").body(Body::empty()).unwrap())
        .await
        .unwrap();
    assert_eq!(res.status(), StatusCode::OK);

    let headers = res.headers();
    assert_eq!(headers["Content-Type"], "text/html; charset=utf-8");

    let body = hyper::body::to_bytes(res.into_body()).await.unwrap();
    // assert_eq!(&body[..], b"Hello, world!");
}
```

Still works. Then I modify https://github.com/djc/askama/blob/5748c357d435b24848d1571df010d777859fede9/askama_axum/templates/hello.html to:
```
Hello, {{ name|markdown }}!
```

If I don't use the markdown filter, e.g. `Hello, {{ name }}!`, then it works. How to fix this error?

My goal was to be able to read in markdown files during runtime (with `fs::read_to_string( >path-to-markdown-file.md< )` which gives `String`).

Thank you

djc:
I think this is because `askama::filters::markdown` takes `S: AsRef<str>` when it should probably really just be taking `&str`. Do you want to try if that improves things, and if so, submit a PR?

plainlevel:
```
async fn my_posts(State(pool): State<PgPool>,) -> impl IntoResponse {
    #[derive(FromRow)]
    struct Post {
        title: String,
        post_body: String,
    }

    let my_post = sqlx::query_as::<_, Post>("select title, post_body from my_posts")
        .fetch_all(&pool)
        .await
        .unwrap();

    // my template looks like this:
    // #[derive(Template)]
    // #[template(path = "post.html"]
    // pub struct MyTemplate {
    //     pub post_title: String,
[...]

PR #941, opened 2024-01-09
Make `String` compatible with the markdown filter.

Wilkuu-2:
Hello, I recently decided to switch from Tera to Askama, and I'm enjoying it so far. 

Yesterday, I stumbled upon issue #719 and decided to see if I can fix the issue with the markdown filter, so I don't have to include Comrak in my own crate as a workaround. 
What I found out is that the issue with using `String` is that the filter moves the value out of the struct. This is fine for `&str` considering `&str` is already a reference. 
```rs
    /// Code generated by Template derive macro  
    ...
            expr0 = &::askama::filters::markdown(::askama::Html, self.content, ::core::option::Option::None)?,
            //                                                   ^^^^^^^^^^^^ 
            //                                              Value gets moved out by filter 
    ...
```

Making the generated code borrow the value did not create any problems in tests.

This PR also adds a test for passing `String`s to the markdown filter and simplifies the function
signature for the filter to accept `&str`.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1200,7 +1200,7 @@ impl<'a> Generator<'a> {
         };
 
         buf.write(&format!(
-            "::askama::filters::markdown({}, ",
+            "::askama::filters::markdown({}, &",
             self.input.escaper
         ));
         self.visit_expr(buf, md)?;
--- a/testing/tests/markdown.rs
+++ b/testing/tests/markdown.rs
@@ -70,3 +70,21 @@ before\
 after",
     );
 }
+
+#[derive(Template)]
+#[template(source = "{{content|markdown}}", ext = "html")]
+struct MarkdownStringTemplate {
+    content: String,
+}
+
+// Tests if the markdown filter accepts String
+#[test]
+fn test_markdown_owned_string() {
+    let template = MarkdownStringTemplate {
+        content: "The markdown filter _indeed_ works with __String__".into(),
+    };
+    assert_eq!(
+        template.render().unwrap(),
+        "<p>The markdown filter <em>indeed</em> works with <strong>String</strong></p>\n"
+    )
+}


askama_derive/src/generator.rs
```diff
@@ -1200,7 +1200,7 @@ impl<'a> Generator<'a> {
         };
 
         buf.write(&format!(
-            "::askama::filters::markdown({}, ",
+            "::askama::filters::markdown({}, &",
```
GuillaumeGomez, line 1203: Normally, only this change should be needed and not the rest. Also a lot more types are actually implementing `AsRef<str>`, so this is a regression in a sense since now it only works with `String` and `str`.

djc: Yeah, that's exactly what I'm worried about too.

askama/src/filters/mod.rs
```diff
@@ -342,14 +342,14 @@ pub fn wordcount<T: fmt::Display>(s: T) -> Result<usize> {
 }
 
 #[cfg(feature = "markdown")]
-pub fn markdown<E, S>(
+pub fn markdown<E,S>(
```
djc, line 345: You've left a bunch of trivial changes in this PR, please actually revert the changes to the previous state.

testing/tests/markdown.rs
```diff
@@ -70,3 +71,18 @@ before\
...
     );
 }
+
+#[derive(Template)]
+#[template(source = "{{content|markdown}}", ext = "html", print="code")]
+struct MarkdownStringTemplate{
+    pub content: String
+}
+
+#[test]
```
GuillaumeGomez, line 81: Could add a comment saying exactly what this test is checking please? Much better if this breaks in the future to understand what the new changes did wrong.

Wilkuu-2: I actually wanted to ask if this test is necessary. Since it only tests if the markdown filter accepts `String`.

GuillaumeGomez: Well, your use-case is valid and I think it's important for us to ensure that it won't regress again. Nothing prevents you to add more checks in this test though. ;)

Wilkuu-2: It's good to hear that the test can stay. 

I think this check is enough for this use-case and this PR. I'll consider adding more checks in a future PR though.

GuillaumeGomez: Fine by me. 👍

GuillaumeGomez:
Ok, now please run `cargo fmt` and then [squash your commits](https://blog.guillaume-gomez.fr/Git-tips/1/2).

djc:
@GuillaumeGomez if there is a single conceptual change we don't need the submitter to squash -- we can just "Squash and merge" instead.

GuillaumeGomez (approved):

GuillaumeGomez:
Ah true. I keep forgetting about it.

djc:
Thanks!

Apollo-XIV:
fixed an issue for me that almost made me swap to a different templating engine before i saw this thread! thank you all :D

PR #942, opened 2024-01-10
parser: limit target recursion

djc:
#911 introduced support for multiple targets, but I missed the need to limit recursion (see #866), which triggered a bunch of issues from OSS Fuzz.


GuillaumeGomez (changes requested):
Changes look good but please add a regression test to ensure it does indeed fail compilation as expected.

djc:
Added a test case.

Kijewski (approved):

GuillaumeGomez (approved):
All good for me, thanks!

Issue #894, opened 2023-10-29
Problems comparing integers


pliniocampinas:
Hi, I just started to test askama but Im having all sorts of problems with integers within a template file. For example, the following code:

 ```
 {% for solution in solutions %}
  <div class="table">
    {% for row in 0..8 %}
    <div class="row">
      {% for col in 0..8 %}
      <div class="col">
        {% let has_queen = (solution[row] + 0) == col %}
        {% if has_queen -%}
          <img class="queen-icon" src="/_assets/chess-queen.svg" alt="Q">
        {%- endif -%}
      </div>
      {% endfor %}
    </div>
    {% endfor %}
  </div>
  {% endfor %}
```

  ` {% let has_queen = (solution[row] + 0) == col %}` Results on an error: no implementation for `&isize == {integer}`
  
  But if I just add zero:

    {% let has_queen = (solution[row] + 0) == col %}
    
It compiles.

I also had problems comparing integers. Resulting on errors as `no implementation for `isize == usize``.

I had no success trying any kind of convertion. Am I missing something here?

vallentin:
The reason adding 0 works, but comparing doesn't is because Rust [implements `Add` for references](https://doc.rust-lang.org/std/ops/trait.Add.html#implementors) as well, out of convenience. The same is however not true for [`PartialEq`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html#implementors). So when you add 0, it's able to perform `&T + T -> T`. Which then allows `T == T` to compile. Whereas `&T == T` does not compile.

So why does this happen? In Askama we're trying to hide away references and borrowing. So some expressions automatically borrows values. For common cases, this is usually fine. But the more logic is being added to a template, then you can run into issues like yours.

The easiest way to fix your issue, is instead of using `==`, then replace it with `.eq()`, i.e. `{% let has_queen = solution[row].eq(col) %}` works fine (assuming `T: PartialEq<Rhs>` was already valid).

When it comes to putting a lot of logic into a template. Then custom [filters](https://djc.github.io/askama/filters.html#custom-filters) or otherwise helper functions or helper traits, can help a lot to move logic out of templates. In your example that isn't by any means a lot of logic though, you just ran into one of the edge cases.

PR #943, opened 2024-01-10
Add `as_ref` builtin

### LINKED ISSUE #330: Add support for reference operator (i.e. `{% let x = &y %}`)

GuillaumeGomez:
Fixes #330.

Following discussions with @djc, to keep `askama` template syntax as simple as possible and to go around the limitations described in #330, a good way to fix it is to add a `as_ref` builtin filter.

### DIFF
--- a/book/src/filters.md
+++ b/book/src/filters.md
@@ -21,6 +21,7 @@ but are disabled by default. Enable them with Cargo features (see below for more
 
 * **[Built-in filters][#built-in-filters]:**  
   [`abs`][#abs],
+  [`as_ref`][#as_ref],
   [`capitalize`][#capitalize],
   [`center`][#center],
   [`escape|e`][#escape],
@@ -60,6 +61,23 @@ Output:
 
 ```
 2
+```
+
+### as_ref
+[#as_ref]: #as_ref
+
+Creates a reference to the given argument.
+
+```
+{{ "a"|as_ref }}
+{{ self.x|as_ref }}
+```
+
+will become:
+
+```
+&a
+&self.x
 ```
 
 ### capitalize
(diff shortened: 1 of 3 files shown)


djc:
Seems kinda reasonable to me. @Kijewski @vallentin any thoughts?

Kijewski:
The PR is well implemented, and (except for the small clippy warning), LGTM.

But another alternative would be to add prefix operators `&` and `*` to the language. This could also be a fix for e.g. #894. Both options would work for me, this PR or adding `&`+`*`, and I have no strong preferences.

djc:
> But another alternative would be to add prefix operators `&` and `*` to the language. This could also be a fix for e.g. #894. Both options would work for me, this PR or adding `&`+`*`, and I have no strong preferences.

I'm not a fan of that particular alternative: https://github.com/djc/askama/pull/937#issuecomment-1874648650. I don't have much of an objection to this approach, although I don't love it.

GuillaumeGomez:
Fixed the clippy lint. :)

book/src/filters.md
```diff
@@ -62,6 +63,23 @@ Output:
 2
 ```
 
+### as_ref
+[#as_ref]: #as_ref
+
+Put the given argument behind a reference.
```
djc, line 69: Okay, one more nit: "Put .. behind a reference" doesn't seem like language that is commonly used for this? I'd describe it as maybe "Creates a reference to ..".

djc (approved):

book/src/filters.md
GuillaumeGomez: :+1:

GuillaumeGomez:
Updated sentence.

djc:
There's still/again a clippy failure? (I approved this already, so feel free to go ahead and merge once it passes CI.)

PR #944, opened 2024-01-11
Clean up filter generation

Kijewski:


askama_derive/src/generator.rs
```diff
@@ -1210,66 +1210,106 @@ impl<'a> Generator<'a> {
...
+        if args.len() != 1 {
+            return Err("unexpected argument(s) in `safe` filter".into());
+        }
+        buf.write("::askama::filters::yaml(");
+        buf.write(self.input.escaper);
+        buf.write(", ");
+        self._visit_args(buf, args)?;
+        buf.write(")?");
+        Ok(DisplayWrap::Wrapped)
+    }
```
GuillaumeGomez, line 1261: Wouldn't it be simpler to instead have:

```suggestion
    fn _visit_yaml_filter(
        &mut self,
        buf: &mut Buffer,
        args: &[Expr<'_>],
    ) -> Result<DisplayWrap, CompileError> {
        if cfg!(not(feature = "serde-yaml")) {
            return Err("the `yaml` filter requires the `serde-yaml` feature to be enabled".into());
        }
        if args.len() != 1 {
            return Err("unexpected argument(s) in `safe` filter".into());
        }
        buf.write("::askama::filters::yaml(");
        buf.write(self.input.escaper);
        buf.write(", ");
        self._visit_args(buf, args)?;
        buf.write(")?");
        Ok(DisplayWrap::Wrapped)
    }
```

?

askama_derive/src/generator.rs
```diff
@@ -1210,66 +1210,106 @@ impl<'a> Generator<'a> {
...
+    #[cfg(not(feature = "serde-json"))]
+    fn _visit_json_filter(
+        &mut self,
+        _: &mut Buffer,
+        _: &[Expr<'_>],
+    ) -> Result<DisplayWrap, CompileError> {
+        Err("the `json` filter requires the `serde-json` feature to be enabled".into())
+    }
+
+    #[cfg(feature = "serde-json")]
```
GuillaumeGomez, line 1222: Same as below.

askama_derive/src/generator.rs
Kijewski: Yes, that's much better!

askama_derive/src/generator.rs
```diff
@@ -1210,10 +1210,34 @@ impl<'a> Generator<'a> {
         Ok(DisplayWrap::Wrapped)
     }
 
+    #[cfg(not(feature = "serde-json"))]
+    fn _visit_json_filter(
```
djc, line 1209: Style nit: please follow top-down ordering for this -- `_visit_json_filter()` should go below `visit_filter()`. (Same for the others.)

djc (approved):
This is great, thanks!

askama_derive/src/generator.rs
Kijewski: Moved around.

Kijewski:
`_visit_as_ref_filter()` should return `DisplayWrap::Unwrapped`, shouldn't it? E.g. `{{ "<script>"|as_ref }}`. @GuillaumeGomez

djc:
Ping?

djc:
Yes, definitely! That's actually a security risk...

Kijewski:
Rebased.

GuillaumeGomez (approved):

PR #946, opened 2024-01-11
Generator: make `normalize_identifier` faster

Kijewski:
`normalize_identifier` is called quite often in the generator, once for every variable name or path element that is written.

This PR aims to speed up the function by

* using a per-input string length replacement map
* binary searching the replacement map instead of a linear search

Diffent, but functionally equivalent implementations were compared:

```text
* linear search in one big map:      348.44 µs
* binary search in one big map:      334.46 µs
* linear search in a per-length map: 178.84 µs
* binary search in a per-length map: 154.54 µs
* perfect hashing:                   170.87 µs
```

The winner of this competition is "binary search in a per-length map". It does not introduce new dependencies, but has the slight disadvantage that it uses one instance of `unsafe` code. I deem this disadvantage acceptable, though.

Nb. It was also tested if a variant that only stores the replaced string would be faster. This "optimization" proved to be slower for all implementations except "binary search in a per-length map", for which it has the same runtime. Without a clear advantage to use the "optimized version", I chose to use the more easy to read slice of tuples variant.

Obviously, for all measurements: YMMV.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -2025,7 +2025,7 @@ enum Writable<'a> {
 // language.
 fn normalize_identifier(ident: &str) -> &str {
     // This table works for as long as the replacement string is the original string
-    // prepended with "r#". The strings bet right-padded to the same length with b'_'.
+    // prepended with "r#". The strings get right-padded to the same length with b'_'.
     // While the code does not need it, please keep the list sorted when adding new
     // keywords.
 
@@ -2100,7 +2100,11 @@ fn normalize_identifier(ident: &str) -> &str {
         while i < KWS.len() {
             let mut j = 0;
             while KWS[i].len() < j {
-                assert!(KWS[i][j].is_ascii());
+                let mut k = 0;
+                while KWS[i][j].len() < k {
+                    assert!(KWS[i][j][k].is_ascii());
+                    k += 1;
+                }
                 j += 1;
             }
             i += 1;
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -2023,60 +2023,81 @@ enum Writable<'a> {
 // because they are not allowed to be raw identifiers, and *loop*
 // because it's used something like a keyword in the template
 // language.
-static USE_RAW: [(&str, &str); 47] = [
-    ("as", "r#as"),
-    ("break", "r#break"),
-    ("const", "r#const"),
-    ("continue", "r#continue"),
-    ("crate", "r#crate"),
-    ("else", "r#else"),
-    ("enum", "r#enum"),
-    ("extern", "r#extern"),
-    ("false", "r#false"),
-    ("fn", "r#fn"),
-    ("for", "r#for"),
-    ("if", "r#if"),
-    ("impl", "r#impl"),
-    ("in", "r#in"),
-    ("let", "r#let"),
-    ("match", "r#match"),
-    ("mod", "r#mod"),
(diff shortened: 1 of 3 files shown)


Kijewski:
I guess the benchmark for the different implementations does not need to be checked in, but I thought it might be good if I actually prove that my claims are not totally made up. :)

I would be interested if how the benchmark performs on different platforms, so if you have a non-AMD x64 CPU, or event a non-x64 CPU, then please post your results!

testing/benches/normalize_identifier.rs
```diff
@@ -0,0 +1,678 @@
+use std::collections::HashSet;
+use std::iter::FusedIterator;
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+
+criterion_main!(benches);
+criterion_group!(benches, functions);
+
+fn functions(c: &mut Criterion) {
+    let words: Vec<_> = Words::default().collect();
```
djc, line 10: Nit: prefer turbofishing over type annotation?

djc (commented):
Thanks for working on this! Agree that this way of using unsafe is fine.

Benchmark results from my Apple Silicon M1 Max:

* normalize_identifier_linear: 269.38 µs
* normalize_identifier_linear_replacement_only: 369.72 µs
* normalize_identifier_binary_search: 233.24 µs
* normalize_identifier_binary_search_replacement_only: 255.46 µs
* normalize_identifier_linear_by_len: 152.49 µs
* normalize_identifier_linear_by_len_replacement_only: 152.81 µs
* normalize_identifier_bisect_by_len: 101.48 µs
* normalize_identifier_bisect_by_len_replacement_only: 108.80 µs
* normalize_identifier_phf: 115.29 µs

Did you somehow get a sense of an average overall latency improvement this might have for procedural macro runs?

Kijewski:
I optimized the implementation to compare arrays `[u8; _]` instead of slices `&[u8]`. The new `normalize_identifier_linear_by_len_replacement_only()` takes only 79.167µs on my machine.

testing/benches/normalize_identifier.rs
Kijewski: Done.

askama_derive/src/generator.rs
```diff
@@ -2023,60 +2023,81 @@ enum Writable<'a> {
...
-    ("yield", "r#yield"),
-    ("try", "r#try"),
-];
-
 fn normalize_identifier(ident: &str) -> &str {
-    if let Some(word) = USE_RAW.iter().find(|x| x.0 == ident) {
-        word.1
-    } else {
-        ident
+    const KW0: &[([u8; 8], [u8; 10])] = &[];
```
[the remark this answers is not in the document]
Kijewski: I replaced the function with a variant that only stores the replacement string.

djc:
> I optimized the implementation to compare arrays `[u8; _]` instead of slices `&[u8]`. The new `normalize_identifier_linear_by_len_replacement_only()` takes only 79.167µs on my machine.

Cool, that one is the fastest here, too.

djc (commented):
Could you add some documentation about what invariants this function tries to uphold?

Kijewski:
I added a few comments, and made sure that all strings are ASCII.

djc (approved):

Kijewski:
Rust 1.60 is almost 2 years old. I guess we can raise the msrv, so I can use [`[u8].is_ascii()`](https://caniuse.rs/features/slice_escape_ascii), what do you say?

PR #947, opened 2024-01-14
parser: add missing license files

decathorpe:
It looks like adding the symlinks to the license files was missed when askama_parser crate was added, since all other crates in this project have them.

If possible, it would be great if you could release a new version of askama_parser that contains the license texts. I'm currently working on updating the packages for the askama* crates in Fedora Linux, and I'm currently blocked on published versions of askama_parser not containing the required license texts.

### DIFF
--- a/askama_parser/LICENSE-APACHE
+++ b/askama_parser/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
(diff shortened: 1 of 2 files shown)


Kijewski (approved):

GuillaumeGomez (approved):

djc:
Thanks! I'll see about getting a release out soon.

djc:
This has been released as part of askama_parser 0.2.1.

PR #951, opened 2024-01-18
Fix support for mixed case variables

### LINKED ISSUE #924: 0.12.1 only allow lower-case variable name in template
I don't know if I asked a stupid question, but I spend one day to find it ! my rust version is 1.74.0, I run it on my win11 and WSL ubuntu 22.04, see the same error below
```rust
use askama::Template;

#[derive(Template)] 
struct HelloTemplate<'a> {
    nAme: &'a str, 
}
fn main() {
    let hello = HelloTemplate { nAme: "world" }; 
    println!("{}", hello.render().unwrap());
}
```
```html
Hello, {{ nAme }}!
```
```toml
[package]
name = "hello"
version = "0.1.0"
edition = "2021"

[dependencies]
askama="0.12.1"
```
[xxx.DESKTOP-ABCDE12] ⮞ cargo run -p hello
   Compiling hello v0.1.0 (C:\xx\xx\xx\hello)
error[E0425]: cannot find value `nAme` in this scope
 --> hello\src\main.rs:3:10
  |
3 | #[derive(Template)]
  |          ^^^^^^^^
  |
help: you might have meant to use the available field
  |
3 | #[derive(self.Template)]
  |          +++++

For more information about this error, try `rustc --explain E0425`.
error: could not compile `hello` (bin "hello") due to previous error

djc:
Fixes #924. Looks like this might have regressed in #840. I think checking only the first letter is a better heuristic than checking if any letters are uppercase.

### DIFF
--- a/testing/tests/simple.rs
+++ b/testing/tests/simple.rs
@@ -484,3 +484,23 @@ fn test_num_literals() {
         "[90, -90, 90, 2, 56, 240, 10.5, 10.5, 100000000000, 105000000000]",
     );
 }
+
+#[allow(non_snake_case)]
+#[derive(askama::Template)]
+#[template(source = "{{ xY }}", ext = "txt")]
+struct MixedCase {
+    xY: &'static str,
+}
+
+/// Test that we can use mixed case in variable names
+///
+/// We use some heuristics to distinguish paths (`std::str::String`) from
+/// variable names (`foo`). Previously, this test would fail because any
+/// name containing uppercase characters would be considered a path.
+///
+/// https://github.com/djc/askama/issues/924
+#[test]
+fn test_mixed_case() {
+    let template = MixedCase { xY: "foo" };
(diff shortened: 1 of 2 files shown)


testing/tests/simple.rs
```diff
@@ -484,3 +484,16 @@ fn test_num_literals() {
         "[90, -90, 90, 2, 56, 240, 10.5, 10.5, 100000000000, 105000000000]",
     );
 }
+
+#[allow(non_snake_case)]
```
GuillaumeGomez, line 488: Would be nice to add comments explaining what this is testing and also eventually link to the issue.

GuillaumeGomez (commented):
Apart from my nit, looks good to me.

testing/tests/simple.rs
djc: Added.

GuillaumeGomez (approved):
Great, thanks!

PR #953, opened 2024-01-19
Remove `syn` and `proc-macro2` and `quote` dependencies from `askama_derive`

GuillaumeGomez:
So overall, I think it'll be much more efficient to parse the derive proc-macro (not sure exactly how to test it...) as this code doesn't try to make sense of most tokens it parses.

However, and I think it's quite important to note: it has a maintenance cost (as does any code) and if you don't think it's worth it, we can close it. The proc-macro performance isn't that critical and the three dependencies are still in the dependency tree since `askama` uses `serde`.

Even if not merged, some improvements in the code can still be retrieved from this code, mostly around how the code is generated. I think that splitting code generation into a lot of small calls to `writeln` makes it much more difficult to have an idea of what the generated code will look like.

Anyway, it was fun to do. Parsing rustc tokens is pretty easy so I did it originally to see how hard it would be to not use `syn` and all the crates (which are great, but because they need to support a lot of cases, are bound to be slower than hand-made code, specific to one use case).

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -196,112 +196,70 @@ impl<'a> Generator<'a> {
     // Implement `From<Template> for hyper::Response<Body>` and `From<Template> for hyper::Body.
     #[cfg(feature = "with-hyper")]
     fn impl_hyper_into_response(&mut self, buf: &mut Buffer) -> Result<(), CompileError> {
-        let (impl_generics, orig_ty_generics, where_clause) =
-            self.input.ast.generics.split_for_impl();
-        let ident = &self.input.ast.ident;
+        let generics_with_bounds = self.input.ast.generics.display_with_bounds();
+        let orig_ty_generics = &self.input.ast.generics;
+        let type_name = &self.input.ast.ident;
+        let where_clause = &self.input.ast.where_clause;
+
         // From<Template> for hyper::Response<Body>
         buf.writeln(&format!(
-            "{} {{",
-            quote!(
-                impl #impl_generics ::core::convert::From<&#ident #orig_ty_generics>
-                for ::askama_hyper::hyper::Response<::askama_hyper::hyper::Body>
-                #where_clause
-            )
-        ))?;
-        buf.writeln("#[inline]")?;
-        buf.writeln(&format!(
-            "{} {{",
-            quote!(fn from(value: &#ident #orig_ty_generics) -> Self)
-        ))?;
-        buf.writeln("::askama_hyper::respond(value)")?;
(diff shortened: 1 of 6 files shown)


askama_derive/src/generator.rs
```diff
@@ -105,12 +104,7 @@ impl<'a> Generator<'a> {
             };
             if path_is_valid {
                 let path = path.to_str().unwrap();
-                buf.writeln(
-                    &quote! {
-                        include_bytes!(#path);
-                    }
-                    .to_string(),
-                )?;
+                buf.writeln(&format!("include_bytes!(\"{path}\");"))?;
```
Kijewski, line 107: Maybe ↓ incase the name contains backslashes or quotation marks?
```rust
format!("include_bytes!({path:?});")
```

GuillaumeGomez: I was wondering about it but your solution is just super elegant. Thanks! :)

GuillaumeGomez:
And fixed mendes (forgot to generate with the where clause).

djc:
So what is/was the motivation for this? What are you hoping to get out of it? It's a little hard to get excited about a +780/-230 diff.

GuillaumeGomez:
Performance mostly. Do you have a way to measure it for `askama_derive`?

Kijewski:
> I think that splitting code generation into a lot of small calls to writeln makes it much more difficult to have an idea of what the generated code will look like.

Currently every line is written in its own statement, because the code is generated with (more or less) correct indentation. I don't know if this feature is actually needed, because "cargo expand" reformats the resulting code anyway. Using multiline strings is more readable, if we can live without our manual indenting.

askama_derive/src/unescape.rs
```diff
@@ -0,0 +1,158 @@
+// This code is a simplified version of the `rustc_lexer` unescape.
```
Kijewski, line 1: Hm, sorry, but I quite dislike this file. It kinda gives me paranoia that we miss some obscure corner case, which could lead to the rise of Skynet. :)

GuillaumeGomez: Don't believe in me, believe in the rust compiler people who wrote it. :laughing:

djc: Yeah, but then by copying this we no longer benefit from their continued maintenance.

GuillaumeGomez: Code hasn't changed in a long time, so we should be fine on that end. But yes, would be much better if we could get the "string value" of a raw string directly from proc-macro crate instead of having to handle it ourselves...

PR #954, opened 2024-01-22
Improve error when whitespace control is used on extends

GuillaumeGomez:
Took me a while to understand why `askama` was complaining about an `extends`. Realized that it was because whitespace control is not allowed on it. Even though it does nothing, I think it's not great user experience to that this opaque error with valid jinja code so I updated the parser to allow it.

### DIFF
--- a/askama_parser/src/node.rs
+++ b/askama_parser/src/node.rs
@@ -928,8 +928,22 @@ pub struct Extends<'a> {
 
 impl<'a> Extends<'a> {
     fn parse(i: &'a str) -> ParseResult<'a, Self> {
-        let (i, path) = preceded(ws(keyword("extends")), cut(ws(str_lit)))(i)?;
-        Ok((i, Self { path }))
+        let start = i;
+
+        let (i, (pws, _, (path, nws))) = tuple((
+            opt(Whitespace::parse),
+            ws(keyword("extends")),
+            cut(pair(ws(str_lit), opt(Whitespace::parse))),
+        ))(i)?;
+        match (pws, nws) {
+            (None, None) => Ok((i, Self { path })),
+            (_, _) => Err(nom::Err::Failure(ErrorContext {
+                input: start,
+                message: Some(Cow::Borrowed(
+                    "whitespace control is not allowed on `extends`",
+                )),
+            })),
+        }
     }
 }
 
--- a/book/src/template_syntax.md
+++ b/book/src/template_syntax.md
@@ -328,6 +328,16 @@ blocks from the base template with those from the child template. Inside
 a block in a child template, the `super()` macro can be called to render
 the parent block's contents.
 
+Because top-level content from the child template is thus ignored, the `extends`
+tag doesn't support whitespace control:
+
+```html
+{%- extends "base.html" +%}
+```
+
+The above code is rejected because we used `-` and `+`. For more information
+about whitespace control, take a look [here](#whitespace-control).
+
 ## HTML escaping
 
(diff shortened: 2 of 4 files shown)


djc:
I think I'd prefer to improve the error message instead of allowing it in the parser as it could be confusing to have it doing nothing.

djc:
I think the whitespace outside `extends` is always ignored because the top-level flow comes from the base template, and the child template only picks up things inside blocks. So it is actually ignored AFAIK.

djc:
How hard would it be to emit a good error message for this? Maybe just parse it like it's allowed but then emit an error from the parser if any of the whitespace controls are non-default for `extends`?

GuillaumeGomez:
Pretty easy. Updating.

GuillaumeGomez:
Done,

askama_parser/src/node.rs
```diff
@@ -928,8 +928,23 @@ pub struct Extends<'a> {
...
+        if pws.is_some() || nws.is_some() {
+            Err(nom::Err::Failure(ErrorContext {
+                input: start,
+                message: Some(Cow::Borrowed(
+                    "whitespace control is not allowed on `extends`",
+                )),
+            }))
+        } else {
+            Ok((i, Self { path }))
+        }
```
djc, line 947: Nit: I'd prefer to either write this with an early return (probably for the `Ok` case) or just as a match:

```rust
match (pws, nws) {
     (None, None) => Ok((i, Self { path })),
     (_, _) => Err(..),
}
```

askama_parser/src/node.rs
```diff
@@ -928,8 +928,23 @@ pub struct Extends<'a> {
...
+        let (i, (pws, _, (path, nws))) = tuple((
+            opt(Whitespace::parse),
+            ws(keyword("extends")),
+            cut(pair(ws(str_lit), opt(Whitespace::parse))),
+        ))(i)?;
+        if pws.is_some() || nws.is_some() {
+            Err(nom::Err::Failure(ErrorContext {
+                input: start,
+                message: Some(Cow::Borrowed(
+                    "whitespace control is not allowed on `extends`",
```
djc, line 942: Should this mention something about why this is the case?

GuillaumeGomez: Hum... I think it'd better to mention this in the book instead.

askama_parser/src/node.rs
GuillaumeGomez: :+1:

askama_parser/src/node.rs
djc: I can live with that.

GuillaumeGomez:
Done.

book/src/template_syntax.md
```diff
@@ -328,6 +328,15 @@ blocks from the base template with those from the child template. Inside
 a block in a child template, the `super()` macro can be called to render
 the parent block's contents.
 
+This is also why the `extends` tag doesn't support whitespace control:
```
djc, line 331: I feel like the "This" reference isn't that obvious. Maybe "Because top-level content from the child template is thus ignored, the `extends tag` ..."?

djc (approved):
Sorry for all the nits!

GuillaumeGomez:
> Sorry for all the nits!

Nah it's fine. I don't mind the review rounds.

book/src/template_syntax.md
GuillaumeGomez: It's better indeed. Updated!

PR #900, opened 2023-11-06
Add Poem integration

cvng:
Add Poem integration based on a previous pull request (#563) which I modified to match the Axum integration. 

Happy to make any necessary changes if you see fit.

PR #957, opened 2024-01-24
Add poem integration

rotmoded:


djc:
See #900. I don't think my stance on this has changed...

Issue #955, opened 2024-01-22
Add support for filter section


GuillaumeGomez:
Would you be open to add support for [filter section](https://jinja.palletsprojects.com/en/3.1.x/templates/#id11)?

Kijewski:
So, this filter first collects all produced text into a buffer, e.g. String, then the collected data is passed to the filter? How does this work with escaped data?

```jinja2
{% filter upper %}
{{ "<script>" }}
{% endfilter %}
```
→
```html
&LT;SCRIPT&GT;
```
because `"<script>"` gets escaped eagerly? Or
```html
&lt;SCRIPT&gt;
```
because the escaping is only done after the whole section was processed?

Could you tell me how you want to use a filter section?

GuillaumeGomez:
I suppose that if we are in "html mode", we want it to be escaped indeed. So generated code would actually look like:

```
let _x = escape_html(format!("<script>"));
```

I need it for the tera to askama migration in docs.rs. Examples:

 * https://github.com/rust-lang/docs.rs/blob/4b3a59af7b7ae4f234537035912acb0adf721b37/templates/crate/build_details.html#L33
 * https://github.com/rust-lang/docs.rs/blob/4b3a59af7b7ae4f234537035912acb0adf721b37/templates/core/about/metadata.html#L17

djc:
I feel like this has come up before, but I can't find any prior issues discussing it.

What alternatives have you come up with?

I don't like it that much, on account of the idea that I don't feel like the code generated it for it is conceptually obvious. But, I can't think of a strong reason to reject it, either... I notice that your use cases are relying on filters we haven't implemented yet (`dedent` and `highlight`), do you have have an implementation for those already?

GuillaumeGomez:
Yes we already have them. Conceptually, I see it as very close as what we do for the rest in general: we generate a `format!` (in a sub-block would be best I think) and call filters on it and "return" the result of all this that we put either in a variable that will be used later on.

djc:
Okay, I'm open to reviewing changes for this. I assume we'll collect, filter, then escape as @Kijewski suggested.

PR #959, opened 2024-01-26
Add support for filter block

GuillaumeGomez:
Fixes https://github.com/djc/askama/issues/955.

This PR still lacks documentation and some more tests would be nice too (if you have ideas?). But I think it's already good enough to go through a first round of reviews.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1068,10 +1068,9 @@ impl<'a> Generator<'a> {
         cacheable: bool,
         expr_cache: &mut HashMap<String, usize>,
     ) -> Result<usize, CompileError> {
-        use self::DisplayWrap::*;
         let expression = match wrapped {
-            Wrapped => expr,
-            Unwrapped => format!(
+            DisplayWrap::Wrapped => expr,
+            DisplayWrap::Unwrapped => format!(
                 "{CRATE}::MarkupDisplay::new_unsafe(&({}), {})",
                 expr, self.input.escaper
             ),
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::hash_map::{Entry, HashMap};
 use std::path::Path;
 use std::{cmp, hash, mem, str};
@@ -22,7 +23,7 @@ pub(crate) struct Generator<'a> {
     // The heritage contains references to blocks and their ancestry
     heritage: Option<&'a Heritage<'a>>,
     // Variables accessible directly from the current scope (not redirected to context)
-    locals: MapChain<'a, &'a str, LocalMeta>,
+    locals: MapChain<'a, Cow<'a, str>, LocalMeta>,
     // Suffix whitespace from the previous literal. Will be flushed to the
     // output buffer unless suppressed by whitespace suppression on the next
     // non-literal.
@@ -43,7 +44,7 @@ impl<'a> Generator<'a> {
         input: &'n TemplateInput<'_>,
         contexts: &'n HashMap<&'n Path, Context<'n>>,
         heritage: Option<&'n Heritage<'_>>,
-        locals: MapChain<'n, &'n str, LocalMeta>,
+        locals: MapChain<'n, Cow<'n, str>, LocalMeta>,
     ) -> Generator<'n> {
         Generator {
             input,
@@ -635,7 +636,7 @@ impl<'a> Generator<'a> {
                         "no argument named `{arg_name}` in macro {name:?}"
                     )));
                 }
-                named_arguments.insert(arg_name, arg);
+                named_arguments.insert(Cow::Borrowed(arg_name), arg);
             }
         }
 
@@ -648,7 +649,7 @@ impl<'a> Generator<'a> {
         //   anything since named arguments are always last).
         let mut allow_positional = true;
         for (index, arg) in def.args.iter().enumerate() {
-            let expr = match named_arguments.get(&arg) {
+            let expr = match named_arguments.get(&Cow::Borrowed(arg)) {
                 Some(expr) => {
                     allow_positional = false;
                     expr
@@ -671,14 +672,16 @@ impl<'a> Generator<'a> {
                 // to avoid moving non-copyable values.
                 &Expr::Var(name) if name != "self" => {
                     let var = self.locals.resolve_or_self(name);
-                    self.locals.insert(arg, LocalMeta::with_ref(var));
+                    self.locals
+                        .insert(Cow::Borrowed(arg), LocalMeta::with_ref(var));
                 }
                 Expr::Attr(obj, attr) => {
                     let mut attr_buf = Buffer::new(0);
                     self.visit_attr(&mut attr_buf, obj, attr)?;
 
                     let var = self.locals.resolve(&attr_buf.buf).unwrap_or(attr_buf.buf);
-                    self.locals.insert(arg, LocalMeta::with_ref(var));
+                    self.locals
+                        .insert(Cow::Borrowed(arg), LocalMeta::with_ref(var));
                 }
                 // Everything else still needs to become variables,
                 // to avoid having the same logic be executed
@@ -696,7 +699,7 @@ impl<'a> Generator<'a> {
                     values.write("(");
                     values.write(&self.visit_expr_root(expr)?);
                     values.write(")");
-                    self.locals.insert_with_default(arg);
+                    self.locals.insert_with_default(Cow::Borrowed(arg));
                 }
             }
         }
@@ -831,7 +834,7 @@ impl<'a> Generator<'a> {
         match var {
             Target::Name(name) => {
                 let name = normalize_identifier(name);
-                match self.locals.get(&name) {
+                match self.locals.get(&Cow::Borrowed(name)) {
                     // declares a new variable
                     None => Ok(false),
                     // an initialized variable gets shadowed
@@ -883,7 +886,7 @@ impl<'a> Generator<'a> {
         }
         if shadowed
             || !matches!(l.var, Target::Name(_))
-            || matches!(&l.var, Target::Name(name) if self.locals.get(name).is_none())
+            || matches!(&l.var, Target::Name(name) if self.locals.get(&Cow::Borrowed(name)).is_none())
         {
             buf.write("let ");
         }
@@ -1676,8 +1679,10 @@ impl<'a> Generator<'a> {
             Target::Name(name) => {
                 let name = normalize_identifier(name);
                 match initialized {
-                    true => self.locals.insert(name, LocalMeta::initialized()),
-                    false => self.locals.insert_with_default(name),
+                    true => self
+                        .locals
+                        .insert(Cow::Borrowed(name), LocalMeta::initialized()),
+                    false => self.locals.insert_with_default(Cow::Borrowed(name)),
                 }
                 buf.write(name);
             }
@@ -1940,10 +1945,10 @@ where
     }
 }
 
-impl MapChain<'_, &str, LocalMeta> {
+impl MapChain<'_, Cow<'_, str>, LocalMeta> {
     fn resolve(&self, name: &str) -> Option<String> {
         let name = normalize_identifier(name);
-        self.get(&name).map(|meta| match &meta.refs {
+        self.get(&Cow::Borrowed(name)).map(|meta| match &meta.refs {
             Some(expr) => expr.clone(),
             None => name.to_string(),
         })
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -8,7 +8,8 @@ use crate::input::{Source, TemplateInput};
 use crate::{CompileError, CRATE};
 
 use parser::node::{
-    Call, Comment, CondTest, If, Include, Let, Lit, Loop, Match, Target, Whitespace, Ws,
+    Call, Comment, CondTest, FilterBlock, If, Include, Let, Lit, Loop, Match, Target, Whitespace,
+    Ws,
 };
 use parser::{Expr, Node};
 use quote::quote;
@@ -310,6 +311,9 @@ impl<'a> Generator<'a> {
                 }
                 Node::Call(ref call) => {
                     size_hint += self.write_call(ctx, buf, call)?;
+                }
+                Node::FilterBlock(ref filter) => {
+                    size_hint += self.write_filter_block(ctx, buf, filter)?;
                 }
                 Node::Macro(ref m) => {
                     if level != AstLevel::Top {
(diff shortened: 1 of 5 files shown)


askama_derive/src/generator.rs
```diff
@@ -819,6 +829,62 @@ impl<'a> Generator<'a> {
...
+    fn write_filter_block(
+        &mut self,
+        ctx: &'a Context<'_>,
+        buf: &mut Buffer,
+        filter: &'a FilterBlock<'_>,
+    ) -> Result<usize, CompileError> {
+        self.flush_ws(filter.ws1);
+        let mut var_name = String::new();
+        for id in 0.. {
+            var_name = format!("__askama{id}");
```
djc, line 841: Let's call this `__filter_block{id}`?

askama_derive/src/generator.rs
```diff
@@ -1025,65 +1120,81 @@ impl<'a> Generator<'a> {
...
+        let mut buf_expr = Buffer::new(indent + 1);
+
         for s in mem::take(&mut self.buf_writable) {
             match s {
                 Writable::Lit(s) => {
                     buf_format.write(&s.replace('{', "{{").replace('}', "}}"));
                     size_hint += s.len();
                 }
                 Writable::Expr(s) => {
-                    use self::DisplayWrap::*;
```
djc, line 1043: Why?

askama_derive/src/generator.rs
```diff
@@ -1025,65 +1120,81 @@ impl<'a> Generator<'a> {
...
-                        Entry::Occupied(e) if is_cacheable(s) => *e.get(),
-                        e => {
-                            let id = self.named;
-                            self.named += 1;
-
-                            buf_expr.write(&format!("expr{id} = "));
-                            buf_expr.write("&");
-                            buf_expr.write(e.key());
-                            buf_expr.writeln(",")?;
+                    let id = self.generate_expr_id(
```
djc, line 1150: This should be a separate commit.

askama_derive/src/generator.rs
```diff
@@ -1008,10 +1074,39 @@ impl<'a> Generator<'a> {
         self.buf_writable.push(Writable::Expr(s));
     }
 
-    // Write expression buffer and empty
-    fn write_buf_writable(&mut self, buf: &mut Buffer) -> Result<usize, CompileError> {
+    fn generate_expr_id(
```
djc, line 1077: This should go below `buf_writable()`, not above.

askama_derive/src/generator.rs
```diff
@@ -1008,10 +1074,39 @@ impl<'a> Generator<'a> {
...
+        &mut self,
+        buf_expr: &mut Buffer,
+        is_cacheable: bool,
+        expression: String,
+        expr_cache: &mut HashMap<String, usize>,
+    ) -> Result<usize, CompileError> {
+        match expr_cache.entry(expression) {
```
djc, line 1084: If this is going to be in a separate function, let's do early return for the `Occupied` case so we can dedent the rest of it?

askama_derive/src/generator.rs
```diff
@@ -1008,10 +1074,39 @@ impl<'a> Generator<'a> {
...
+                if let Entry::Vacant(e) = e {
+                    e.insert(id);
+                }
+
+                Ok(id)
+            }
+        }
+    }
+
+    fn buf_writable(
```
djc, line 1104: This is not a great method name.

askama_derive/src/generator.rs
```diff
@@ -1025,65 +1120,81 @@ impl<'a> Generator<'a> {
...
-        buf.writeln("::std::write!(")?;
-        buf.indent();
-        buf.writeln("writer,")?;
-        buf.writeln(&format!("{:#?},", &buf_format.buf))?;
-        buf.writeln(buf_expr.buf.trim())?;
-        buf.dedent()?;
-        buf.writeln(")?;")?;
-        Ok(size_hint)
+    // Write expression buffer and empty
+    fn write_buf_writable(&mut self, buf: &mut Buffer) -> Result<usize, CompileError> {
```
djc, line 1180: Should be above `buf_writable()`, not below.

askama_derive/src/generator.rs
```diff
@@ -1915,18 +2036,19 @@ where
     }
 }
 
-impl MapChain<'_, &str, LocalMeta> {
-    fn resolve(&self, name: &str) -> Option<String> {
-        let name = normalize_identifier(name);
-        self.get(&name).map(|meta| match &meta.refs {
+impl MapChain<'_, Cow<'_, str>, LocalMeta> {
+    fn resolve(&self, name: Cow<'_, str>) -> Option<String> {
```
djc, line 2040: Could we avoid the caller having to wrap their `name` in `Cow`? Same for `resolver_or_self()`.

askama_parser/src/expr.rs
```diff
@@ -72,6 +72,8 @@ pub enum Expr<'a> {
     Call(Box<Expr<'a>>, Vec<Expr<'a>>),
     RustMacro(Vec<&'a str>, &'a str),
     Try(Box<Expr<'a>>),
+    /// This variant should never be used directly. It is created when generating filter blocks.
+    Generated(String),
```
djc, line 76: That seems ugly. Why is it necessary?

askama_parser/src/node.rs
```diff
@@ -558,6 +560,53 @@ impl<'a> Macro<'a> {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub struct FilterBlock<'a> {
+    pub ws1: Ws,
+    pub name: &'a str,
```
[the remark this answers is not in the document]
GuillaumeGomez: Let's go for `filter_name`. :)

askama_derive/src/generator.rs
GuillaumeGomez: Ah right, I needed to add the import to the other branch too and then my OCD triggered: it's horrible to have pattern matching with just idents and no path as then my brain doesn't know whether it means "bind everything" or "yep, look at the next one, it's actually binding each variant". I can revert that if you prefer.

askama_derive/src/generator.rs
GuillaumeGomez: I'm not sure to see what you have in mind. Can you show me please?

askama_derive/src/generator.rs
GuillaumeGomez: Agreed. I'm very bad at naming so any suggestion is very welcome. ^^'

askama_derive/src/generator.rs
GuillaumeGomez: Sure.

askama_parser/src/expr.rs
GuillaumeGomez: I need a variant with these two constraints:

 1. Doesn't take a reference (since it's a local variable).
 2. Doesn't make any transformation to it.

GuillaumeGomez:
Applied all suggestions (still need a new name for the `buf_writable` method).

askama_derive/src/generator.rs
djc: I can live with it in this case since it's just two arms.

askama_parser/src/expr.rs
djc: I mean, why do we need store any generated stuff back into an `Expr`?

askama_derive/src/generator.rs
```diff
@@ -1915,7 +2014,7 @@ where
     }
 }
 
-impl MapChain<'_, &str, LocalMeta> {
+impl MapChain<'_, &'_ str, LocalMeta> {
```
djc: This probably belongs in the other commit?

askama_derive/src/generator.rs
```diff
@@ -1025,65 +1116,81 @@ impl<'a> Generator<'a> {
...
-                                e.insert(id);
-                            }
-
-                            id
-                        }
+                    buf_format.write(&format!("{{expr{id}}}"));
+                    size_hint += 3;
+                }
+                Writable::Generated(s, wrapped) => {
+                    let expression = match wrapped {
```
djc, line 1157: Can we deduplicate more between the `Expr` and `Generated` arms?

askama_derive/src/generator.rs
```diff
@@ -1008,10 +1073,36 @@ impl<'a> Generator<'a> {
...
+                if let Entry::Vacant(e) = e {
+                    e.insert(id);
+                }
+
+                Ok(id)
+            }
+        }
+    }
+
+    fn buf_writable(&mut self, indent: u8) -> Result<WriteParts, CompileError> {
```
djc, line 1103: As for the name -- can you describe what this does now?

askama_derive/src/generator.rs
```diff
@@ -21,7 +23,7 @@ pub(crate) struct Generator<'a> {
     // The heritage contains references to blocks and their ancestry
     heritage: Option<&'a Heritage<'a>>,
     // Variables accessible directly from the current scope (not redirected to context)
-    locals: MapChain<'a, &'a str, LocalMeta>,
+    locals: MapChain<'a, Cow<'a, str>, LocalMeta>,
```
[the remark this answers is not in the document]
GuillaumeGomez: Done.

askama_derive/src/generator.rs
GuillaumeGomez: And of course it's obvious when seen. ^^'

askama_derive/src/generator.rs
GuillaumeGomez: It generates the literal string and the format arguments without adding the "wrapper" (ie `::std::write!(` or `format!(`).

askama_derive/src/generator.rs
GuillaumeGomez: Done.

askama_parser/src/expr.rs
GuillaumeGomez: Because we want to add the variable that was created to be used in the next `writer` use. I want this to be done in two steps: first we generate a `String` with the content of the filter and store it into a variable, then we use this variable alongside the rest of the template.

GuillaumeGomez:
Rebasing is now too difficult so I'll rework git history once PR is approved on your side. ^^'

askama_derive/src/generator.rs
djc: Maybe `prepare_format()` or some such?

djc:
> Rebasing is now too difficult so I'll rework git history once PR is approved on your side. ^^'

The actual purpose of clean history is making review easier, so that creates a circular dependency. 😛

GuillaumeGomez:
Oh boy. XD

Ok, let me rework the history now then. ^^'

GuillaumeGomez:
New clean history and all comments fixed/implemented.

GuillaumeGomez:
Ping. :)

GuillaumeGomez:
Any news about this? :)

GuillaumeGomez:
Ping @djc. :)

askama_derive/src/generator.rs
```diff
@@ -1025,65 +1096,93 @@ impl<'a> Generator<'a> {
...
 
-        buf.writeln("::std::write!(")?;
-        buf.indent();
-        buf.writeln("writer,")?;
-        buf.writeln(&format!("{:#?},", &buf_format.buf))?;
-        buf.writeln(buf_expr.buf.trim())?;
-        buf.dedent()?;
-        buf.writeln(")?;")?;
-        Ok(size_hint)
+    fn generate_expr_with_id(
```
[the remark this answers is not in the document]
GuillaumeGomez: What is a prefix commit? You mean moving this function into its own commit before the rest?

askama_derive/src/generator.rs
```diff
@@ -8,11 +8,14 @@ use crate::input::{Source, TemplateInput};
...
 
 use parser::node::{
-    Call, Comment, CondTest, If, Include, Let, Lit, Loop, Match, Target, Whitespace, Ws,
+    Call, Comment, CondTest, FilterBlock, If, Include, Let, Lit, Loop, Match, Target, Whitespace,
+    Ws,
 };
 use parser::{Expr, Node};
 use quote::quote;
 
+type WriteParts = (usize, Option<(Buffer, Option<Buffer>)>);
```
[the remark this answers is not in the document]
GuillaumeGomez: Let's go with a struct then.

askama_derive/src/generator.rs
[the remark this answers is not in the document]
djc: Yup!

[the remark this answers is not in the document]
GuillaumeGomez: Ok doing that.

askama_derive/src/generator.rs
```diff
@@ -1010,8 +1061,28 @@ impl<'a> Generator<'a> {
...
+                buf.writeln(&format!("{:#?},", &buf_format.buf))?;
+                buf.writeln(buf_expr.buf.trim())?;
+                buf.dedent()?;
+                buf.writeln(")?;")?;
+                Ok(size_hint)
+            }
+        }
+    }
+
+    fn prepare_format(&mut self, indent: u8) -> Result<WriteParts, CompileError> {
```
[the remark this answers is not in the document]
GuillaumeGomez: :+1:

askama_derive/src/generator.rs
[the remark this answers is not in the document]
GuillaumeGomez: That is actually much more tricky than I expected. I'd need to spend quite some time to make it work. Is it mandatory? ^^'

GuillaumeGomez:
Updated without moving the `generate_expr_with_id` for now.

askama_derive/src/generator.rs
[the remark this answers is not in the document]
djc: Why is it hard?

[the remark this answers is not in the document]
GuillaumeGomez: Mix of old and new code. I'm gonna give it another try, I'm sure I missed something obvious.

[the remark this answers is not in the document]
djc: I was just looking for something like this:

https://github.com/djc/askama/compare/wip-named-expression?expand=1

Seems straightforward to me?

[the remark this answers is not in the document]
GuillaumeGomez: I was making the opposite changes first from the current PR and then tried to cherry-pick from there. It ended in a disaster. It was as terrible as it sounds. ^^'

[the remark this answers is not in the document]
djc: So don't do that. 😄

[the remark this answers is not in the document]
GuillaumeGomez: Yes sir! o7

Joke aside, don't know why I did it this way... Oh well, end of the week. ^^'

GuillaumeGomez:
Move the change in a commit on its own (and as first commit).

askama_derive/src/generator.rs
```diff
@@ -8,11 +8,30 @@ use crate::input::{Source, TemplateInput};
...
 
 use parser::node::{
-    Call, Comment, CondTest, If, Include, Let, Lit, Loop, Match, Target, Whitespace, Ws,
+    Call, Comment, CondTest, FilterBlock, If, Include, Let, Lit, Loop, Match, Target, Whitespace,
+    Ws,
 };
 use parser::{Expr, Node};
 use quote::quote;
 
+struct WriteParts {
```
djc, line 18: Nit: as mentioned, should be defined near the bottom.

askama_derive/src/generator.rs
```diff
@@ -8,11 +8,30 @@ use crate::input::{Source, TemplateInput};
...
+    /// ```ignore
+    /// let var = buf_format;
+    /// ```
+    ///
+    /// Otherwise we need to format the "buf expr" using the "buf format":
+    ///
+    /// ```ignore
+    /// let var = format!(buf_format, buf_expr);
+    /// ```
+    buffers: Option<(Buffer, Option<Buffer>)>,
```
djc, line 33: Reading this docstring I definitely think we should just make the inner tuple a `struct` with named fields, too.

askama_derive/src/generator.rs
```diff
@@ -1025,14 +1123,19 @@ impl<'a> Generator<'a> {
                     buf_lit.write(s);
                 };
             }
-            buf.writeln(&format!("writer.write_str({:#?})?;", &buf_lit.buf))?;
-            return Ok(buf_lit.buf.len());
+            let size_hint = buf_lit.buf.len();
```
djc, line 1140: Nit: inline this binding (into the `WriteParts` initializer).

djc (commented):
I think we're on the home stretch here...

GuillaumeGomez:
Ah the push was finally taken into account by github. \o/

Now just have to wait for CI. ^^'

djc (approved):
Feel free to merge after the rebase -- and thanks for sticking with it.

GuillaumeGomez:
:tada:

Issue #810, opened 2023-04-03
Broken feature unification when multiple packages depend on `askama`


dpc:
I'm trying to make a workspace that in two different places of dependency tree uses `askama`

One imports only `askama`, the other one `askama_axum`. The one that only imports askama fails with:

```
error[E0433]: failed to resolve: could not find `askama_axum` in the list of imported crates                                                                    
   --> /home/dpc/.cargo/git/checkouts/uniffi-rs-5e0d6f877b5a61c0/b24850f/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs:120:10
    |                                                                                                                                                           
120 | #[derive(Template)]                                                                                                                                       
    |          ^^^^^^^^ could not find `askama_axum` in the list of imported crates
    |                                                                                                                                                           
    = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)                                          
```

Seems to me that the other crate enabled `with-axum` feature, and `askama_derive` is being picked up by the `uniffi-rs` so it now generates code that wants to access `askama_axum`, which is not a dependency of it.

I've read through https://github.com/djc/askama/issues/588#issuecomment-1485986574 which seems somewhat related.

Unless I'm missing something somewhere, this might be a hard circle to square. Downstream crates with different extenson-sets, can't share the `askama_derive`, it seems.

djc:
Yeah, not sure if there's a solution for this. The easiest solution is probably to have both of your crates enable with `with-axum` feature.

djc:
Having to decorate each type with all the integrations needed seems like a pain in the ass. But, we could set this up in the configuration? So you declare in the `askama.toml` for your crate which integrations you want.

dpc:
If that's possible, then it does seem like an improvement.

5iddy:
I am experiencing this issue as well but with `with-actix-web` feature.
Here is the errror:
```
error[E0433]: failed to resolve: could not find `askama_actix` in the list of imported crates
 --> src/templates/index.rs:3:17
  |
3 | #[derive(Debug, Template)]
  |                 ^^^^^^^^ could not find `askama_actix` in the list of imported crates
  |
  = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0433`.
error: could not compile `xprs` due to previous error
```

djc:
[...]

PR #965, opened 2024-02-13
Auto-add integration prefix in derive

### LINKED ISSUE #961: Document how integration should be imported
The integration section of the "book" looks like this:

> ## {{TARGET}} integration
>
> Enabling the `with-{{target}}` feature appends an implementation of {{target}}'s {{corresponding_trait}} trait for each template type. This makes it easy to trivially return a value of that type in a {{target}} handler.

(repeat for each target)

However, a quick search shows that [you](https://github.com/djc/askama/issues/569#issuecomment-986238576) [need](https://github.com/djc/askama/issues/588#issuecomment-1002212182) [askama_{{target}}](https://github.com/djc/askama/issues/810#issuecomment-1497074654), not askama itself with `with-{{target}}` enabled (or [askama_{{target}} alone](https://github.com/djc/askama/issues/569#issuecomment-986553248)).

On the other hand, the getting started section says nothing more than adding askama itself.

If askama_{{target}} is the canonical way to use integration, maybe document that? I'm happy to PR with some advice on writing alignment.

### LINKED ISSUE #962: Remove the need to import askama itself when using integration crates
While wording on https://github.com/djc/askama/issues/961, I wandered off to find the cause for the need as the title goes. A rough scan reveals that askama_derive generates some uses of `::askama::`, for example, this:

https://github.com/djc/askama/blob/c2c9bd81708a0201aef49b83df89446e12a41817/askama_derive/src/generator.rs#L93-L96

… which leads to the need, as sub-dependencies can't be used directly. If they are removed, probably we can depend solely on integration crates?

A possible way to do this might be something like

```rust
#[inline]
fn _askama(tail: &str) -> String {
	format!("::{}::askama{tail}",
		#[cfg(feature = "with-actix-web")]
		"askama_actix",
		#[cfg(feature = "with-axum")]
		"askama_axum",
		// ...
	)
}

buf.write("fn render_into(&self, writer: &mut (impl ::std::fmt::Write + ?Sized)) -> ")?;
buf.writeln(&_askama("::Result<()> {"))?;
```

… which isn't pretty, but works.

nc7s:
… to avoid need for importing askama itself. Closes #962.

Performance is negligible in a `cargo bench` comparison (criterion reports either in error margin or no performance change detected). If you think it's still too much, the branch can be easily moved to a separate `fn`.

Update: as suggested, also closes #961.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -5,7 +5,7 @@ use std::{cmp, hash, mem, str};
 use crate::config::WhitespaceHandling;
 use crate::heritage::{Context, Heritage};
 use crate::input::{Source, TemplateInput};
-use crate::CompileError;
+use crate::{CompileError, CRATE};
 
 use parser::node::{
     Call, Comment, CondTest, If, Include, Let, Lit, Loop, Match, Target, Whitespace, Ws,
@@ -90,11 +90,10 @@ impl<'a> Generator<'a> {
         ctx: &'a Context<'_>,
         buf: &mut Buffer,
     ) -> Result<(), CompileError> {
-        self.write_header(buf, "::askama::Template", None)?;
-        buf.writeln(
-            "fn render_into(&self, writer: &mut (impl ::std::fmt::Write + ?Sized)) -> \
-             ::askama::Result<()> {",
-        )?;
+        self.write_header(buf, &format!("{CRATE}::Template"), None)?;
+        buf.write("fn render_into(&self, writer: &mut (impl ::std::fmt::Write + ?Sized)) -> ");
+        buf.write(CRATE);
+        buf.writeln("::Result<()> {")?;
 
         // Make sure the compiler understands that the generated code depends on the template files.
         for path in self.contexts.keys() {
@@ -121,7 +120,8 @@ impl<'a> Generator<'a> {
         }?;
 
         self.flush_ws(Ws(None, None));
-        buf.writeln("::askama::Result::Ok(())")?;
+        buf.write(CRATE);
+        buf.writeln("::Result::Ok(())")?;
         buf.writeln("}")?;
 
         buf.writeln("const EXTENSION: ::std::option::Option<&'static ::std::primitive::str> = ")?;
@@ -145,7 +145,8 @@ impl<'a> Generator<'a> {
         self.write_header(buf, "::std::fmt::Display", None)?;
         buf.writeln("#[inline]")?;
         buf.writeln("fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {")?;
-        buf.writeln("::askama::Template::render_into(self, f).map_err(|_| ::std::fmt::Error {})")?;
+        buf.write(CRATE);
+        buf.writeln("::Template::render_into(self, f).map_err(|_| ::std::fmt::Error {})")?;
         buf.writeln("}")?;
         buf.writeln("}")
     }
@@ -226,13 +227,13 @@ impl<'a> Generator<'a> {
                 #where_clause
             )
         ))?;
-        buf.writeln("type Error = ::askama::Error;")?;
+        buf.writeln("type Error = ::askama_hyper::Error;")?;
         buf.writeln("#[inline]")?;
         buf.writeln(&format!(
             "{} {{",
             quote!(fn try_from(value: &#ident #orig_ty_generics) -> Result<Self, Self::Error>)
         ))?;
-        buf.writeln("::askama::Template::render(value).map(Into::into)")?;
+        buf.writeln("::askama_hyper::Template::render(value).map(Into::into)")?;
         buf.writeln("}")?;
         buf.writeln("}")
     }
@@ -307,7 +308,7 @@ impl<'a> Generator<'a> {
         buf.writeln("#[inline]")?;
         buf.writeln(
             "fn respond_to(self, _: &'askama1 ::askama_rocket::Request) \
-             -> ::askama_rocket::Result<'askama2> {",
+             -> ::askama_rocket::RocketResult<'askama2> {",
         )?;
         buf.writeln("::askama_rocket::respond(&self)")?;
 
@@ -324,9 +325,10 @@ impl<'a> Generator<'a> {
             None,
         )?;
         buf.writeln(
-            "type Error = ::askama_tide::askama::Error;\n\
+            "type Error = ::askama_tide::Error;\n\
             #[inline]\n\
-            fn try_into(self) -> ::askama_tide::askama::Result<::askama_tide::tide::Body> {",
+            fn try_into(self) -> \
+			::askama_tide::Result<::askama_tide::tide::Body> {",
         )?;
         buf.writeln("::askama_tide::try_into_body(&self)")?;
         buf.writeln("}")?;
@@ -638,7 +640,9 @@ impl<'a> Generator<'a> {
         self.locals.push();
         buf.write("for (");
         self.visit_target(buf, true, true, &loop_block.var);
-        buf.writeln(", _loop_item) in ::askama::helpers::TemplateLoop::new(_iter) {")?;
+        buf.write(", _loop_item) in ");
+        buf.write(CRATE);
+        buf.writeln("::helpers::TemplateLoop::new(_iter) {")?;
 
         if has_else_nodes {
             buf.writeln("_did_loop = true;")?;
@@ -1046,7 +1050,7 @@ impl<'a> Generator<'a> {
                     let expression = match wrapped {
                         Wrapped => expr_buf.buf,
                         Unwrapped => format!(
-                            "::askama::MarkupDisplay::new_unsafe(&({}), {})",
+                            "{CRATE}::MarkupDisplay::new_unsafe(&({}), {})",
                             expr_buf.buf, self.input.escaper
                         ),
                     };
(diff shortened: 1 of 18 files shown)


GuillaumeGomez:
Nice! Please also update the tests to remove the unneeded imports too.

nc7s:
Ouch, didn't notice the repetition when grepping over integration crates. This would be fixed along with the tests change.

GuillaumeGomez:
Oh and also the askama book too please.

nc7s:
Sure, I intended to update the book as part of #961 though.

GuillaumeGomez:
Well, you're changing how it's done here. So makes more sense to do it here so we're sure it won't be forgotten. You can mark the PR as fix for both issues (or at least mention the related issues if other things need to be done before they're marked as fixed).

nc7s:
> Well, you're changing how it's done here. So makes more sense to do it here so we're sure it won't be forgotten.

Makes sense, will do.

> Please also update the tests to remove the unneeded imports too.

Can you be more clear on "unneeded imports", i.e. do you mean `use askama::Template;` or `pub use askama::{self, *};`, or something else? The `self` part is needed for the derive, I don't know if `*` is still needed.

nc7s:
Yeah… that's kinda the whole point of this PR. `askama` being an indirect dependency isn't in the global scope, that is. `*` is needed for `askama::Error` in askama_actix, not sure about others.

book/src/getting_started.md
```diff
@@ -35,3 +35,27 @@ fn main() {
...
+## Using integrations
+
+To use one of the [integrations](./integrations.md), with axum as an example:
+
+First, add this to your `Cargo.toml` instead:
+
+```toml
+# in section [dependencies]
+askama_axum = "0.4.0"
+
```
GuillaumeGomez, line 48: Please remove this empty line.

book/src/getting_started.md
```diff
@@ -35,3 +35,27 @@ fn main() {
...
+```
+
+Then, import from askama_axum instead of askama:
+
+```rust
+//use askama::Template;
```
GuillaumeGomez, line 54: Forgot this maybe?

askama_hyper/tests/basic.rs
```diff
@@ -1,6 +1,6 @@
 use std::convert::Infallible;
 
-use askama::Template;
+use askama_hyper::Template;
```
GuillaumeGomez, line 3: Why changing this (same for all others)?

nc7s: This seems to be a misunderstanding of "unused imports", should I revert it?

book/src/getting_started.md
nc7s: This is a visualization for "instead" ;) if you don't feel like it I'll remove it.

book/src/getting_started.md
```diff
@@ -4,7 +4,7 @@ First, add the following to your crate's `Cargo.toml`:
 
 ```toml
 # in section [dependencies]
-askama = "0.11.2"
+askama = "0.12.1"
```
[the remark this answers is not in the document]
nc7s: This one was inherited from the original. The other one followed suit. Sure can be removed if they shouldn't be there.

book/src/getting_started.md
GuillaumeGomez: it's a bit strange. I'll let another reviewer give the last word here.

book/src/getting_started.md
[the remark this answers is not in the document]
GuillaumeGomez: Yes please! :)

nc7s:
Friendly ping that all reviews opinions so far are implemented ;)

askama_derive/src/generator.rs
```diff
@@ -13,6 +13,27 @@ use parser::node::{
...
+} else if cfg!(feature = "with-mendes") {
+    "::askama_mendes"
+} else if cfg!(feature = "with-rocket") {
+    "::askama_rocket"
+} else if cfg!(feature = "with-tide") {
+    "::askama_tide"
+} else if cfg!(feature = "with-warp") {
+    "::askama_warp"
+} else {
+    ""
```
djc, line 34: Why shouldn't this be `::askama`? If this is acting as a sentinel value, I'd prefer to make this an `Option` instead.

askama_derive/src/generator.rs
```diff
@@ -1815,7 +1836,21 @@ impl Buffer {
             }
             self.start = false;
         }
-        self.buf.push_str(s);
+
+        if _PREFIX.is_empty() || !s.contains(_ASKAMA_) {
+            self.buf.push_str(s);
```
djc, line 1841: Let's do an early return here, so we can reduce rightward drift in the `else` case?

djc (approved):

askama_derive/src/generator.rs
nc7s: This (`const _PREFIX: &str`) is the integration prefix. It's "concatenated" with `const _ASKAMA_: &str` i.e. `"::askama::"`, so it needs to remain empty when no integration is present, otherwise the result would be `::askama::askama::`. It's not a sentinel value AFAICT.

djc: Ahh, yeah, that's fair.

askama_derive/src/generator.rs
```diff
@@ -1839,16 +1839,16 @@ impl Buffer {
...
-                self.buf.push_str(part);
-                if !part.ends_with(_PREFIX) && it.peek().is_some() {
-                    self.buf.push_str(_PREFIX);
-                }
-                if it.peek().is_some() {
-                    self.buf.push_str(_ASKAMA_);
-                }
+            return;
+        }
+        let mut it = s.split(_ASKAMA_).peekable();
```
djc, line 1844: Please squash this into your first commit?

Kijewski:
I did not read the whole discussion properly, so maybe my question is stupid: Why are you replacing `askama` → `askama_integration` in the generated code instead of directly replacing `::askama` → `{PREFIX}` the `format!()` statements? The compiler is actually smart enough to concatenate constant strings at compile time.

nc7s:
> … instead of directly replacing ::askama → {PREFIX} the format!() statements? The compiler is actually smart enough…

Good point! However, I worry that would make future changes more fragile, if not annoying: you then _have to remember_ to add `{_PREFIX}` to every `format!` invocation that writes `::askama::`. On the other hand, this codegen action _is_ run at compile time, exactly once every compilation, like the compiler doing smart concatenation.

djc:
> > … instead of directly replacing ::askama → {PREFIX} the format!() statements? The compiler is actually smart enough…
> 
> Good point! However, I worry that would make future changes more fragile, if not annoying: you then _have to remember_ to add `{_PREFIX}` before `::askama::` (or replace it with a `cfg(feature)` gated const) to every `format!` invocation that writes it, while plain `.write{,ln}(&'static str)` invocations don't have that benefit. On the other hand, this codegen action _is_ run at compile time, exactly once every compilation, like the compiler doing smart concatenation.

Hmm, I think this is probably not actually a big deal? Like, it seems likely that future contributors will look at the surrounding code and copy what it's doing -- and otherwise we'll get an issue from an integration user soon enough. So I think on balance, I prefer the simplicity of requiring `write{,ln}()` callers to insert `PREFIX` explicitly.

nc7s:
Here is a preview commit then. It will be squashed after review.

askama_actix/src/lib.rs
```diff
@@ -9,7 +9,7 @@ use actix_web::body::BoxBody;
 use actix_web::http::header::HeaderValue;
 use actix_web::http::StatusCode;
 use actix_web::{HttpResponse, HttpResponseBuilder, ResponseError};
-pub use askama::*;
+pub use askama::{self, *};
```
djc, line 12: Why do we need both? Can we just keep it as `*` and adjust the macro paths accordingly?

(Also I'd like this third commit to be squashed into the first one.)

nc7s: Well, that's overlooked. Fixed in the following (preview) commit (they will be squashed once review is done).

askama_derive/src/generator.rs
```diff
@@ -226,13 +247,13 @@ impl<'a> Generator<'a> {
                 #where_clause
             )
         ))?;
-        buf.writeln("type Error = ::askama::Error;")?;
+        buf.writeln("type Error = ::askama_hyper::askama::Error;")?;
```
Kijewski, line 250: I think all instances of `::askama_something::askama::` have to be replaced with `::askama_something::` now that we don't export a symbol `askama` in the integration crates.

nc7s:
Oh I almost forgot this: there are some `::askama::{Html,Text,Js}` in askama_derive/src/config.rs. The original find and replace handles that, now this is more subtle.

djc:
> Oh I almost forgot this: there are some `::askama::{Html,Text,Js}` in askama_derive/src/config.rs. The original find and replace handles that, now this is more subtle.

Can we move `CRATE` to `lib.rs` and use `concat!("::", CRATE, "::Html")` for those?

nc7s:
> Can we move `CRATE` to `lib.rs` and use `concat!("::", CRATE, "::Html")` for those?

`concat!` I'm afraid not. Already tried for generator.rs - no good. Do you mind `format!`-ing there, as @Kijewski suggested? Or do it manually?

djc:
If `format!()` works in `const` context that sounds like exactly what I'm after.

Kijewski:
> If format!() works in const context that sounds like exactly what I'm after.

That's an unstable feature for now.

But all members of `DEFAULT_ESCAPERS` are prefixed by `::askama::`. I guess we can simply strip this prefix and only add it in the line

<https://github.com/djc/askama/blob/4dab5f91ba15e7c238ddf6223ec7b45eef32cab4/askama_derive/src/config.rs#L96>

We turn the &str into a String, anyway, so there should be no harm in using format!().

TheBlueFireFox:
1) Thanks for the great create - I love using it
2) Will this PR be also fixing #810? As askama itself will not need the flag like "with_axum", or did I misunderstand something here?

djc (approved):

Kijewski (approved):
Very nice! I think that's big usability improvement when you're new to askama or rust in general. :+1:

djc:
@nc7s thanks for working through this -- I cleaned up some minor things to prepare this for merging to avoid further round trips.

PR #967, opened 2024-02-23
Explain how to go around rust macro limitations

### LINKED ISSUE #966: Rust macro has no access to parameters passed into `{% macro f(param) %}`
Not sure if this is a question, feature request or a bug.

From the documentation it's not clear to me in what scope rust macros are evaluated.
For instance, I have this Askama macro, which calls a Rust macro internally:

```jinja
{% macro stop(entity) %}
    {{ create_action!(entity, LightOn, {brightness:Some(0.1), color_temperature: None})|json }}
{% endmacro %}
```

In some situations this works perfectly, while in other situations this does not and results in a "cannot find value `entity` in this scope". (`entity` being the argument passed to the Askama macro.)

It does work for one situation where this macro is called in another template within a loop like this: `{% for entity in entities.values() %}`. In any case, I've the impression that it doesn't pick up the argument passed to the Askama macro.

Is that understanding correct? Is there something we can improve here to make this possible?

GuillaumeGomez:
Fixes #966.

Does it look good to you @jonathanslenders?

### DIFF
--- a/book/src/template_syntax.md
+++ b/book/src/template_syntax.md
@@ -642,3 +642,38 @@ In here it's invalid because `arg2` is the second argument and would be used by
 {# Equivalent of: #}
 {% call heading("something", "title", "b", arg4="ah") %}
 ```
+
+## Calling Rust macros
+
+It is possible to call rust macros directly in your templates:
+
+```jinja
+{% let s = format!("{}", 12) %}
+```
+
+One important thing to note is that contrary to the rest of the expressions,
+Askama cannot know if a token given to a macro is a variable or something
+else, so it will always default to generate it "as is". So if you have:
+
+```rust
+macro_rules! test_macro{
+    ($entity:expr) => {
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Perfect! Thanks for the quick response!

Kijewski (approved):
:+1:

book/src/template_syntax.md
```diff
@@ -642,3 +642,38 @@ In here it's invalid because `arg2` is the second argument and would be used by
...
+## Calling Rust macros
+
+It is possible to call rust macros directly in your templates:
+
+```jinja
+{% let s = format!("{}", 12) %}
+```
+
+One important thing to note is that contrary to the rest of the expressions,
+askama cannot know if a token given to a macro is a variable or something
```
djc, line 655: Nit: I think we capitalize Askama elsewhere.

djc (approved):
Thanks!

PR #972, opened 2024-03-05
Integration spring cleaning

djc:
Recent crates.io downloads:

* askama_axum: 30k
* askama_actix: 4k
* askama_rocket: 1130
* askama_tide: 978
* askama_warp: 882
* askama_gotham: 366
* askama_hyper: 143
* askama_mendes: 0 (never published)

The overhead of maintaining these integrations isn't very large, but it's also not zero. I think it makes sense to clean some of these out.

PR #975, opened 2024-03-06
Remove tide integration

djc:
See discussion in #972. This gets CI passing again because the unmaintained crates are gone from the dependency graph.

If anyone thinks we should keep/resurrect it, speak up!


book/src/integrations.md
```diff
@@ -50,22 +50,3 @@ Enabling the `with-warp` feature appends an implementation of Warp's `Reply`
...
-[`askama_tide::Template`][askama_tide].
-
-Enabling the `with-tide` feature appends `Into<tide::Response>` and
-`TryInto<tide::Body>` implementations for each template type. This
-provides the ability for tide apps to build a response directly from
-a template, or to append a templated body to an existing
-`Response`. See [the example](https://github.com/djc/askama/blob/main/askama_tide/tests/tide.rs)
-from the Askama test suite for more on how to integrate.
-
-[askama_rocket]: https://docs.rs/askama_rocket
```
GuillaumeGomez, line 66: Most of these links are still used, they should not be removed.

djc: Oops!

PR #977, opened 2024-03-06
Unify integration API 

### LINKED ISSUE #974: Unify integration API
There is

* [`fn askama_warp::reply(&impl Template) -> Response`](https://github.com/djc/askama/blob/178a6895d47a397e8d6bffcbc34c9782b08b0b8c/askama_warp/src/lib.rs)
* [`fn askama_tide::try_into_body(&impl Template) -> Result<Body>`](https://github.com/djc/askama/blob/178a6895d47a397e8d6bffcbc34c9782b08b0b8c/askama_tide/src/lib.rs)
* [`fn askama_tide::into_response(&impl Template) -> Response`](https://github.com/djc/askama/blob/178a6895d47a397e8d6bffcbc34c9782b08b0b8c/askama_tide/src/lib.rs)
* [`fn askama_rocket::respond(&impl Template) -> Result<_>`](https://github.com/djc/askama/blob/178a6895d47a397e8d6bffcbc34c9782b08b0b8c/askama_rocket/src/lib.rs)
* [`fn askama_axum::into_response(&impl Template) -> Response`](https://github.com/djc/askama/blob/178a6895d47a397e8d6bffcbc34c9782b08b0b8c/askama_axum/src/lib.rs)
* [`<T as TemplateToResponse>::to_response(&self) -> Response`](https://github.com/djc/askama/blob/178a6895d47a397e8d6bffcbc34c9782b08b0b8c/askama_actix/src/lib.rs)

Shouldn't we use the moment, if we upload new versions of the integration crates anyway, to unify the APIs amongst the crates?

I like the API of `askama_tide` the best: One function `try_into_response(&impl Template)` to generate a `Result<_>`, and one function `into_response(&impl Template)` to generate a response that might be an error page.

Kijewski:
Resolves #974.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ in a for-profit context, please consider supporting my open source work on
 * Construct templates using a familiar, easy-to-use syntax
 * Benefit from the safety provided by Rust's type system
 * Template code is compiled into your crate for [optimal performance][benchmarks]
-* Optional built-in support for Actix, Axum, Gotham, Mendes, Rocket, tide, and warp web frameworks
+* Optional built-in support for Actix, Axum, Rocket, and warp web frameworks
 * Debugging features to assist you in template development
 * Templates must be valid UTF-8 and produce UTF-8 when rendered
 * IDE support available in [JetBrains products](https://plugins.jetbrains.com/plugin/16591-askama-template-support)
--- a/askama_actix/src/lib.rs
+++ b/askama_actix/src/lib.rs
@@ -4,12 +4,29 @@
 
 use std::fmt;
 
+#[doc(no_inline)]
 pub use actix_web;
 use actix_web::body::BoxBody;
 use actix_web::http::header::HeaderValue;
 use actix_web::http::StatusCode;
 use actix_web::{HttpResponse, HttpResponseBuilder, ResponseError};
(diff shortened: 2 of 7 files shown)


Kijewski:
Now every integration crate exports `askama::*`, the framework crate, `try_into_response()` and `into_response()`.

Kijewski:
> I'd prefer to standardize on using `&impl askama::Template + ?Sized`.

The `try_into_response()` functions all depend on `T` for `T::MIME_TYPE`. I wanted both functions to look the same, both using explicit generics, because the implicit variant cannot be used for `try_into_response()`. Should I change `into_response()` anyway?

djc:
> > I'd prefer to standardize on using `&impl askama::Template + ?Sized`.
> 
> The `try_into_response()` functions all depend on `T` for `T::MIME_TYPE`. I wanted both functions to look the same, both using explicit generics, because the implicit variant cannot be used for `try_into_response()`. Should I change `into_response()` anyway?

Ah, that's fair. Let's leave it then.

Would you mind squashing the reordering into the originating commits?

Kijewski:
Squashed.

djc (approved):

PR #984, opened 2024-03-07
Add support for filter chaining in filter blocks

GuillaumeGomez:
Follow-up of https://github.com/djc/askama/pull/959.

It allows to chain filters on filter blocks, making the feature much more useful. :)

With this, the feature will be complete.

### DIFF
--- a/book/src/template_syntax.md
+++ b/book/src/template_syntax.md
@@ -98,6 +98,17 @@ blocks**:
 
 The `lower` filter will be applied on the whole content.
 
+Just like filters, you can combine them:
+
+```text
+{% filter lower|capitalize %}
+    {{ t }} / HELLO / {{ u }}
+{% endfilter %}
+```
+
+In this case, `lower` will be called and then `capitalize` will be
+called on what `lower` returned.
+
 ## Whitespace control
 
 Askama considers all tabs, spaces, newlines and carriage returns to be


testing/tests/filter_block.rs
```diff
@@ -128,3 +128,22 @@ fn filter_block_not_html_escape() {
...
+{%- endfilter %}"#,
+    ext = "txt"
+)]
+struct F {
+    u: &'static str,
+}
+
+#[test]
+fn filter_block_chaining() {
+    let template = F { u: "yOU" };
```
GuillaumeGomez, line 147: A friend by reading this told me it could be badly interpreted. Didn't pay attention, sorry about that. Changing right away.

Kijewski: I work at the FU Berlin (Freie Universität, Free University). I'm desensitized :D

askama_parser/src/node.rs
```diff
@@ -563,25 +563,44 @@ impl<'a> Macro<'a> {
...
+            Ok((i, (fname, args)))
+        }
+
         let mut start = tuple((
             opt(Whitespace::parse),
             ws(keyword("filter")),
             cut(tuple((
                 ws(identifier),
                 opt(|i| Expr::arguments(i, s.level.get(), false)),
+                many0(|i| filter(i, s.level.get())),
```
GuillaumeGomez, line 589: This one will need to be updated (related to https://github.com/djc/askama/pull/979).

testing/tests/filter_block.rs
GuillaumeGomez: Didn't expect this comeback :laughing:

askama_derive/src/generator.rs
```diff
@@ -15,6 +15,14 @@ use parser::node::{
 use parser::{Expr, Node};
 use quote::quote;
 
+fn insert_variable_at_the_end(args: &mut Vec<Expr<'_>>, var_name: String) {
```
djc, line 18: You should know better about where to put this by now. 😛

askama_parser/src/node.rs
```diff
@@ -563,25 +563,44 @@ impl<'a> Macro<'a> {
...
     fn parse(i: &'a str, s: &State<'_>) -> ParseResult<'a, Self> {
+        #[allow(clippy::type_complexity)]
+        fn filter(i: &str, level: crate::Level) -> ParseResult<'_, (&str, Option<Vec<Expr<'_>>>)> {
+            let (i, (_, fname, args)) = tuple((
+                char('|'),
+                ws(identifier),
+                opt(|i| Expr::arguments(i, level, false)),
+            ))(i)?;
+            Ok((i, (fname, args)))
+        }
```
djc, line 581: Can we reuse this across `Expression::filtered` and `FilterBlock`?

askama_derive/src/generator.rs
```diff
@@ -15,6 +15,14 @@ use parser::node::{
...
+    if let Some(Expr::Filter(_, ref mut args)) = args.first_mut() {
```
djc, line 19: Do we need `ref mut` here? Why don't match ergonomics thingies do their thing?

askama_derive/src/generator.rs
```diff
@@ -15,6 +15,14 @@ use parser::node::{
...
+        insert_variable_at_the_end(args, var_name);
+    } else {
+        args.insert(0, Expr::Generated(var_name.clone()));
```
djc, line 22: Why do we need to `clone()` here?

askama_derive/src/generator.rs
GuillaumeGomez: Oh no. :o I was trying to remember if it was "a la C" (ie, defined before used) or the opposite, picked the wrong one obviously. ^^'

askama_parser/src/node.rs
GuillaumeGomez: I was wondering about that if I should wait for your other PR to be merged first. Merging both functions.

askama_derive/src/generator.rs
GuillaumeGomez: Ah, the old debate: better for reading or writing? I tend to prefer the reading but yes, you're absolutely right, ergonomics handle this case. Updating it.

askama_derive/src/generator.rs
GuillaumeGomez: We don't, good catch!

GuillaumeGomez:
Applied suggestions.

askama_derive/src/generator.rs
djc: It says `struct Generator` like 5 lines below, which should be a decent indication?

GuillaumeGomez: Sorry, I don't understand what you mean.

askama_derive/src/generator.rs
djc: (Personally I don't find the explicit `ref mut` better for reading -- it's noise that I don't need when we're matching against `first_mut()`.)

askama_parser/src/lib.rs
```diff
@@ -371,6 +371,16 @@ fn path_or_identifier(i: &str) -> ParseResult<'_, PathOrIdentifier<'_>> {
     }
 }
 
+#[allow(clippy::type_complexity)]
```
djc, line 374: Let's move this below `Level`.

askama_parser/src/node.rs
```diff
@@ -563,8 +563,7 @@ impl<'a> Macro<'a> {
 #[derive(Debug, PartialEq)]
 pub struct FilterBlock<'a> {
     pub ws1: Ws,
-    pub filter_name: &'a str,
-    pub args: Vec<Expr<'a>>,
+    pub filters: Expr<'a>,
```
djc, line 566: I propose you extract a `struct Filter<'a> { name: &'a str, args: Vec<Expr<'a>> }` from `Expr::Filter` (in a separate commit), and then give `filters` the type `Vec<Filter<'a>>` in this commit.

GuillaumeGomez: :+1:

djc: (If we end up needing to insert in front, maybe `args` should be a `VecDeque`?)

GuillaumeGomez: Considering we very likely always have less than 10 arguments and most of the time less than 2, I think the advantage is quite small. Could be interesting to perf-check.

askama_parser/src/lib.rs
GuillaumeGomez: Ok!

GuillaumeGomez:
Created a new `Filter` type in its own commit. It required to update a lot of tests, forcing me to remove the `use Expr::*` because of name conflict, it's mostly because of cases like this that I'm not a big fan matching on just variant names. :wink:

askama_parser/src/tests.rs
```diff
@@ -25,33 +25,47 @@ fn test_invalid_block() {
...
 #[test]
 fn test_parse_filter() {
-    use Expr::*;
     let syntax = Syntax::default();
     assert_eq!(
         Ast::from_str("{{ strvar|e }}", &syntax).unwrap().nodes,
-        vec![Node::Expr(Ws(None, None), Filter("e", vec![Var("strvar")]),)],
+        vec![Node::Expr(
+            Ws(None, None),
+            Expr::Filter(Filter {
```
djc, line 33: Looks like these changes belong in the previous commit.

GuillaumeGomez: Arf indeed! `git reflog` to the rescue!

GuillaumeGomez: Rescue was a success :laughing: .

GuillaumeGomez:
Fixed the commit mess. :)

askama_parser/src/expr.rs
```diff
@@ -51,6 +51,12 @@ macro_rules! expr_prec_layer {
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
```
djc, line 55: .. this should be somewhere below `Expr`, probably just above `Suffix`.

GuillaumeGomez: You should add a lint about this, that would make your reviews much simpler. I keep forgetting about how you prefer the definitions to come after their use. ^^'

GuillaumeGomez: (yep, C/C++ syndrom is hitting hard haha)

djc: I don't think linting for this exists?

(Most contributors pick up on it more quickly than you have so far... 😛 )

GuillaumeGomez: I'm horribly slow on style things. I'm too used to have them enforced by lints. I've been spoiled. :crying_cat_face: 

And no, you'd need to write either a check or write your own linter (I wrote a blog post on how you can do it [here](https://blog.guillaume-gomez.fr/articles/2024-01-18+Writing+your+own+Rust+linter) if you're interested :wink: ) for this to be checked in the CI. Shouldn't be too hard I think.

askama_derive/src/generator.rs
```diff
@@ -2033,6 +2038,14 @@ fn median(sizes: &mut [usize]) -> usize {
     }
 }
 
+fn insert_variable_at_the_end(args: &mut Vec<Expr<'_>>, var_name: String) {
```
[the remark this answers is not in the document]
GuillaumeGomez: It's a bit counter-intuitive because the first filter called is actually the one at the "bottom". I'll rename and add documentation to clarify all this.

askama_parser/src/expr.rs
GuillaumeGomez: Oh boy. More work on top of my clippy/rustdoc/rustc_codegen_gcc/browser-ui-test/sysinfo/rustc pattern matching works? :o

Well, if you're not in a hurry, why not. ^^'

GuillaumeGomez:
Renamed and added documentation.

djc (approved):

PR #986, opened 2024-03-11
Improve error message by providing the file where the error occurred if it is not an inlined source

GuillaumeGomez:
When you have templates included in templates, the errors become quite difficult to understand as we don't know which file is triggering it. With this change, it adds the path of the file where the error is located:

```
error: problems parsing template source in `templates/core/about/builds.html` at row 34, column 32 near:
       "=\"rust\") %}{% filter dedent(levels=3) -%"...
   --> src/web/sitemap.rs:109:10
    |
109 | #[derive(Template)]
    |          ^^^^^^^^
    |
```

Making my life and very likely everyone else's better. :)

Sadly, I can't enforce it with `trybuild` as it only allows to have inline tests (because the folder where the ui tests are run is different and I couldn't find a way to do it...).


askama_parser/src/lib.rs
```diff
@@ -487,3 +505,43 @@ fn filter(i: &str, level: Level) -> ParseResult<'_, (&str, Option<Vec<Expr<'_>>>
...
+/// The goal of this function is to reduce the path length based on the current directory
+/// the programming is running into. For example:
+///
+/// ```text
+/// current dir: /a/b/c
+/// path:        /a/b/c/d/e.txt
+/// ```
+///
+/// `strip_common` will return `d/e.txt`.
+fn strip_common(path: &Path) -> String {
```
djc, line 518: I'd like to see a bunch of unit tests for this in a `tests` module at the bottom of the file. As such, we probably want to have the caller call in a `cwd`.

askama_derive/src/input.rs
```diff
@@ -356,7 +356,8 @@ fn extension(path: &Path) -> Option<&str> {
 #[derive(Debug)]
 pub(crate) enum Source {
     Path(String),
-    Source(String),
+    /// Contains the text of the template and its path (if it's not an inline template).
+    Source(String, Option<PathBuf>),
```
djc, line 360: This is confusing. If my memory serves correctly, `Source::Source` exactly describes "inline templates", so the `Option<PathBuf>` should actually be on the `Path` variant? Did you test this end to end?

askama_parser/src/lib.rs
GuillaumeGomez: I can generate paths based on the cwd to make things simpler. I'll add tests to enforce it. :+1:

askama_derive/src/input.rs
GuillaumeGomez: Let me double-check. Maybe I got confused on the way...

GuillaumeGomez: I can't reproduce what I did yesterday so I reverted this change.

GuillaumeGomez:
Finally made tests work on CI. \o/

I added a ui test as well (took me more time than I expected...).

askama_parser/src/lib.rs
djc: My suggestion was to extract the call to `std::env::current_dir()` from `strip_common()`, having the caller pass in a `&Path` containing the `cwd`.

GuillaumeGomez: Oh I see. That will make tests easier indeed.

GuillaumeGomez:
Updated!

askama_derive/src/input.rs
```diff
@@ -104,15 +104,15 @@ impl TemplateInput<'_> {
         &self,
         map: &mut HashMap<PathBuf, Parsed>,
     ) -> Result<(), CompileError> {
-        let source = match &self.source {
-            Source::Source(s) => s.clone(),
-            Source::Path(_) => get_template_source(&self.path)?,
+        let (source, source_path) = match &self.source {
+            Source::Source(s) => (s.clone(), None),
+            Source::Path(_) => (get_template_source(&self.path)?, Some(self.path.clone())),
```
djc, line 109: Do we actually need to `clone()` here? Why?

GuillaumeGomez: Unless I missed something, we can't pass `self.path` as value because it's part of `self`. And if we pass as reference, it'll create lifetime issues below in the `add_to_check` closure. Do you see a way to go around that?

GuillaumeGomez: Something I thought about: we could use `Rc` to prevent having all these clones. What do you think?

GuillaumeGomez:
Rebased on #994.

testing/tests/ui.rs
```diff
@@ -21,4 +23,27 @@ fn ui() {
...
+
+    // To be able to use existing templates, we create a link to the `templates` folder.
+    let manifest_dir = match std::env::var("CARGO_MANIFEST_DIR") {
+        Ok(manifest_dir) => manifest_dir,
+        Err(_) => panic!("you need to run tests with `cargo`"),
+    };
+    let target = Path::new(&manifest_dir).join("../target/tests/trybuild/askama_testing");
+    if !target.exists() {
+        if let Err(err) = std::fs::create_dir_all(&target) {
+            panic!("Failed to create folder `{}`: {err:?}", target.display());
```
djc, line 35: Nit: lowercase panic messages, please.

testing/tests/ui.rs
```diff
@@ -21,4 +23,27 @@ fn ui() {
...
+        }
+    }
+    let target = target.canonicalize().unwrap().join("templates");
+    if !target.exists() {
```
djc, line 39: Nit: please invert the condition and return early.

testing/tests/ui.rs
```diff
@@ -21,4 +23,27 @@ fn ui() {
...
+        let original = Path::new(&manifest_dir).join("templates");
+        if std::os::unix::fs::symlink(&original, &target).is_err() {
```
djc, line 41: Nit: please import `use std::os::unix::fs::symlink` at the top of the file.

testing/tests/ui.rs
```diff
@@ -21,4 +23,27 @@ fn ui() {
...
+            panic!(
+                "Failed to create to create link on `{}` as `{}`",
```
djc, line 43: Nit: lowercase panic message.

askama_parser/src/lib.rs
```diff
@@ -97,13 +113,21 @@ impl<'a> Ast<'a> {
         let (row, last_line) = source_before.lines().enumerate().last().unwrap_or_default();
         let column = last_line.chars().count();
 
+        let file_info = file_path.and_then(|file_path| {
+            let cwd = std::env::current_dir().ok()?;
+            Some((cwd, file_path))
+        });
         let msg = format!(
-            "{}problems parsing template source at row {}, column {} near:\n{}",
+            "{}problems parsing template source{} at row {}, column {} near:\n{}",
```
djc, line 121: So Rust messages look like this:

```rust
error: mismatched closing delimiter: `}`
   --> askama_derive/src/config.rs:95:13
    |
25  |     ) -> std::result::Result<Config<'a>, CompileError> {
    |                                                        - closing delimiter possibly meant for this
...
95  |         for (extensions, path in DEFAULT_ESCAPERS {
    |             ^ unclosed delimiter
...
106 |     }
    |     ^ mismatched closing delimiter
```

I propose that we make ours look more like:

```
error: failed to parse template source
  --> testing/templates/invalid_syntax.html:1:14
       "}"
 --> tests/ui/error_file_path.rs:3:10
  |
3 | #[derive(Template)]
  |          ^^^^^^^^
  |
```

(With exactly the second line appearing only if we have a source location here.)

askama_parser/src/lib.rs
```diff
@@ -487,3 +511,77 @@ fn filter(i: &str, level: Level) -> ParseResult<'_, (&str, Option<Vec<Expr<'_>>>
...
+    for current_path_component in base.components() {
+        match components_iter.peek() {
+            Some(path_component) => {
+                if current_path_component != *path_component {
+                    break;
+                }
+                components_iter.next();
+            }
+            None => return path.display().to_string(),
+        }
```
djc, line 540: Nit: looks like a case for `let-else`?

askama_parser/src/lib.rs
```diff
@@ -487,3 +511,77 @@ fn filter(i: &str, level: Level) -> ParseResult<'_, (&str, Option<Vec<Expr<'_>>>
     ))(i)?;
     Ok((i, (fname, args)))
 }
+
+/// The goal of this function is to reduce the path length based on the `base` argument
+/// (generally the path where the program is running into). For example:
```
djc, line 516: Please give this a proper description that fits on one line as the first line of the docstring, then an empty line, then more details.

testing/tests/ui.rs
GuillaumeGomez: I will need to add a `cfg` for it so I'll import it in the function instead (unless you prefer a `cfg` on top of the `use`?),

askama_parser/src/lib.rs
GuillaumeGomez: We don't assign here so `if let` maybe?

djc: I meant this:

```rust
let Some(path_component) = components_iter.peek() else {
    return path.display().to_string();
}
```

GuillaumeGomez: ... It's obvious when you see it. :sweat_smile:

testing/tests/ui.rs
djc: Maybe a separate commit that just hoists `#![cfg(not(windows))]` to the module level?

GuillaumeGomez: Sounds good to me.

askama_parser/src/lib.rs
GuillaumeGomez: Looks better indeed. :+1: Makes a lot of diff though. ^^'

testing/tests/ui.rs
```diff
@@ -1,9 +1,11 @@
+#![cfg(not(windows))]
```
djc, line 1: If we're going to rely on `std::os::unix`, should we use `#![cfg(unix)]` instead?

askama_parser/src/lib.rs
```diff
@@ -487,3 +511,77 @@ fn filter(i: &str, level: Level) -> ParseResult<'_, (&str, Option<Vec<Expr<'_>>>
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::strip_common;
+    use std::path::Path;
+
+    #[test]
+    #[cfg(not(windows))]
```
djc, line 558: Why can't we run this test on Windows? Feels like it might actually be important that we test this bit of platform-sensitive logic on Windows, too?

askama_parser/src/lib.rs
djc: Why does it make a lot of diff?

GuillaumeGomez: Because we updated the message and all UI tests get a new output.

askama_parser/src/lib.rs
GuillaumeGomez: Because the top component of a path on Windows seems to be multiple potential things, even if it's all the same (take a look at [Component::Prefix](https://doc.rust-lang.org/nightly/std/path/enum.Component.html#variant.Prefix) documentation for some extra info.

But in short, when generating the canonicalized path, it generated `\\C:\\thing` whereas normally you would expect `C:\\thing`. Based on that, I simply decided to postpone it for a later time to someone who actually cares about Windows. ^^'

testing/tests/ui/elif.stderr
```diff
@@ -1,5 +1,5 @@
 error: unknown `elif` keyword; did you mean `else if`?
-       problems parsing template source at row 1, column 16 near:
+       failed to parse template source at row 1, column 16 near:
```
djc, line 2: Note that this doesn't have the shorter location information as suggested in my other comment.

GuillaumeGomez: If there is no file name (for error coming from the parser, which isn't handled in here), I preferred to leave it as is.

askama_parser/src/lib.rs
djc: So what is the behavior going to be on Windows? Maybe adding these error locations is actually a good reason that we should have UI tests enabled on Windows.

GuillaumeGomez: > So what is the behavior going to be on Windows? 

Keeping the full path for most cases (which isn't the best but still better than no path).

> Maybe adding these error locations is actually a good reason that we should have UI tests enabled on Windows.

That wouldn't work because of paths `/` vs `\\`. Or we would need to have two sets of ui tests.

testing/tests/ui/error_file_path.stderr
```diff
@@ -0,0 +1,29 @@
+error: failed to parse template source
+ --> testing/templates/invalid_syntax.html:1:14
```
djc, line 2: Okay, final nit: there should be two space before the `-->` to match rustc output, this will also help align the errors in case both show up.

djc (approved):

testing/tests/ui/error_file_path.stderr
GuillaumeGomez: Oh interesting. There are two whitespace characters. I tried adding more but it continues to display. I can enter the unicode realm to try to trick my way around if you want?

djc: No, never mind then.

GuillaumeGomez:
Thanks for the reviews! Merging then. :)

Issue #989, opened 2024-03-11
Scope changes inside `block`, preventing to call macros which should be in scope


GuillaumeGomez:
The bug can be reproduced with this code:

File `macro.html`:

```jinja
{% macro what(a, b) %}
    {{ a }} = {{ b }}
{% endmacro %}
```

File `base.html`:

```jinja
{% import "macro.html" as bla %}

{% block header %}{% endblock %}
```

File `main.rs`:

```rust
use askama::Template;

#[derive(Template)]
#[template(source = r#"{% extends "base.html" %}

{% block header %}
{% call bla::what(1, 2) %}
{% endblock %}"#, ext = "txt")]
struct F;

fn main() {
    assert_eq!(F.render().unwrap(), r#"Hello You"#);
}
```

This code will fail with this error:

```
error: no import found for scope "bla"
 --> src/main.rs:4:10
  |
4 | #[derive(Template)]
  |          ^^^^^^^^
  |
  = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)
```

djc:
So you're expecting that macro scopes imported in the base are available in the child template?

Seems kinda reasonable, let's see how much complexity in the generator it requires?

djc:
So this is supported by Tera, right? What about Jinja? I'm still on the fence about whether this is a good idea -- file-scoped imports actually seem pretty reasonable to me, similar to how things work in Rust.

And, this is actually easy to work around, right?

GuillaumeGomez:
The whole explanation is [here](https://jinja.palletsprojects.com/en/3.1.x/templates/#block-nesting-and-scope). And I also need to not allow access to variables which are not defined in the `block` in my PR (currently it does allow it, which is invalid based on the description).

djc:
I wasn't suggesting per-block imports but per-file imports. It feels to me like macros and inheritance should operate independently.

GuillaumeGomez:
Apparently jinja2 is fine with this? We're supposed to inherits whatever the "parents" imported.

GuillaumeGomez:
Anyway, updated #991 to prevent inheriting variables.

PR #991, opened 2024-03-12
Allow to have access to imports in block calls

GuillaumeGomez:
Fixes https://github.com/djc/askama/issues/989.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -65,7 +65,7 @@ impl<'a> Generator<'a> {
     }
 
     // Takes a Context and generates the relevant implementations.
-    pub(crate) fn build(mut self, ctx: &'a Context<'_>) -> Result<String, CompileError> {
+    pub(crate) fn build(mut self, ctx: &Context<'a>) -> Result<String, CompileError> {
         let mut buf = Buffer::new(0);
 
         self.impl_template(ctx, &mut buf)?;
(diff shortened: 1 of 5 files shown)


askama_derive/src/generator.rs
```diff
@@ -60,7 +60,7 @@ impl<'a> Generator<'a> {
     }
 
     // Takes a Context and generates the relevant implementations.
-    pub(crate) fn build(mut self, ctx: &'a Context<'_>) -> Result<String, CompileError> {
+    pub(crate) fn build(mut self, ctx: &Context<'a>) -> Result<String, CompileError> {
```
GuillaumeGomez, line 63: I relaxed the lifetime rules as they were too strict for my use case. Like this we can pass a temporary `Context` as long as it points to the same data.

GuillaumeGomez:
Solved clippy lints as well.

Hopefully with this one, I'll have everything I need to end the switch from `tera` to `askama` on docs.rs. :)

GuillaumeGomez:
As discussed in https://github.com/djc/askama/issues/989, variables defined outside a block should not be accessible inside a block.

GuillaumeGomez:
Ah I need https://github.com/djc/askama/pull/986 to be merged first for this ui test to work. ^^'

GuillaumeGomez:
Rebased. It's now ready for review!

askama_derive/src/generator.rs
```diff
@@ -909,35 +905,57 @@ impl<'a> Generator<'a> {
...
             CompileError::from(match name {
                 None => format!("no super() block found for block '{}'", cur.0),
                 Some(name) => format!("no block found for name '{name}'"),
             })
         })?;
 
+        // We clone the context of the child in order to preserve their macros and imports.
+        // But also add all the imports and macros from this template that don't override the
+        // child's ones to preserve this template's context.
+        let mut child_ctx: Context<'a> = child_ctx.clone();
```
djc, line 923: Do we need the type annotation here? If so, why?

djc (approved):

GuillaumeGomez:
Since you approved, merging. Thanks for the reviews!

PR #993, opened 2024-03-29
Re-export comrak for version matching.

Zethtren:
Just added a feature gated re-export so comrak will be accesible via askama::comrak::{}; instead of having to install the correct version.

(This is my first time re-exporting so please deny if I'm incorrect).


djc (approved):
Seems okay. @Kijewski any thoughts?

Kijewski:
Thank you for your work! In #1003 we decided to remove the built-in `markdown` filter, so this PR is obsolete.

PR #994, opened 2024-03-29
derive: wrap paths in Rc

djc:


GuillaumeGomez (approved):

GuillaumeGomez:
So you went with what I suggested in https://github.com/djc/askama/pull/986#discussion_r1536635419, glad to see it. I'll update my PR.

PR #998, opened 2024-04-01
Remove options argument from built-in markdown filter

djc:
comrak does relatively many semver-incompatible releases, and because we expose direct access to `comrak::Options` via the `markdown` filter that makes Askama semver-incompatible. Since I doubt a lot of people depend on this and it can easily a custom filter, I'd prefer to just remove the semver hazard here.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -1201,10 +1201,9 @@ impl<'a> Generator<'a> {
         buf: &mut Buffer,
         args: &[Expr<'_>],
     ) -> Result<DisplayWrap, CompileError> {
-        let (md, options) = match args {
-            [md] => (md, None),
-            [md, options] => (md, Some(options)),
-            _ => return Err("markdown filter expects no more than one option argument".into()),
+        let md = match args {
+            [md] => md,
+            _ => return Err("markdown filter expects one argument".into()),
         };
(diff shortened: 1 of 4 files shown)


GuillaumeGomez (approved):

askama_derive/src/generator.rs
```diff
@@ -1201,25 +1201,16 @@ impl<'a> Generator<'a> {
         buf: &mut Buffer,
         args: &[Expr<'_>],
     ) -> Result<DisplayWrap, CompileError> {
-        let (md, options) = match args {
-            [md] => (md, None),
-            [md, options] => (md, Some(options)),
-            _ => return Err("markdown filter expects no more than one option argument".into()),
+        let md = match args {
+            [md] => md,
+            _ => return Err("markdown filter expects one argument".into()),
```
Kijewski, line 1206: I'd say "no argument", because the input is an implicit argument.

Kijewski (changes requested):
Yes, it makes sense to keep the built-in filters simple.

The last paragraph of <https://github.com/djc/askama/blob/main/book/src/filters.md#markdown> needs to be updated.

Kijewski:
We removed the markdown filter in #1003, so this PR is obsolete.

Issue #999, opened 2024-04-06
Axum documentation states `askama_axum` import is required, but it's not? 


jonkoops:
Hi, consider this as much a question as an issue report. I am quite new to the Rust ecosystem, so forgive me if this comes across as a tad ignorant. The documentation for the [Axum integration](https://djc.github.io/askama/integrations.html#axum-integration) states:

> In your template definitions, replace `askama::Template` with [`askama_axum::Template`](https://docs.rs/askama_axum).

However, when I have both `askama` and `askama_axum` installed, I can also use the 'regular' `askama::Template` without getting any compilation issues. For example:

```rust
use askama::Template;
use axum::{routing::get, Router};

#[tokio::main]
async fn main() {
    tracing_subscriber::fmt::init();

    let app = Router::new().route("/", get(index));
    let listener = tokio::net::TcpListener::bind("0.0.0.0:3000").await.unwrap();

    axum::serve(listener, app).await.unwrap();
}

#[derive(Template)]
#[template(path = "index.html")]
struct IndexTemplate {}

async fn index() -> IndexTemplate {
    IndexTemplate {}
}
```

Is this intended to work as such, and is it ok to keep using it like this? Why would one need to use `asaka_axum` and what extra does it do on top of the regular template?

I am also interested in how Cargo manages dependencies. If I install `askama_axum` I can see that `askama` is [listed as a dependency](https://crates.io/crates/askama_axum/0.4.0/dependencies), so I would expect that if I list `askama_axum` as the **only dependency** in `Cargo.toml`, it would also bring `axum` along. 

However, if I modify the code above to use `askama_axum::Template` as the documentation states, and only install `askama_axum`, then I get the following compilation error:

```log
error[E0433]: failed to resolve: could not find `askama` in the list of imported crates
  --> src/main.rs:14:10
   |
14 | #[derive(Template)]
   |          ^^^^^^^^ could not find `askama` in the list of imported crates
   |
   = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This is strange because `Template` is imported from `askama_axum` and not `askama`. But it leads me to believe that I need to have both `askama_axum` and `askama` listed in `Cargo.toml`, even though my code does not directly depend on `askama`. Is this intentional behavior? Should I be listing both? If so, why is `askama` not pulled in with `askama_axum`?

Again, total novice here. Feel free to close this issue if it's too much of a hassle to answer. Thanks in advance.

[...]

PR #1000, opened 2024-04-06
 Use re-exported askama::* in integration tests, etc.

Kijewski:
Integration crates should use `askama_…::Template` instead of `askama::Template` <https://github.com/djc/askama/issues/999#issuecomment-2041132503>.

Also some minor clean ups in the integration crates. Removes dev-dependencies:

```
Removing futures-lite v2.3.0
Removing h2 v0.4.4
Removing hyper v1.2.0
Removing parking v2.2.0
Removing tokio-test v0.4.4
```

### DIFF
--- a/askama_axum/Cargo.toml
+++ b/askama_axum/Cargo.toml
@@ -14,7 +14,7 @@ workspace = ".."
 readme = "README.md"
 
 [dependencies]
-askama = { version = "0.13", path = "../askama", default-features = false, features = ["with-axum", "mime", "mime_guess"] }
+askama = { version = "0.13", path = "../askama", default-features = false, features = ["with-axum"] }
 axum-core = "0.4"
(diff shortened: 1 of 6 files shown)


GuillaumeGomez (approved):

jonkoops (commented):
Why are `mime` and `mime_guess` being removed as dependent features? Will the mime type still be determined correctly. And if not, then should this be considered a breaking change?

Kijewski:
> Why are mime and mime_guess being removed as dependent features?

The features don't do anything for quite some time: #594. We determine the mime type of the Template at compile time now (using the extension of the template file, but with the option of a manual override).

djc (approved):
Nice!

djc:
PR/issue #1000, cool milestone!

PR #1001, opened 2024-04-06
Clean up features

Kijewski:
* Use `dep:` prefix to suppress implicit features: Our msrv is 1.65 by now, so we can safely use this rust 1.60 feature. This is a breaking change, because users could use either `features = ["serde-json"]` or `["serde_json"]` to get the same result. With this change, only the former remains valid.

* The features `"mime"` and `"mime_guess"` are deprecated and do nothing since [#594]. Keeping such deprecated features around can still cause confusion whether one needs to select them or not. It's been 2,5 year since [#594] was merged, so I guess we can drop the features.

[#594]: https://github.com/djc/askama/pull/594

### DIFF
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -30,10 +30,6 @@ with-axum = ["askama_derive/with-axum"]
 with-rocket = ["askama_derive/with-rocket"]
 with-warp = ["askama_derive/with-warp"]
 
-# deprecated
-mime = []
-mime_guess = []
-
 [dependencies]
 askama_derive = { version = "0.13", path = "../askama_derive" }
 askama_escape = { version = "0.10.3", path = "../askama_escape" }
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -19,12 +19,12 @@ maintenance = { status = "actively-developed" }
 [features]
 default = ["config", "humansize", "num-traits", "urlencode"]
 config = ["askama_derive/config"]
-humansize = ["askama_derive/humansize", "dep_humansize"]
-markdown = ["askama_derive/markdown", "comrak"]
-num-traits = ["askama_derive/num-traits", "dep_num_traits"]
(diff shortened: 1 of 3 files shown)


askama/Cargo.toml
```diff
@@ -19,12 +19,12 @@ maintenance = { status = "actively-developed" }
...
-humansize = ["askama_derive/humansize", "dep_humansize"]
-markdown = ["askama_derive/markdown", "comrak"]
-num-traits = ["askama_derive/num-traits", "dep_num_traits"]
-serde-json = ["askama_derive/serde-json", "askama_escape/json", "serde", "serde_json"]
-serde-yaml = ["askama_derive/serde-yaml", "serde", "serde_yaml"]
-urlencode = ["askama_derive/urlencode", "percent-encoding"]
+humansize = ["askama_derive/humansize", "dep:humansize"]
+markdown = ["askama_derive/markdown", "dep:comrak"]
+num-traits = ["askama_derive/num-traits", "dep:num-traits"]
+serde-json = ["askama_derive/serde-json", "askama_escape/json", "dep:serde", "dep:serde_json"]
```
djc, line 25: For `serde_json` I think we should prefer the upstream spelling for the feature, now that we can do this. Maybe keep `serde-json` around as an alias (but change any mentions in documentation to the underscore variant)?

Kijewski:
@GuillaumeGomez, do you know if it is possible to generate this page <https://docs.rs/crate/askama/0.12.1/features> locally?

djc (approved):

GuillaumeGomez:
> @GuillaumeGomez, do you know if it is possible to generate this page https://docs.rs/crate/askama/0.12.1/features locally?

We use `cargo metadata --format-version 1` and then list the features from there. We could then generate a markdown from that and pass it as argument to rustdoc (like `rustdoc features.md`, because yes, rustdoc can generate HTML from markdown directly :smile: ).

djc:
@Kijewski ping?

PR #1003, opened 2024-04-06
Remove `yaml` filter

### LINKED ISSUE #1002: Do we need a built-in yaml filter?
I don't really see a compelling use case for a built-in yaml filter. It's a serialization format widely used for input configurations, not for generated web output. The feature was asked for in #192, and it was added in #201.

The package serde_yaml is now marked as [deprecated and unmaintained](https://github.com/dtolnay/serde-yaml/commit/3ba8462f7d3b603d832e0daeb6cfc7168a673d7a) by its author. If we keep the feature, then maybe we should look around if there is an active fork of the project?

Cc @antifuchs.

GuillaumeGomez:
Fixes #1002.


askama_derive/src/generator.rs
```diff
@@ -1273,29 +1273,24 @@ impl<'a> Generator<'a> {
...
-        }
 
-        const FILTERS: [&str; 2] = ["safe", "yaml"];
-        if FILTERS.contains(&name) {
+        let display_wrap = if name == "safe" {
             buf.write(&format!(
                 "{CRATE}::filters::{}({}, ",
                 name, self.input.escaper
             ));
+            DisplayWrap::Wrapped
```
djc, line 1282: Please move this into a separate commit if you want it (Honestly not sure if I like it.)

djc (commented):
So do we think the Markdown filter does get to stay? Why?

askama_derive/src/generator.rs
```diff
@@ -1273,29 +1273,24 @@ impl<'a> Generator<'a> {
...
-        }
+            DisplayWrap::Unwrapped
+        };
 
         self._visit_args(buf, args)?;
         buf.write(")?");
-        Ok(match FILTERS.contains(&name) {
-            true => DisplayWrap::Wrapped,
-            false => DisplayWrap::Unwrapped,
-        })
```
Kijewski, line 1298: Thanks for that change! Not searching the list twice is such a low-hanging fruit that I wonder why no one noticed that before. :)

GuillaumeGomez:
> So do we think the Markdown filter does get to stay? Why?

I was wondering about the same thing. Should I remove it too?

GuillaumeGomez:
And what about `json` btw?

djc:
JSON feels like it's like to be an order of magnitude more popular, so I'm on board with it staying. Maybe we could do some code search to see what features people enable in the wild...

Kijewski:
>  So do we think the Markdown filter does get to stay? Why?


Yeah, I guess running the input through comrak manually is easy enough. There seem to be users of the feature (#656, #719, #993), but I guess that's not enough to keep this feature around, esp. since comrak is (still) fast moving.

GuillaumeGomez:
Ok, let's remove `markdown` as well then. We can check for `json` later on.

Kijewski:
A json filter is essential, IMHO. Any Javascript interaction or `<a data-abc="…">` uses this feature. Or, at least I use this filter for these use cases.

askama_derive/src/generator.rs
Kijewski: Hehe, seems like we have the opposite opinion on this matter.

GuillaumeGomez:
Removed the `markdown` filter as well, which allowed to remove the `comrak` dependency (yeay).

Kijewski (approved):
Thanks, lgtm!

GuillaumeGomez:
Let's go!

djc:
Glad you managed to resolve the issue! Yes, I think things will improve from here.

PR #1005, opened 2024-04-11
Fix cyclic dependency

### LINKED ISSUE #1004: Template compilation infinite loop / failed cycle detection
Hello,

I ran into the same bug as is reported is issue #515. That issue was closed as solved in 2021, with some cycle detection added in, but I have a simple case the cycle detection fails to trigger on.

The following project will reproduce it, running cargo build never halts.

`src/lib.rs`:
```rust
use askama::Template;

#[derive(Template)]
#[template(path = "page.html")]
pub struct Page;
```

`templates/base.html`:
```
{% extends base.html %}
```
`templates/page.html`:
```
{% extends base.html %}
```

GuillaumeGomez:
Fixes #1004.

We forgot to check a corner-case apparently.

### DIFF
--- a/askama_derive/src/input.rs
+++ b/askama_derive/src/input.rs
@@ -140,15 +140,12 @@ impl TemplateInput<'_> {
                         Extends(extends) if top => {
                             let extends = self.config.find_template(extends.path, Some(&path))?;
                             let dependency_path = (path.clone(), extends.clone());
-                            if dependency_graph.contains(&dependency_path) {
-                                return Err(format!(
-                                    "cyclic dependency in graph {:#?}",
-                                    dependency_graph
-                                        .iter()
-                                        .map(|e| format!("{:#?} --> {:#?}", e.0, e.1))
-                                        .collect::<Vec<String>>()
-                                )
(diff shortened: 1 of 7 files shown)


djc (approved):

and-reas-se:
I looked at the PR noticed that both tests appear to test the same thing. Did you mean to have one of the test templates extend the other?

GuillaumeGomez:
They don't test the same things. One is the case that was already handled ("cyclic1.html" extending itself) and the other checks the case your discovered ("cyclic2.html" extending "cyclic1.html").

and-reas-se:
In the PR that does not appear to be the case, cycle1.html extends cycle1.html, and cycle2.html extends cycle2.html.

GuillaumeGomez:
... I'm ashamed. Thanks for the catch! Want to send a PR to update `cyclic2.html`?

and-reas-se:
Sure, I can do that.

GuillaumeGomez:
Awesome! To update UI test, you can do: `TRYBUILD=overwrite cargo test`.

PR #1006, opened 2024-04-11
Fix typo in regression test causing the wrong thing to be tested

and-reas-se:
See discussion in PR #1005


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1008, opened 2024-04-17
Make `|tojson` and few more filters faster 

Kijewski:
This ~~proof-of-concept~~ PR changes the filter `|tojson` so that it does not collect the serialized data into a string, but writes the data into the target writer directly.

This makes the benchmark run 81% (only serializing) or 39% (serializing and escaping for HTML) faster. The benchmarked data is not a fair representation for the data you would most likely escape, though.

### DIFF
--- a/askama/src/error.rs
+++ b/askama/src/error.rs
@@ -1,3 +1,4 @@
+use std::convert::Infallible;
 use std::fmt::{self, Display};
 
 pub type Result<I, E = Error> = ::std::result::Result<I, E>;
@@ -68,6 +69,13 @@ impl From<fmt::Error> for Error {
 impl From<::serde_json::Error> for Error {
     fn from(err: ::serde_json::Error) -> Self {
         Error::Json(err)
+    }
+}
+
+impl From<Infallible> for Error {
+    #[inline]
+    fn from(value: Infallible) -> Self {
+        match value {}
     }
 }
 
--- a/askama/src/filters/json.rs
+++ b/askama/src/filters/json.rs
@@ -19,12 +19,12 @@ use serde_json::to_writer_pretty;
 /// or in apostrophes with the (optional) safe filter `'{{data|json|safe}}'`.
 /// In HTML texts the output of e.g. `<pre>{{data|json|safe}}</pre>` is safe, too.
 #[inline]
-pub fn json<S: Serialize>(s: S) -> Result<ToJson<S>, Infallible> {
+pub fn json<S: Serialize>(s: S) -> Result<impl fmt::Display, Infallible> {
     Ok(ToJson(s))
 }
 
 #[derive(Debug, Clone)]
-pub struct ToJson<S: Serialize>(S);
+struct ToJson<S: Serialize>(S);
 
 impl<S: Serialize> fmt::Display for ToJson<S> {
     #[inline]
--- a/askama/src/filters/json.rs
+++ b/askama/src/filters/json.rs
@@ -1,5 +1,6 @@
-use crate::error::{Error, Result};
-use askama_escape::JsonEscapeBuffer;
+use std::convert::Infallible;
+use std::{fmt, io, str};
+
 use serde::Serialize;
 use serde_json::to_writer_pretty;
 
@@ -17,10 +18,58 @@ use serde_json::to_writer_pretty;
 /// To use it in HTML attributes, you can either use it in quotation marks `"{{data|json}}"` as is,
 /// or in apostrophes with the (optional) safe filter `'{{data|json|safe}}'`.
 /// In HTML texts the output of e.g. `<pre>{{data|json|safe}}</pre>` is safe, too.
-pub fn json<S: Serialize>(s: S) -> Result<String> {
-    let mut writer = JsonEscapeBuffer::new();
-    to_writer_pretty(&mut writer, &s).map_err(Error::from)?;
-    Ok(writer.finish())
+#[inline]
+pub fn json<S: Serialize>(s: S) -> Result<ToJson<S>, Infallible> {
+    Ok(ToJson(s))
+}
+
+#[derive(Debug, Clone)]
+pub struct ToJson<S: Serialize>(S);
+
+impl<S: Serialize> fmt::Display for ToJson<S> {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        to_writer_pretty(JsonWriter(f), &self.0).map_err(|_| fmt::Error)
+    }
+}
+
+struct JsonWriter<'a, 'b: 'a>(&'a mut fmt::Formatter<'b>);
+
+impl io::Write for JsonWriter<'_, '_> {
+    #[inline]
+    fn write(&mut self, bytes: &[u8]) -> io::Result<usize> {
+        self.write_all(bytes)?;
+        Ok(bytes.len())
+    }
+
+    #[inline]
+    fn write_all(&mut self, bytes: &[u8]) -> io::Result<()> {
+        write(self.0, bytes).map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err))
+    }
+
+    #[inline]
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
+fn write(f: &mut fmt::Formatter<'_>, bytes: &[u8]) -> fmt::Result {
+    let mut last = 0;
+    for (index, byte) in bytes.iter().enumerate() {
+        let escaped = match byte {
+            b'&' => Some(br"\u0026"),
+            b'\'' => Some(br"\u0027"),
+            b'<' => Some(br"\u003c"),
+            b'>' => Some(br"\u003e"),
+            _ => None,
+        };
+        if let Some(escaped) = escaped {
+            f.write_str(unsafe { str::from_utf8_unchecked(&bytes[last..index]) })?;
+            f.write_str(unsafe { str::from_utf8_unchecked(escaped) })?;
+            last = index + 1;
+        }
+    }
+    f.write_str(unsafe { str::from_utf8_unchecked(&bytes[last..]) })
 }
 
 #[cfg(test)]
@@ -29,12 +78,16 @@ mod tests {
 
     #[test]
     fn test_json() {
-        assert_eq!(json(true).unwrap(), "true");
-        assert_eq!(json("foo").unwrap(), r#""foo""#);
-        assert_eq!(json(true).unwrap(), "true");
-        assert_eq!(json("foo").unwrap(), r#""foo""#);
+        assert_eq!(json(true).unwrap().to_string(), "true");
+        assert_eq!(json("foo").unwrap().to_string(), r#""foo""#);
+        assert_eq!(json(true).unwrap().to_string(), "true");
+        assert_eq!(json("foo").unwrap().to_string(), r#""foo""#);
         assert_eq!(
-            json(vec!["foo", "bar"]).unwrap(),
+            json("<script>").unwrap().to_string(),
+            r#""\u003cscript\u003e""#
+        );
+        assert_eq!(
+            json(vec!["foo", "bar"]).unwrap().to_string(),
             r#"[
   "foo",
   "bar"
--- a/askama/src/filters/mod.rs
+++ b/askama/src/filters/mod.rs
@@ -138,9 +138,9 @@ impl fmt::Display for FilesizeFormatFilter {
 /// To encode `/` as well, see [`urlencode_strict`](./fn.urlencode_strict.html).
 ///
 /// [`urlencode_strict`]: ./fn.urlencode_strict.html
-pub fn urlencode<T: fmt::Display>(s: T) -> Result<String> {
-    let s = s.to_string();
-    Ok(utf8_percent_encode(&s, URLENCODE_SET).to_string())
+#[inline]
+pub fn urlencode<T: fmt::Display>(s: T) -> Result<UrlencodeFilter<T>, Infallible> {
+    Ok(UrlencodeFilter(s, URLENCODE_SET))
 }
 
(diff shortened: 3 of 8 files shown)


Kijewski:
Also, I think the default `|tojson` filter should not prettify the JSON data. I guess it's unlikely that you want to show a JSON representation to the user, and it's much more likely that you want to fill some Javascript variable, never to be seen by the user.

askama/src/error.rs
```diff
@@ -71,6 +72,13 @@ impl From<::serde_json::Error> for Error {
     }
 }
 
+impl From<Infallible> for Error {
+    #[track_caller]
+    fn from(_: Infallible) -> Self {
+        unreachable!()
+    }
```
vallentin, line 79:
```suggestion
    fn from(value: Infallible) -> Self {
        match value {}
    }
```

Alternatively use an empty `match`, as that results in _no_ generated code, whereas `unreachable!()` inserts panic related code

askama/src/filters/json.rs
```diff
@@ -17,10 +18,58 @@ use serde_json::to_writer_pretty;
...
+    Ok(ToJson(s))
+}
+
+#[derive(Debug, Clone)]
+pub struct ToJson<S: Serialize>(S);
+
+impl<S: Serialize> fmt::Display for ToJson<S> {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        to_writer_pretty(JsonWriter(f), &self.0).map_err(|_| fmt::Error)
```
vallentin, line 32:
```suggestion
        #[cfg(debug_assertions)]
        to_writer_pretty(JsonWriter(f), &self.0).map_err(|_| fmt::Error)
        #[cfg(not(debug_assertions))]
        to_writer(JsonWriter(f), &self.0).map_err(|_| fmt::Error)
```

Would we be able to do something like this? Because I agree with what you said. It's more likely that it's just going to be used to fill some JavaScript variable. However, I could definitely see the case of a debug build, where you're trying to inspect the output

askama/src/filters/json.rs
```diff
@@ -17,10 +18,58 @@ use serde_json::to_writer_pretty;
...
+        let escaped = match byte {
+            b'&' => Some(br"\u0026"),
+            b'\'' => Some(br"\u0027"),
+            b'<' => Some(br"\u003c"),
+            b'>' => Some(br"\u003e"),
+            _ => None,
+        };
+        if let Some(escaped) = escaped {
+            f.write_str(unsafe { str::from_utf8_unchecked(&bytes[last..index]) })?;
+            f.write_str(unsafe { str::from_utf8_unchecked(escaped) })?;
```
vallentin, line 68:
```suggestion
        let escaped = match byte {
            b'&' => Some(r"\u0026"),
            b'\'' => Some(r"\u0027"),
            b'<' => Some(r"\u003c"),
            b'>' => Some(r"\u003e"),
            _ => None,
        };
        if let Some(escaped) = escaped {
            f.write_str(unsafe { str::from_utf8_unchecked(&bytes[last..index]) })?;
            f.write_str(escaped)?;
```

We can at least remove one `unsafe` :)

Kijewski: I copied this code verbatim from the current implementation. The rationale is that you cannot have a sized `str`, so `r"\u0026"` will load two words instead of just one.

I think the current implementation is "good enough" until [`#feature(ascii_char)` is stable. Then we can strip one `unsafe` instruction without making the escaper slower. :)

askama/src/filters/json.rs
Kijewski: Yes, I like this solution. Or could it be confusion this the user that the debug and production built produce different data?

Alternatively, maybe we could add a `|pretty_json` filter?

vallentin: Overall, maybe there should be a `pretty_json`/`tojson_pretty` indeed.

For debugging it would however still require manually changing `tojson` -> `tojson_pretty`. But yeah, explicitness probably wins, compared to the potential confusion.

_Somewhat related, I also think `#[template(..., print = "ast")]` should print using `{:#?}` instead of `{:?}`_

askama/src/filters/json.rs
vallentin: Maybe it's because it's late and I'm tired. But can you elaborate on the difference? I just glanced through the generated assembly, between using `r"\u0026"` vs `br"\u0026"`, and they result in bytes? Also since `&str` is essentially _just_ a fancy `&[u8]` _anyways_. What am I missing? :)

Kijewski: Please have a look at <https://github.com/djc/askama/pull/805>. It can be that the compiler got better at optimizing in the meantime, though. I did not compare the generated code.

But I think changing that the implementation should be done in another PR, what do you think?

vallentin: Ah, that is interesting. I compared the assembly of before and after #805, along with a quick third variant I just wrote.

I definitely do see some differences. I do notice the increase of quad-words you mentioned before. I wonder if the compiler is simply better at optimizing a byte array. Whereas is seems it is less likely to treat a (even static) string as a byte array.

All in all, very interesting. But yes I agree, this is for a future separate PR.

askama/src/filters/mod.rs
```diff
@@ -297,20 +297,53 @@ pub fn trim<T: fmt::Display>(s: T) -> Result<String> {
...
 /// Limit string length, appends '...' if truncated
 #[inline]
-pub fn truncate(s: impl ToString, len: usize) -> Result<String, Infallible> {
-    fn truncate(s: String, len: usize) -> Result<String, Infallible> {
-        let mut s = s.to_string();
-        if s.len() > len {
-            let mut real_len = len;
-            while !s.is_char_boundary(real_len) {
-                real_len += 1;
+pub fn truncate(s: impl fmt::Display, len: usize) -> Result<String> {
```
vallentin, line 300: Would it maybe be possible to instead return `Result<Truncator>`, where `Truncator` `impl fmt::Display`, so we 100% avoid the string allocation.

_Also, maybe it's just a coincidence. But I was going to run through the filters, and refactor them into `impl fmt::Display` when possible, after adding `display_some` ;)_

djc:
> Also, I think the default `|tojson` filter should not prettify the JSON data. I guess it's unlikely that you want to show a JSON representation to the user, and it's much more likely that you want to fill some Javascript variable, never to be seen by the user.

What are the costs of making JSON pretty? I'm inclined to think the cost of JSON encoding will probably not show up for people and in most contexts pretty JSON will be helpful more than it will be harmful (because of increased size). So there's a trade-off:

1. Performance cost of rendering pretty vs compact
2. Bandwidth cost of transmitting pretty vs compact
3. Developer experience benefits of rendering pretty vs compact

I would argue that (3) is likely to outstrip (1) and (2), but obviously it depends on the situation.

askama/src/filters/mod.rs
Kijewski: Good idea! Will change.

Kijewski: Changed.

Kijewski:
Let's keep `|tojson` pretty for now, and maybe discuss in a new issue if we should change it later?

askama/src/filters/mod.rs
```diff
@@ -297,20 +297,69 @@ pub fn trim<T: fmt::Display>(s: T) -> Result<String> {
...
+impl<S: fmt::Display> fmt::Display for TruncateFilter<S> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        struct Writer<'a, 'b> {
+            dest: Option<&'a mut fmt::Formatter<'b>>,
+            remaining: usize,
+        }
+
+        impl fmt::Write for Writer<'_, '_> {
+            fn write_str(&mut self, s: &str) -> fmt::Result {
+                if let Some(dest) = self.dest.as_deref_mut() {
```
djc, line 321: Invert this for an early return?

askama/src/error.rs
```diff
@@ -71,6 +72,13 @@ impl From<::serde_json::Error> for Error {
     }
 }
 
+impl From<Infallible> for Error {
```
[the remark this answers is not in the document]
Kijewski: Every single commit makes the return type `Result<_, Infallible>` for its respective filter. I could merge the change with the first commit of the PR, but I guess it can keep it its own change.

askama_escape/src/lib.rs
```diff
@@ -165,51 +165,10 @@ pub trait Escaper {
         W: Write;
 }
 
-/// Escape chevrons, ampersand and apostrophes for use in JSON
```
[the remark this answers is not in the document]
Kijewski: Squashed.

askama/src/filters/mod.rs
```diff
@@ -259,25 +262,43 @@ where
...
-        if num > 0 {
-            rv.push_str(separator);
+impl<I, S> fmt::Display for JoinFilter<I, S>
+where
+    I: IntoIterator,
+    I::Item: fmt::Display,
+    S: fmt::Display,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some((iter, separator)) = self.0.take() {
```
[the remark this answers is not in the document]
Kijewski: Done.

askama/src/filters/mod.rs
```diff
@@ -259,25 +262,43 @@ where
...
-    let mut rv = String::new();
+/// Result of the filter [`join()`].
+///
+/// ## Note
+///
+/// This struct implements [`fmt::Display`], but only produces a string once.
+/// Any subsequent call to `.to_string()` will result in an empty string, because the iterator is
+/// already consumed.
+#[derive(Debug)]
+pub struct JoinFilter<I, S>(RefCell<Option<(I, S)>>);
```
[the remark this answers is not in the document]
Kijewski: It does not actually need `RefCell`. It only needs RefCell in order to implement `Debug`, but I guess the type does not need to be `Debug`. Changed to `Cell`.

It needs to be a `Cell`, though, because we consume the iterator, and I don't want to force it to be `Clone`.

askama/src/filters/json.rs
```diff
@@ -17,10 +18,58 @@ use serde_json::to_writer_pretty;
...
-    let mut writer = JsonEscapeBuffer::new();
-    to_writer_pretty(&mut writer, &s).map_err(Error::from)?;
-    Ok(writer.finish())
+#[inline]
+pub fn json<S: Serialize>(s: S) -> Result<ToJson<S>, Infallible> {
+    Ok(ToJson(s))
+}
+
+#[derive(Debug, Clone)]
+pub struct ToJson<S: Serialize>(S);
```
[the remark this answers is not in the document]
Kijewski: I did not read [RFC 3498](https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html) properly, but until I do, I tried to stay away from `-> impl Something` if I can.

askama/src/filters/mod.rs
```diff
@@ -139,9 +139,17 @@ impl fmt::Display for FilesizeFormatFilter {
 /// To encode `/` as well, see [`urlencode_strict`](./fn.urlencode_strict.html).
 ///
 /// [`urlencode_strict`]: ./fn.urlencode_strict.html
-pub fn urlencode<T: fmt::Display>(s: T) -> Result<String> {
-    let s = s.to_string();
-    Ok(utf8_percent_encode(&s, URLENCODE_SET).to_string())
+#[inline]
+pub fn urlencode(s: impl fmt::Display) -> Result<UrlencodeFilter, Infallible> {
+    Ok(UrlencodeFilter(s.to_string()))
```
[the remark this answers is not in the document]
Kijewski: Not using this [`percent-encoding`](https://docs.rs/percent-encoding/latest/percent_encoding/fn.percent_decode.html). The library wants to work on byte slices.

I guess we could implement the functionality ourself, but the implementor won't be me. :) If I need to encode query strings, then I'd prefer using [`serde_qs`](https://crates.io/crates/serde_qs) 9/10 times, so I can use the same datatype for serialization and deserialization.

askama/src/filters/mod.rs
Kijewski: Done.

askama/src/filters/mod.rs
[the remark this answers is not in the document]
Kijewski: Ah, I did not think this through: You can simply escape substrings, and print out the result. This works because we know that we won't get truncated UTF-8 characters, or surrogate pairs.

askama/src/filters/json.rs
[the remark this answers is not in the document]
djc: Not sure I understand your argument here -- I've used a bunch of times and it's usually pretty straightforward. If lifetime captures don't do the right thing, just specify it with an explicit bound.

I'd like to see this in this PR.

askama/src/filters/mod.rs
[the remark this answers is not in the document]
djc: Can you add a comment explaining why it needs to be in a `Cell`?

askama/src/filters/json.rs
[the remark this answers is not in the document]
Kijewski: You're not using it on nightly with edition="2024", though, do you? That will change the capturing rules.

My preferred solution would be to return a concrete type, but making that type inaccessible. But you added `#![deny(unreachable_pub)]`, so I guess you aren't a fan of that solution. :)

I can make the return type `-> impl Display` if you like that better.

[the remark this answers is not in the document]
djc: Yes, I do.

Kijewski:
So, I let all filters return `impl fmt::Display`. This has the added advantage that we can change the implementation of e.g. `|trim` without it being a semver incompatible change.

djc (approved):
Thanks, great work!

PR #1009, opened 2024-04-18
Updated book to include if let

vallentin:
Title :)

_There was also a handful of lines with a trailing single space._


Kijewski (approved):
Thanks! 👍🏻

djc:
Thanks!

PR #1012, opened 2024-04-18
Make `build_fallback()` infallible

Kijewski:
Fix up to <https://github.com/djc/askama/pull/1011>.

### DIFF
--- a/askama_derive/src/input.rs
+++ b/askama_derive/src/input.rs
@@ -103,6 +103,25 @@ impl TemplateInput<'_> {
         })
     }
 
+    pub(crate) fn fallback<'n>(
+        ast: &'n syn::DeriveInput,
+        config: &'n Config<'n>,
+        args: &'n TemplateArgs,
+    ) -> TemplateInput<'n> {
+        let ext = args.ext.as_deref();
+        TemplateInput {
+            ast,
+            config,
+            syntax: config.syntaxes.iter().next().unwrap().1,
+            source: args.source.as_ref().unwrap(),
+            print: args.print,
+            escaper: config.escapers.first().unwrap().1.as_str(),
+            ext,
+            mime_type: extension_to_mime_type(ext.unwrap()).to_string(),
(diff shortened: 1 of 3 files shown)


askama_derive/src/lib.rs
```diff
@@ -21,26 +21,29 @@ use input::{Print, TemplateArgs, TemplateInput};
...
+    // This way the user only sees the error message about the parsing error, not that `.render()`
+    // is not implemented.
+    let mut e: TokenStream = e.into_compile_error();
+    e.extend(build_fallback(&ast));
+    e
 }
 
-fn build_skeleton(ast: &syn::DeriveInput) -> Result<String, CompileError> {
+// This function generates empty implementations for `askama::Template` and friends for a `struct`.
+fn build_fallback(ast: &syn::DeriveInput) -> TokenStream {
```
djc, line 38: Would prefer to inline this.

askama_derive/src/lib.rs
```diff
@@ -21,26 +21,29 @@ use input::{Print, TemplateArgs, TemplateInput};
...
-            if let Ok(source) = build_skeleton(&ast) {
-                let source: TokenStream = source.parse().unwrap();
-                e.extend(source);
-            }
-            e
-        }
-    }
+    let e = match build_template(&ast) {
+        Ok(source) => return source.parse().unwrap(),
+        Err(e) => e,
```
djc, line 26: Let's do `e.into_compile_error()` here (no type annotation needed), and bind it to `tokens`.

askama_derive/src/input.rs
```diff
@@ -103,6 +103,25 @@ impl TemplateInput<'_> {
         })
     }
 
+    pub(crate) fn fallback<'n>(
```
djc, line 106: What's the purpose of this function? What does it do that `new()` doesn't (since it takes the same arguments anyway)? Should have comments for this.

Kijewski:
Not pursued.

PR #1014, opened 2024-04-21
Implemented display_some and display_some_or

### LINKED ISSUE #1007: Add display_some filter and friends
The question of how to render a `Option<T>` comes up every now and then. So I think it's time to provide a simpler built-in solution.

- Related #717
- Partially related #814

Today, I answered a [question on Stack Overflow], including 4 solutions. I think we should include a built-in `display_some` filter.

[question on Stack Overflow]: https://stackoverflow.com/a/78335414/2470818

In #717 @djc suggested naming it `or_empty`.

_Personally, I think the name `display_some` is more clear, since `val|or_empty`, might be confusing for e.g. `val: Option<i32>` or any other `Option<T>` than `Option<String>`._

-----

I've had a handful of `display_*` filters locally since forever. So I propose that we include (some of) the following:

```rust
pub fn display_some<T>(value: &Option<T>) -> askama::Result<String>
where
    T: std::fmt::Display;

pub fn display_some_or<T, U>(value: &Option<T>, otherwise: U) -> askama::Result<String>
where
    T: std::fmt::Display,
    U: std::fmt::Display;
```

Personally, I think `display_some_or`'s `otherwise` should remain as `U`, instead of `&Option<U>`. Since the main need I've had over the years, have been in the form of e.g.:

```jinja
{{ self.id|display_some_or("[no id]") }}
```

If I needed `a.or(b).or(c).unwrap_or("default")`, then I would instead do:

```jinja
{% if self.a.is_some() %}
    {{ self.a|display_some }}
{% else if self.b.is_some() %}
    {{ self.b|display_some }}
{% else if self.c.is_some() %}
    {{ self.c|display_some }}
{% else %}
    default
{% endif %}
```

_Even though this logic should probably be handled outside of the template in the first place._

-----

Additionally, my local implementation doesn't actually return `String`, they return `Display*`, e.g. `DisplaySome`, where `DisplaySome` impl `fmt::Display` to avoid the `String` allocation.

_So actually, what I want to add is:_

```rust
pub fn display_some<T>(value: &Option<T>) -> askama::Result<DisplaySome<'_, T>>
where
    T: std::fmt::Display,
{
    Ok(DisplaySome(value))
}

...
```

-----

Additionally, I also have `display_*` functions for `Result`, which might also be useful to include:

```rust
pub fn display_ok<T, E>(value: &Result<T, E>) -> askama::Result<String>
where
    T: std::fmt::Display;

pub fn display_err<T, E>(value: &Result<T, E>) -> askama::Result<String>
where
    E: std::fmt::Display;

pub fn display_ok_err<T, E>(value: &Result<T, E>) -> askama::Result<String>
where
    T: std::fmt::Display,
    E: std::fmt::Display;
```

_The `display_ok_err()` is mainly useful for debugging._

The `display_ok` and `display_err` filters, are mainly useful when you want to render `Ok` vs `Err` in separate places, e.g.

```jinja
{% if self.res.is_ok() %}
    {{ self.res|display_ok }}
{% endif %}

...

{% if self.res.is_err() %}
    Error: {{ self.res|display_err }}
{% endif %}
```

Compared to:

```jinja
{% match self.res %}
    {% when Ok with (val) %}
        {{ val }}
    {% else %}
{% endmatch %}

...

{% match self.res %}
    {% when Err with (err) %}
        Error: {{ err }}
    {% else %}
{% endmatch %}
```

I have the implementation locally, so I just need some opinions on which filters to include as built-in, and any other comments, before making a PR for them.

vallentin:
Resolves #1007

Questions:
- Should we `#[doc(hidden)]` `DisplaySome` and `DisplaySomeOr`?
- Should we move them into a separate module? (Maybe `#[doc(hidden)]` that?)
- Maybe even expand it to also hide the JSON related types in `json.rs`
  - Related #1008

Alternatively, we could also turn `DisplaySome` private, and then use:

```rust
fn display_some(...) -> askama::Result<impl fmt::Display + 'a>
```

### DIFF
--- a/askama/src/filters/mod.rs
+++ b/askama/src/filters/mod.rs
@@ -630,4 +630,43 @@ mod tests {
         assert_eq!(wordcount("foo").unwrap(), 1);
         assert_eq!(wordcount("foo bar").unwrap(), 2);
     }
+
+    #[test]
+    fn test_display_some() {
+        assert_eq!(display_some(&None::<String>).unwrap().to_string(), "");
+        assert_eq!(display_some(&None::<i32>).unwrap().to_string(), "");
+
+        assert_eq!(
+            display_some(&Some("hello world")).unwrap().to_string(),
+            "hello world"
+        );
+        assert_eq!(display_some(&Some(123)).unwrap().to_string(), "123");
+    }
+
+    #[test]
+    fn test_display_some_or() {
+        assert_eq!(
+            display_some_or(&None::<String>, "default")
+                .unwrap()
+                .to_string(),
+            "default"
+        );
+        assert_eq!(
+            display_some_or(&None::<i32>, "default")
+                .unwrap()
+                .to_string(),
+            "default"
+        );
+
+        assert_eq!(
+            display_some_or(&Some("hello world"), "default")
+                .unwrap()
+                .to_string(),
+            "hello world"
+        );
+        assert_eq!(
+            display_some_or(&Some(123), "default").unwrap().to_string(),
+            "123"
+        );
+    }
 }
--- a/askama/src/filters/mod.rs
+++ b/askama/src/filters/mod.rs
@@ -336,6 +336,54 @@ pub fn wordcount<T: fmt::Display>(s: T) -> Result<usize> {
     Ok(s.split_whitespace().count())
 }
 
+pub struct DisplaySome<'a, T>(Option<&'a T>);
+
+impl<T> fmt::Display for DisplaySome<'_, T>
+where
+    T: fmt::Display,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(val) = self.0 {
+            write!(f, "{val}")?;
+        }
+        Ok(())
+    }
+}
+
+pub fn display_some<T>(value: &Option<T>) -> Result<DisplaySome<'_, T>>
+where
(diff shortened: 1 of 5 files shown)


askama/src/filters/mod.rs
```diff
@@ -336,6 +336,54 @@ pub fn wordcount<T: fmt::Display>(s: T) -> Result<usize> {
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(val) = self.0 {
+            write!(f, "{val}")?;
+        }
+        Ok(())
+    }
+}
+
+pub fn display_some<T>(value: &Option<T>) -> Result<DisplaySome<'_, T>>
```
djc, line 358: Please put the filter functions before the types that need it.

Can we make `DisplaySome` and `DisplaySomeOr` private and yield `impl Display` from the filter functions?

book/src/filters.md
```diff
@@ -108,6 +110,46 @@ Output:
 -  a  -
 ```
 
+### display_some
+
+[#display_some]: #display_some
+
+The `display_some` filter is essentially a shorthand for:
```
djc, line 117: Nit: I think we don't need the "a" before "shorthand".

Kijewski:
I think we can omit the explicit lifetimes: 

```rust
```

Issue #1016, opened 2024-04-23
Float literals are not completely supported


GuillaumeGomez:
Can be checked with this code:

```jinja
{% set x = 12. %}
```

and this code:

```jinja
{% set x = .12 %}
```

GuillaumeGomez:
Only the first syntax is correct but because of the reasons mentioned in #1017, for now it's not supported.

PR #1017, opened 2024-04-23
Add more tests for floating parsing

GuillaumeGomez:
Fixes https://github.com/djc/askama/issues/1016.

I couldn't find a way to ensure that `1.` wasn't the start of `1..` so for now, I just added more tests about it in case someone else wants to give it a try.

As for `.1` syntax, rust doesn't allow it's easy: just adding tests for it and not changing anything.


djc (approved):

djc:
Thanks!

Issue #1018, opened 2024-04-23
Add support for single quote strings?


GuillaumeGomez:
[jinja spec](https://jinja.palletsprojects.com/en/3.0.x/templates/#literals) mentions that strings can be either single or double quotes. It is often useful for syntax highlighting when you have cases like:

```jinja
<a href="{{ func('bla') }}">
```

Would you be open for me to implement it?

djc:
It seems like an anti-feature to me in the context of Askama/Rust -- we would not be able to support `char` properly.

(Also not sure why double quotes wouldn't work in the example you gave.)

GuillaumeGomez:
Fair point. What about emitting an error if it's not a char then to prevent having to debug with generated code instead?

Also, as I mentioned, it was only for syntax highlighting. In my IDEs (emacs and sublime text), if I use double quotes strings in double quotes string, they're both a bit lost. :)

djc:
> Fair point. What about emitting an error if it's not a char then to prevent having to debug with generated code instead?

Does it currently error? How? I think making this properly a `char` would be okay.

> Also, as I mentioned, it was only for syntax highlighting. In my IDEs (emacs and sublime text), if I use double quotes strings in double quotes string, they're both a bit lost. :)

Hmm, sounds they're like confused about what syntax they're parsing?

GuillaumeGomez:
> Does it currently error? How? I think making this properly a `char` would be okay.

jinja considers `'string'` the same as `"string"`: both are strings. In the case of askama, it takes `'string'` and doesn't make any changes to it, generating invalid rust code since `'string'` is not a valid char, which generates this error:

```
error: character literal may only contain one codepoint
```

We could detect this error in askama and I think we should. It would make user experience much better.

> Hmm, sounds they're like confused about what syntax they're parsing?

Very likely. It's not a big issue, just something I thought could be improved. But having stricter types directly into the templates sounds better to me. We just need to improve how this error is handled in askama.

GuillaumeGomez:
Gonna send a PR later on then. I'm reaching the end of the port to askama in docs.rs. It's exciting. :)

djc:
Good to hear!

PR #1020, opened 2024-04-25
Add checks for chars

GuillaumeGomez:
Fixes https://github.com/djc/askama/issues/1018.

Now if a `char` is invalid, it will fail at parsing-time and not at compile-time anymore.


askama_parser/src/lib.rs
```diff
@@ -599,4 +686,37 @@ mod test {
...
+        // Should fail.
+        assert!(char_lit("''").is_err());
+        assert!(char_lit("'\\o'").is_err());
+        assert!(char_lit("'\\x'").is_err());
+        assert!(char_lit("'\\x1'").is_err());
+        assert!(char_lit("'\\x80'").is_err());
+        assert!(char_lit("'\\u'").is_err());
+        assert!(char_lit("'\\u{}'").is_err());
+        assert!(char_lit("'\\u{110000}'").is_err());
+    }
```
djc, line 721: Can you try with some multi-byte characters, like emoji or Chinese/Japanese code points? `anychar` is called that, but the documentation suggest it actually consumes bytes.

GuillaumeGomez: Luckily for us you thought about that: I was using `.len()` instead of `.chars().count()`, leading to a wrong result. Fixed it.

askama_parser/src/lib.rs
```diff
@@ -350,13 +350,100 @@ fn str_lit(i: &str) -> ParseResult<'_> {
...
+            }));
+        };
+        if nb > max_value {
+            return Err(nom::Err::Failure(ErrorContext {
+                input: start,
+                message: Some(Cow::Borrowed(err2)),
+            }));
+        }
+        Ok(())
+    };
```
djc, line 421: The only variable this is capturing is `s`, right? I'd prefer to make this a real function with a docstring to explain what it tries to do and what the arguments/return value mean.

GuillaumeGomez: `start` as well. I didn't make it a function because it's a lot of arguments but I can if you prefer.

djc: Also I think you don't need to take `strip_last_bracket`, can check whether the prefix ends with `{`.

GuillaumeGomez: It's a (small) performance impact that can be avoided, so I preferred to rely on a boolean instead. I can also do it as such if you prefer though.

djc: Yes, I'd prefer to avoid it. I think we're into edge cases of edge cases here, so making minor optimizations like this definitely feels like premature optimization.

GuillaumeGomez: :+1:

GuillaumeGomez:
Transformed the function into a closure and now map the error on failure.

askama_parser/src/lib.rs
```diff
@@ -350,13 +350,112 @@ fn str_lit(i: &str) -> ParseResult<'_> {
     Ok((i, s.unwrap_or_default()))
 }
 
+/// If this is an ascii or unicode endpoint, it has a maximum value which is checked in this
+/// function.
+///
+/// If it fails, we return `Err` which contains the error message.
+fn validate_endpoint(
```
djc, line 357: Nit: this should go below `char_lit`.

askama_parser/src/lib.rs
```diff
@@ -350,13 +350,112 @@ fn str_lit(i: &str) -> ParseResult<'_> {
...
+        return Err(err1);
+    };
+    if nb > max_value {
+        return Err(err2);
+    }
+    Ok(())
+}
+
+// Information about allowed character escapes is available at:
+// <https://doc.rust-lang.org/reference/tokens.html#ascii-escapes>.
```
djc, line 382: I think we should reference https://doc.rust-lang.org/reference/tokens.html#character-literals here. It feels like `recognize()` should yield something more like an `enum Char { CharLiteral(_), QuoteEscape(_), AsciiEscape(_), UnicodeEscape(_) }`, then we can add inline code that applies extra validation for the latter two variants.

Right now this code just passes different lists of 4 arguments that actually all depend on the single bit of ASCII vs Unicode so I think we should make that more clear.

askama_parser/src/lib.rs
```diff
@@ -350,13 +350,130 @@ fn str_lit(i: &str) -> ParseResult<'_> {
...
-    Ok((i, s.unwrap_or_default()))
+    let Some(s) = s else {
+        return Err(nom::Err::Failure(ErrorContext {
+            input: start,
+            // Same error as rustc.
+            message: Some(Cow::Borrowed("empty character literal")),
+        }));
+    };
+    let res: ParseResult<'_, Char<'_>> = Char::parse(s);
+    match res {
```
djc, line 443: Let's inline `res`.

askama_parser/src/lib.rs
```diff
@@ -350,13 +350,130 @@ fn str_lit(i: &str) -> ParseResult<'_> {
...
+    /// An escaped character (like `\n`) which doesn't require any extra check.
+    Escaped,
+    /// Ascii escape (like `\x12`).
+    AsciiEscape(&'a str),
+    /// Unicode escape (like `\u{12}`).
+    UnicodeEscape(&'a str),
+}
+
+impl<'a> Char<'a> {
+    fn parse(i: &'a str) -> ParseResult<'a, Self> {
```
djc, line 366: Make this a `FromStr` impl?

askama_parser/src/lib.rs
```diff
@@ -350,13 +350,130 @@ fn str_lit(i: &str) -> ParseResult<'_> {
...
+    match res {
+        Ok(("", c)) => {
+            c.validate().map_err(|error| {
+                nom::Err::Failure(ErrorContext {
+                    input: start,
+                    message: Some(Cow::Borrowed(error)),
+                })
+            })?;
+            Ok((i, s))
+        }
```
djc, line 452: Yield `c` out of the `match` here and return early for the `_ => Err(..)` branch.

askama_parser/src/lib.rs
GuillaumeGomez: It requires to implement:

```rust
fn from_str(s: &str) -> Result<Self, Self::Err> {
```

So it's gonna be more trouble than it's worth imo.

askama_parser/src/lib.rs
```diff
@@ -350,13 +350,130 @@ fn str_lit(i: &str) -> ParseResult<'_> {
...
+                    map(char('0'), |_| Self::Escaped),
+                    map(char('\''), |_| Self::Escaped),
+                    // Not useful but supported by rust.
+                    map(char('"'), |_| Self::Escaped),
+                    map(
+                        tuple((
+                            char('x'),
+                            take_while_m_n(2, 2, |c: char| c.is_ascii_hexdigit()),
+                        )),
+                        |_| Self::AsciiEscape(i),
```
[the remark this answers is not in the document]
GuillaumeGomez: Oh! Great idea!

GuillaumeGomez:
Reworked as suggested.

djc (approved):

GuillaumeGomez:
Thanks for the reviews!

PR #1023, opened 2024-04-28
Make JSON prettifying optional

### LINKED ISSUE #1019: Allow output of not prettified JSON data
Hi,

right know, `|tojson` will always prettify the output. This can lead to noisy output, e. g. when constructing a stringified object for a `data-` attribute. I think it would be a nice addition to toggle this functionality, perhaps with a new filter or with a parameter on the existing filter if that it possible? 

This idea was already brought up in #192 and #1008.

What do you think?

Kijewski:
This PR adds an optional argument to the `|tojson` filter, which controls if the serialized JSON data gets prettified or not. The arguments works the same as flask's [`|tojson`][flask] filter, which passes the argument to python's [`json.dumps()`][python]:

* Omitting the argument, providing a negative integer, or `None`, then compact JSON data is generated.
* Providing a non-negative integer, then this amount of ASCII spaces is used to indent the data. (Capped to 16 characters.)
* Providing a string, then this string is used as prefix. No attempts are made to ensure that the prefix actually consists of whitespaces, because chances are, that if you provide e.g. `&nsbp;`, then you are doing it intentionally.

This is a breaking change, because it changes the default behavior to not prettify the data. This is done intentionally, because this is how it works in flask.

[flask]: https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.tojson
[python]: https://docs.python.org/3/library/json.html#json.dump

---

This PR is based on #1008, i.e. only the last commit is actually part of this PR.

Resolves #1019.

### DIFF
--- a/askama/src/filters/json.rs
+++ b/askama/src/filters/json.rs
@@ -2,7 +2,7 @@ use std::convert::Infallible;
 use std::{fmt, io, str};
 
 use serde::Serialize;
-use serde_json::to_writer_pretty;
+use serde_json::ser::{to_writer, PrettyFormatter, Serializer};
 
 /// Serialize to JSON (requires `json` feature)
 ///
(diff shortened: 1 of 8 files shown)


askama/src/filters/json.rs
```diff
@@ -19,17 +19,74 @@ use serde_json::to_writer_pretty;
...
+}
+
+impl AsIndent for str {
+    #[inline]
+    fn as_indent(&self) -> Option<&str> {
+        Some(self)
+    }
+}
+
+impl AsIndent for isize {
```
djc, line 40: Does this still work if we impl for `usize` directly?

askama/src/filters/json.rs
```diff
@@ -19,17 +19,74 @@ use serde_json::to_writer_pretty;
 /// or in apostrophes with the (optional) safe filter `'{{data|json|safe}}'`.
 /// In HTML texts the output of e.g. `<pre>{{data|json|safe}}</pre>` is safe, too.
 #[inline]
-pub fn json<S: Serialize>(s: S) -> Result<impl fmt::Display, Infallible> {
-    Ok(ToJson(s))
+pub fn json<S: Serialize, I: AsIndent>(
+    value: S,
+    indent: I,
```
djc, line 24: Nit: would like to use `impl AsIndent` directly here (same for `value`).

askama/src/filters/json.rs
```diff
@@ -19,17 +19,74 @@ use serde_json::to_writer_pretty;
...
+) -> Result<impl fmt::Display, Infallible> {
+    Ok(ToJson { value, indent })
```
djc, line 26: Is there reason not to store an `Option<'static str>` in `ToJson::indent`?

askama/src/filters/json.rs
```diff
@@ -19,17 +19,74 @@ use serde_json::to_writer_pretty;
...
+struct ToJson<S, I> {
+    value: S,
+    indent: I,
+}
+
+impl<S: Serialize, I: AsIndent> fmt::Display for ToJson<S, I> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        to_writer_pretty(JsonWriter(f), &self.0).map_err(|_| fmt::Error)
+        let f = JsonWriter(f);
+        if let Some(indent) = self.indent.as_indent() {
```
djc, line 81: Nit: would prefer to `match` this, yielding out the `self.indent` value and doing an early return for the `None` case.

book/src/filters.md
```diff
@@ -415,6 +415,18 @@ Ugly: <script>var data = "{{data|json}}";</script>
 Ugly: <script>var data = '{{data|json|safe}}';</script>
 ```
 
+By default, a compact representation of the data is generated, i.e. no whitespaces are generated
```
djc, line 418: I'd still prefer to stick with pretty by default.

askama/src/filters/json.rs
Kijewski: Changed.

askama/src/filters/json.rs
Kijewski: Not if it should support jinja's API where negative value = no beautification.

askama/src/filters/json.rs
Kijewski: Yes, the lifetimes wouldn't work.

askama/src/filters/json.rs
Kijewski: No indentation is not the same as no beautification. A Some("") prefix still generates newlines, None does not. So both cases are entirely different. I could replace the `if` with a `match`, though.

book/src/filters.md
Kijewski: IMO, that's a bad default. The cases when you want to look at the generated JSON data will be few, esp. because you cannot specify an initial indentation level. And if you need it, why are two characters a good default? Why not four? Why not a tab character? Why not hot dogs?

[](https://archive.ph/xwm35)

AndrolGenhald: FWIW as a user I've used the json filter exclusively to assign js variables, so I think unpretty is a perfectly sane default.

Kijewski:
Are there open questions left?

GuillaumeGomez:
I'm not a big fan of this change. Wouldn't it better to have different filters for each case? Or have the current `json` filter unchanged and new one which allows to pass options?

GuillaumeGomez:
Mostly because most of the time, the current `json` filter does exactly what I want. Rust doesn't allow to have default values if you don't specify them, which is fine but not really compatible with jinja here. So in case you don't want the default filter values, you could call the other `json` filter (`json_with_options` for example) and give the options you want for this call.

djc:
> Mostly because most of the time, the current `json` filter does exactly what I want. Rust doesn't allow to have default values if you don't specify them, which is fine but not really compatible with jinja here.

So you want prettified JSON most of the time? While Rust doesn't allow default values, the Askama template language can.

GuillaumeGomez:
If you mean allowing a variable number of arguments to the `json` filter, then yes, I very much like this approach.

PR #1024, opened 2024-04-28
Remove obsolete trybuild outputs

Kijewski:
The current main branch does not compile with rust 1.58 or earlier anyway.


testing/Cargo.toml
```diff
@@ -18,8 +18,7 @@ serde_json = { version = "1.0", optional = true }
 
 [dev-dependencies]
 criterion = "0.5"
-trybuild = "1.0.76"
-version_check = "0.9"
+trybuild = "1.0.91"
```
djc, line 21: What's the rationale for bumping the trybuild version here? Are we actually depending on newer API?

Kijewski: That change was in the wrong commit. I tried to make the ui tests run on windows, too, but I cannot get it to output the paths like in windows. I thought that would be an easy fix, but, well, it's not. Maybe would have to error locations with forward slashes?

I undid this change.

djc (approved):

djc:
Nice, thanks!

PR #1028, opened 2024-05-01
fuzz: Add cifuzz integration

manunio:
Add cifuzz workflow action to have fuzzers build and run on each PR.
This is a service offered by oss-fuzz where askama already runs. cifuzz can help catch shallow bugs, regressions and fuzzing build issues before they are merged into the repository, to do this fuzzing will be run for ~5min.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks! 5 minutes feels kind of long when the complete CI workflow right now completes in 3.5. Can we make it like 3 minutes, or do we lose a lot of that to setup stuff?

PR #1029, opened 2024-05-02
Nested templates block

### LINKED ISSUE #1022: Compilation error when using nested template blocks with in a single template file
Hello! I Encountered a compilation error when using[ block fragments](https://github.com/djc/askama/pull/824) with Vec<&str> in a single template file. The error does not occur when the template is split into separate files.

This is working like expected
``` html
{% block index %}
Section 1: {{ s1 }}
{% endblock %}

{% block section %}
  Value ={{ value }}
{% endblock %}
```

``` rust
use askama::Template;

#[derive(Template)]
#[template(path = "combain.txt", block="index")]
struct RenderInPlace<'a> {
   s1: Section<'a>
}

#[derive(Template)]
#[template(path = "combain.txt", block="section")]
struct Section<'a> {
   value: &'a str,
}
fn main() {
    let value = "x";

    let t = RenderInPlace { s1: Section { value } };
    println!("{}", t);
}
```
``` sh
Section 1: 


  Value =x
```

When changing to Vec for the value i get a compilation error..
``` html 
{% block index %}
Section 1: {{ s1 }}
{% endblock %}

{% block section %}
  {% for value in  values %}
    Value = {{ value }}
  {% endfor %}
{% endblock %}
```
``` rust
use askama::Template;

#[derive(Template)]
#[template(path = "combain.txt", block="index")]
struct RenderInPlace<'a> {
   s1: Section<'a>
}

#[derive(Template)]
#[template(path = "combain.txt", block="section")]
struct Section<'a> {
   values: Vec<&'a str>,
}

fn main() {

    let values: Vec<&str>= vec!["a", "b", "c"];

    let t = RenderInPlace { s1: Section { values } };
    println!("{}", t);
}

```
``` sh
error[E0609]: no field `value` on type `&Section<'a>`
 --> src/main.rs:9:10
  |
9 | #[derive(Template)]
  |          ^^^^^^^^ unknown field
  |

```
However when i split the template file in two it is working as expected.
`index.html`
``` html
{% block index %}
Section 1: {{ s1 }}
{% endblock %}
```
`section.html`
``` html
{% block section %}
  {% for value in  values %}
    Value ={{ value }}
  {% endfor %}
{% endblock %}
```

``` rust
use askama::Template;

#[derive(Template)]
#[template(path = "index.txt", block="index")]
struct RenderInPlace<'a> {
   s1: Section<'a>
}

#[derive(Template)]
#[template(path = "section.txt", block="section")]
struct Section<'a> {
   values: Vec<&'a str>,
}

fn main() {

    let values: Vec<&str>= vec!["a", "b", "c"];

    let t = RenderInPlace { s1: Section { values } };
    println!("{}", t);
}
```
```sh 
Section 1: 
  
    Value =a
  
    Value =b
  
    Value =c
  
```

GuillaumeGomez:
Fixes #1022.


GuillaumeGomez:
And fixed clippy lints.

testing/tests/block_fragments.rs
```diff
@@ -11,7 +11,7 @@ struct FragmentSimple<'a> {
 fn test_fragment_simple() {
     let simple = FragmentSimple { name: "world" };
 
-    assert_eq!(simple.render().unwrap(), "\n\n<p>Hello world!</p>\n");
+    assert_eq!(simple.render().unwrap(), "\n<p>Hello world!</p>");
```
GuillaumeGomez, line 14: For all lines removal, I'm not sure whether it was bad before and this PR fixed it as a side effect or the opposite...

askama_derive/src/generator.rs
```diff
@@ -1813,6 +1817,7 @@ struct Buffer {
     indent: u8,
     // Whether the output buffer is currently at the start of a line
     start: bool,
+    discard: bool,
```
djc, line 1819: So the previous PR added `discard` in `WritableBuffer`. Remind me why we have both and why we need to discard in both (instead of discarding in one of them only)?

testing/tests/block_fragments.rs
```diff
@@ -79,5 +76,27 @@ fn test_fragment_unused_expression() {
...
+#[derive(Template)]
+#[template(path = "blocks.txt", block = "index")]
+struct RenderInPlace<'a> {
+    s1: Section<'a>,
+}
+
+#[derive(Template)]
+#[template(path = "blocks.txt", block = "section")]
+struct Section<'a> {
+    values: Vec<&'a str>,
```
djc, line 91: Nit: why couldn't this just be a slice?

testing/tests/block_fragments.rs
```diff
@@ -79,5 +76,27 @@ fn test_fragment_unused_expression() {
...
+}
+
+#[test]
+fn test_specific_block() {
+    let values: Vec<&str> = vec!["a", "b", "c"];
```
djc, line 96: Nit: pretty sure this doesn't need the type annotation.

askama_derive/src/generator.rs
```diff
@@ -734,7 +734,7 @@ impl<'a> Generator<'a> {
             }
         };
 
-        mem::drop(mem::replace(&mut self.buf_writable, current_buf));
+        mem::drop(mem::replace(&mut self.buf_writable.buf, current_buf));
```
Kijewski, line 737: This line could simply be `self.buf_writable.buf = current_buf;` and be much easier to read. Maybe you could sneak that change into the PR? :)

testing/tests/block_fragments.rs
djc: Looking at this particular case, IMO the previous state was good and this PR breaks it.

```
{% block body %}
<p>Hello {{ name }}!</p>
{% endblock %}
```

Arguably this should render newlines before and after the `p` node?

testing/tests/inheritance.rs
```diff
@@ -311,14 +311,9 @@ fn test_flat_deep() {
 
 #[derive(Template)]
 #[template(path = "let-base.html")]
+#[allow(dead_code)]
```
djc, line 314: The commit message doesn't quite make sense here, maybe clean that up?

askama_derive/src/generator.rs
GuillaumeGomez: Definitely!

askama_derive/src/generator.rs
GuillaumeGomez: One is the actual string where we push content and the other is pending AST that wasn't pushed yet into the string.

So in this case, if we only want to render a specific named block, which is then followed by another one (which we're not interested into), both will get rendered. However, the block we want to render could also be in other blocks, so we still need to go through the whole tree of each present block, but only render the block that was specified.

testing/tests/inheritance.rs
GuillaumeGomez: True, it's not a clippy lint but new rustc warnings. Gonna update the commit message.

GuillaumeGomez:
Mystery solved (for the missing trailing backline characters): I accidentaly removed `child.flush_ws(def.ws2);`. Much less diff in the tests now. \o/

rellfy:
I can confirm this resolved an issue I had where contents from the template leaked into the block-fragmented template. I added another test case here: https://github.com/djc/askama/pull/1032

djc (approved):

PR #1031, opened 2024-05-02
fuzz: trigger cifuzz only for askama_parser

manunio:
Cifuzz is triggered only for changes to askama_parser.


djc (approved):

djc:
Great idea, thanks!

djc:
Is your testing complete?

PR #1032, opened 2024-05-03
Test partial renders with block fragments

rellfy:
This complements the testing for https://github.com/djc/askama/pull/1029/files with respect to inheriting templates. The tests will pass after the fix PR is merged and this branch is updated.


.gitignore
```diff
@@ -1,3 +1,4 @@
 target
 Cargo.lock
 .DS_Store
+.idea
```
GuillaumeGomez, line 4: Missing a backline after this line.

Also I'm curious: what's this file?

GuillaumeGomez:
More tests are always welcome, thanks!

.gitignore
rellfy: Just pushed, thanks.

.idea is an autogenerated config directory for JetBrains IDEs

PR #1037, opened 2024-05-12
Fix invalid `if let` code generation

### LINKED ISSUE #1033: if let chain has a weird behaviour
I expected this code:

```jinja
{%- if let Some(query) = search_query && !query.is_empty() %}
```

To either emit an error about the `&&` after `if let Some` or to work, instead it generated:

```rust
if let Some(query,) = &(search_query && !self.query.is_empty()) {
```

Which is a very unexpected in-between.

I think it should generate `if let Some(query) = search_query && !self.query.is_empty()`. What do you think?

GuillaumeGomez:
Fixes #1033.

We cannot test the generated code since the if-let chain feature is still unstable. So I thought about two possibilities to still test this:
 * Testing the generated code.
 * Adding a new "nightly" feature for these tests and add a nightly check in the CI.

Here are the cases to be checked:

```jinja
{% if let Some(query) = s && !query.is_empty() %}{{query}}{% endif %}

{% if let Some(s) = s && !s.is_empty() %}{{s}}{% endif %}

{% if let Some(s) = s %}{{ s }}{% endif %}
```

Where `s` is `Option<String>`.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -364,12 +364,24 @@ impl<'a> Generator<'a> {
 
                 if let Some(target) = target {
                     let mut expr_buf = Buffer::new(0);
-                    self.visit_expr(&mut expr_buf, expr)?;
                     buf.write("let ");
-                    self.visit_target(buf, true, true, target);
-                    buf.write(" = &(");
+                    // If this is a chain condition, then we need to declare the variable after the
+                    // left expression has been handled but before the right expression is handled
+                    // but this one should have access to the let-bound variable.
+                    match expr {
+                        Expr::BinOp(op, ref left, ref right) if *op == "||" || *op == "&&" => {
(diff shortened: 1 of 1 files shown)


Kijewski (approved):

djc:
Maybe the generator should desugar let chains into something that works on stable? It's okay if the generated code isn't the prettiest.

GuillaumeGomez:
I don't think we should implement our own condition lexer, we should just rely on the rust compiler to handle this part for us. For example, it becomes particularly tricky in cases like:

```rust
if let Some(x) = y || z == "x" && whatever
```

It requires to generate `else if z == "x"` and in both `if` branches to have also `whatever`.

GuillaumeGomez:
I opened https://github.com/djc/askama/issues/1038 about this.

djc:
(Would prefer if you don't merge non-trivial PRs without my approval.)

PR #1039, opened 2024-05-14
Add `title` filters

GuillaumeGomez:
Part of https://github.com/djc/askama/issues/18.

I also used this opportunity to update the list in the first message of the issue.

### DIFF
--- a/askama/src/filters/mod.rs
+++ b/askama/src/filters/mod.rs
@@ -502,13 +502,49 @@ pub fn center(src: impl ToString, dst_len: usize) -> Result<impl fmt::Display, I
     center(src.to_string(), dst_len)
 }
 
-/// Count the words in that string
+/// Count the words in that string.
 #[inline]
 pub fn wordcount(s: impl ToString) -> Result<usize, Infallible> {
     fn wordcount(s: String) -> Result<usize, Infallible> {
         Ok(s.split_whitespace().count())
     }
     wordcount(s.to_string())
+}
+
+/// Converts a type to a `String`.
+#[inline]
+pub fn string(s: impl ToString) -> Result<String, Infallible> {
+    Ok(s.to_string())
+}
+
+/// Return a title cased version of the value. Words will start with uppercase letters, all
+/// remaining characters are lowercase.
+#[inline]
+pub fn title(s: impl ToString) -> Result<String, Infallible> {
+    let s = s.to_string();
+    let mut need_capitalization = true;
+
+    // Sadly enough, we can't mutate a string when iterating over its chars, likely because it could
+    // change the size of a char, "breaking" the char indices.
+    let mut output = String::with_capacity(s.len());
+    for c in s.chars() {
+        if c.is_whitespace() {
+            output.push(c);
+            need_capitalization = true;
+        } else if need_capitalization {
+            match c.is_uppercase() {
+                true => output.push(c),
+                false => output.extend(c.to_uppercase()),
+            }
+            need_capitalization = false;
+        } else {
+            match c.is_lowercase() {
+                true => output.push(c),
+                false => output.extend(c.to_lowercase()),
+            }
+        }
+    }
+    Ok(output)
 }
 
 #[cfg(test)]
@@ -775,5 +811,18 @@ mod tests {
         assert_eq!(wordcount(" \n\t").unwrap(), 0);
         assert_eq!(wordcount("foo").unwrap(), 1);
         assert_eq!(wordcount("foo bar").unwrap(), 2);
+        assert_eq!(wordcount("foo  bar").unwrap(), 2);
+    }
+
+    #[test]
+    fn test_title() {
+        assert_eq!(&title("").unwrap(), "");
+        assert_eq!(&title(" \n\t").unwrap(), " \n\t");
+        assert_eq!(&title("foo").unwrap(), "Foo");
+        assert_eq!(&title(" foo").unwrap(), " Foo");
+        assert_eq!(&title("foo bar").unwrap(), "Foo Bar");
+        assert_eq!(&title("foo  bar ").unwrap(), "Foo  Bar ");
+        assert_eq!(&title("fOO").unwrap(), "Foo");
+        assert_eq!(&title("fOo BaR").unwrap(), "Foo Bar");
     }
 }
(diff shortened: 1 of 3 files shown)


askama/src/filters/mod.rs
```diff
@@ -511,6 +511,39 @@ pub fn wordcount(s: impl ToString) -> Result<usize, Infallible> {
...
+    let mut need_capitalization = true;
+
+    // Sadly enough, we can't mutate a string when iterating over its chars, likely because it could
+    // change the size of a char, "breaking" the char indices.
+    Ok(s.chars().fold(String::new(), |mut acc, c| {
+        if c.is_whitespace() {
+            acc.push(c);
+            need_capitalization = true;
+        } else if need_capitalization {
+            if !c.is_uppercase() {
```
djc, line 528: I think the best formulation of this would use `match` instead of `if` and `push_str()` or `extend()` instead of the loop.

GuillaumeGomez: :+1:

GuillaumeGomez:
Applied suggestions and also added the `string` filter.

djc:
What's the point of the `string` filter if simply using an expression already relies on `Display`?

GuillaumeGomez:
In case you want to do operation on the generated string?

djc:
Do you have an actual use case for that?

askama/src/filters/mod.rs
```diff
@@ -502,7 +502,7 @@ pub fn center(src: impl ToString, dst_len: usize) -> Result<impl fmt::Display, I
     center(src.to_string(), dst_len)
 }
 
-/// Count the words in that string
+/// Counts the words in that string
```
djc: We're already inconsistent on the verb form and this is unrelated to the changes in this commit, so maybe just leave it for now?

(On the whole I think I prefer "Count"/"Center"/"Capitalize" over "Counts"/"Centers"/"Capitalizes".)

GuillaumeGomez:
Not personally. My potential use case would be to format a struct that was turned into a string. But I could write my own filter for that. I can remove the second commit of the PR if you're not convinced. ;)

askama/src/filters/mod.rs
```diff
@@ -511,6 +511,35 @@ pub fn wordcount(s: impl ToString) -> Result<usize, Infallible> {
     wordcount(s.to_string())
 }
 
+/// Returns a titlecased version of the value. I.e. words will start with uppercase letters, all
```
djc, line 520: Let's avoid "I.e." which is an obscure reference to latin words, and defer the second sentence to a second line of the docstring (after an empty line; and also update the book accordingly).

Also it appears "titlecase" is not an actual English word (per Merriam-Webster and DuckDuckGo), so just use "title case"? (Also in the docs.)

askama/src/filters/mod.rs
```diff
@@ -511,6 +511,35 @@ pub fn wordcount(s: impl ToString) -> Result<usize, Infallible> {
...
+/// remaining characters are lowercase.
+#[inline]
+pub fn title(s: impl ToString) -> Result<String, Infallible> {
+    let s = s.to_string();
+    let mut need_capitalization = true;
+
+    // Sadly enough, we can't mutate a string when iterating over its chars, likely because it could
+    // change the size of a char, "breaking" the char indices.
+    Ok(s.chars().fold(String::new(), |mut acc, c| {
```
djc, line 529: Maybe `String::new()` -> `String::with_capacity(s.len())`?

book/src/filters.md
```diff
@@ -20,25 +20,28 @@ Enable it with Cargo features (see below for more information).
...
+  * [`linebreaks`][#linebreaks],
+  * [`linebreaksbr`][#linebreaksbr],
+  * [`lower|lowercase`][#lower],
+  * [`safe`][#safe],
+  * [`string`][#string],
+  * [`title`][#title],
+  * [`trim`][#trim],
+  * [`truncate`][#truncate],
+  * [`upper|uppercase`][#upper],
+  * [`wordcount`][#wordcount]
```
djc, line 44: Would prefer the other additions to this be in a separate commit.

askama/src/filters/mod.rs
GuillaumeGomez: Very good point!

book/src/filters.md
GuillaumeGomez: Done!

askama/src/filters/mod.rs
GuillaumeGomez: Reverted this change.

askama/src/filters/mod.rs
GuillaumeGomez: Fixed.

djc:
> Not personally. My potential use case would be to format a struct that was turned into a string. But I could write my own filter for that. I can remove the second commit of the PR if you're not convinced. ;)

Yeah, let's wait until someone comes up with a use case.

GuillaumeGomez:
Removed it!

askama/src/filters/mod.rs
```diff
@@ -511,6 +511,42 @@ pub fn wordcount(s: impl ToString) -> Result<usize, Infallible> {
...
+/// Return a title cased version of the value. Words will start with uppercase letters, all
+/// remaining characters are lowercase.
+#[inline]
+pub fn title(s: impl ToString) -> Result<String, Infallible> {
+    let s = s.to_string();
+    let mut need_capitalization = true;
+
+    // Sadly enough, we can't mutate a string when iterating over its chars, likely because it could
+    // change the size of a char, "breaking" the char indices.
+    Ok(s.chars()
```
djc, line 529: I've been thinking about this but this latest reformatting (due to `new()` -> `with_capacity()`) I think pushes it over the line: a `for` loop would be a nice simplification instead of the `fold()` call.

askama/src/filters/mod.rs
djc: Note:

> (after an empty line; and also update the book accordingly).

GuillaumeGomez: :sweat_smile:

askama/src/filters/mod.rs
GuillaumeGomez: Ok, updating.

GuillaumeGomez:
Updated the book and replaced the `fold` iterator with a `for` loop.

djc (approved):

PR #1044, opened 2024-05-15
 Better error message for `{% macro super %} `

Kijewski:


askama_parser/src/node.rs
```diff
@@ -520,7 +520,13 @@ impl<'a> Macro<'a> {
                 |i| s.tag_block_end(i),
             ))),
         ));
-        let (i, (pws1, _, (name, params, nws1, _))) = start(i)?;
+        let (j, (pws1, _, (name, params, nws1, _))) = start(i)?;
+        if name == "super" {
```
Kijewski, line 524: I guess all [reserved keywords](https://doc.rust-lang.org/reference/keywords.html) should be restricted? What do you think?

GuillaumeGomez: I think that'd be better indeed.

djc (approved):

PR #1046, opened 2024-05-15
Remove instances of `use Enum::*`

Kijewski:
It's an anti-pattern to arbitrarily import enum variant, because it hides typos (`SpellingErrr => {}` will still work), renamed and removed variants.


djc (approved):

djc:
Disagree that it's always an anti-pattern (and I think clippy lints against the bad cases), but these changes seem fine.

GuillaumeGomez (commented):
I also think it's anti-pattern so big :+1: for me.

PR #1050, opened 2024-05-17
 Better error messages for truncated files

Kijewski:


askama_parser/src/node.rs
```diff
@@ -93,9 +93,21 @@ impl<'a> Node<'a> {
...
         s.leave();
-
         let (i, node) = result?;
-        let (i, _) = cut(|i| s.tag_block_end(i))(i)?;
+
+        let (i, closed) = cut(alt((
+            value(true, |i| s.tag_block_end(i)),
+            value(false, ws(eof)),
+        )))(i)?;
+        if !closed {
```
GuillaumeGomez, line 102: Since it's always almost the same except for one word, better put it into a function returning `Result<(), nom::Err>`, no?

GuillaumeGomez (commented):
Very nice improvement! Just one nit and looks good to me.

GuillaumeGomez (approved):
:+1:

askama_parser/src/node.rs
Kijewski: Yes, makes sense to unify that into one helper function. Done.

PR #1051, opened 2024-05-18
Re-add `{% set %}` as alias for `{% let %}`

Kijewski:
Commit [615fb82] introduced a regression. `{% set %}` and `{% let %}` should be interchangeable, but only the latter syntax works.

This PR fixes the problem, and adds a regression test.

[615fb82]: <https://github.com/djc/askama/commit/615fb82c65f38071b1ce1d9edab52cc0009e985d>


djc (approved):

djc:
Thanks!

PR #1054, opened 2024-05-21
Fix block partial rendering

### LINKED ISSUE #1052: Unnecessary fields required when using blocks for partial rendering
Given a template like this:

`templates/base.html`
```html
{% block empty %}
{% endblock %}

{% if let Some(var) = var %}
{{ var }}
{% endif %}
```

The following code:

```rust
#[derive(Template)]
#[template(path = "base.html", block = "empty")]
struct Empty {}
```

Gives an error:

```text
no field `var` on type `&Empty`
```

Due to the (empty) match statement generated by the derive macro, even though the field `var` is not necessary to render just the `empty` block.

GuillaumeGomez:
Fixes #1052.

It didn't take into account that there would be other elements at the top level. Now it does.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -989,6 +989,10 @@ impl<'a> Generator<'a> {
         if block_fragment_write {
             self.buf_writable.discard = true;
         }
+        // If we are rendering a specific block and the discard changed, it means that we're done
+        // with the block we want to render and that from this point, everything will be discarded.
+        //
+        // To get this block content rendered as well, we need to write to the buffer before then.
         if buf.discard != prev_buf_discard {
             self.write_buf_writable(buf)?;
         }
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -90,6 +90,7 @@ impl<'a> Generator<'a> {
         buf.write(CRATE);
         buf.writeln("::Result<()> {")?;
 
+        buf.discard = self.buf_writable.discard;
         // Make sure the compiler understands that the generated code depends on the template files.
         for path in self.contexts.keys() {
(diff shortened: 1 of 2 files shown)


askama_derive/src/generator.rs
```diff
@@ -987,6 +989,9 @@ impl<'a> Generator<'a> {
         if block_fragment_write {
             self.buf_writable.discard = true;
         }
+        if buf.discard != prev_buf_discard {
```
djc, line 992: This seems surprising to me -- why do we only want to write if `buf.discard` is not equal to `prev_buf_discard` vs writing it out if `buf.discard` is just `false`?

GuillaumeGomez: Interesting question. In my previous PR about this, if I wrote too early, we had content from sub-block that was rendered before its parent's, which was problematic. So in this case, if `buf.discard != prev_buf_discard`, then it's like that the parent is actually not something being displayed and so we should flush everything before the `discard` is `true` again.

Does my reasoning sound good to you and do you see potential corner-cases?

GuillaumeGomez: Just thought: do you want me to add a code comment to explain this?

djc: Yes, that would be a good improvement.

GuillaumeGomez: Done!

djc: This comment is too long IMO and still doesn't make it very clear what case this is trying to fix/prevent. I want a comment here in like 4-5 lines that explains conceptually what case it is trying to handle.

GuillaumeGomez: Updated the comment.

PR #1057, opened 2024-05-23
Add support for `elif`

### LINKED ISSUE #1056: Add  `elif` keyword
Hi,

we recently migrated from Tera to Askama and overall it was quite straightforward. One thing that did not work out of the box was the `elif` keyword, so it would be nice to include that. In addition, `elif` seems to be [supported by jinja](https://jinja.palletsprojects.com/en/3.1.x/templates/#if).

GuillaumeGomez:
Fixes #1056.

### DIFF
--- a/askama_parser/src/node.rs
+++ b/askama_parser/src/node.rs
@@ -372,21 +372,28 @@ impl<'a> Cond<'a> {
         let mut p = tuple((
             |i| s.tag_block_start(i),
             opt(Whitespace::parse),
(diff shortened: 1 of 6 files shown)


Kijewski:
I added a commit to deduplicate the both branches a bit. Please feel free to un-cherry-pick the commit if you don't like it. :)

djc (approved):

djc:
Nice!

Issue #926, opened 2023-12-06
Keep the template source code portion corresponding to each AST token


GuillaumeGomez:
In a lot of cases, we emit errors when generating the Rust code in the proc-macro, however we only emit an error without showing where it comes from in the template.

Don't improve this situation, we could keep a `Option<&str>` or at least a `Option<Range>` (which would point to the template source, however this second solution brings its own issues, like when the error comes from imports/extends).

What do you think?

djc:
Makes sense, looking forward to review!

djc:
@Kijewski friendly ping?

Kijewski:
Sorry, I'm kinda overwhelmed by work right now and did not see the message.

Of course, fine with me. Welcome, @GuillaumeGomez! :)

GuillaumeGomez:
:heart:

djc:
> Sorry, I'm kinda overwhelmed by work right now and did not see the message.

No problem, hope you can enjoy a nice holiday toward the end of the year!

PR #1058, opened 2024-05-24
Better (post-parser) errors

GuillaumeGomez:
Fixes https://github.com/djc/askama/issues/926.

This is very big because a lot of small changes everywhere. However, the result is really great (at least for me :smiley_cat: ). Take a look at the updated ui tests to see the impact of this PR.

### DIFF
--- a/askama_derive/src/heritage.rs
+++ b/askama_derive/src/heritage.rs
@@ -5,7 +5,7 @@ use std::rc::Rc;
 use crate::config::Config;
 use crate::CompileError;
 use parser::node::{BlockDef, Macro};
-use parser::Node;
+use parser::{Node, Parsed};
 
 pub(crate) struct Heritage<'a> {
     pub(crate) root: &'a Context<'a>,
@@ -36,26 +36,40 @@ impl Heritage<'_> {
 
 type BlockAncestry<'a> = HashMap<&'a str, Vec<(&'a Context<'a>, &'a BlockDef<'a>)>>;
 
-#[derive(Default, Clone)]
+#[derive(Clone)]
 pub(crate) struct Context<'a> {
     pub(crate) nodes: &'a [Node<'a>],
     pub(crate) extends: Option<Rc<Path>>,
     pub(crate) blocks: HashMap<&'a str, &'a BlockDef<'a>>,
     pub(crate) macros: HashMap<&'a str, &'a Macro<'a>>,
     pub(crate) imports: HashMap<&'a str, Rc<Path>>,
+    path: Option<&'a Path>,
+    parsed: &'a Parsed,
 }
 
 impl Context<'_> {
+    pub(crate) fn empty(parsed: &Parsed) -> Context<'_> {
+        Context {
+            nodes: &[],
+            extends: None,
+            blocks: HashMap::new(),
+            macros: HashMap::new(),
+            imports: HashMap::new(),
+            path: None,
+            parsed,
+        }
+    }
+
     pub(crate) fn new<'n>(
         config: &Config<'_>,
-        path: &Path,
-        nodes: &'n [Node<'n>],
+        path: &'n Path,
+        parsed: &'n Parsed,
     ) -> Result<Context<'n>, CompileError> {
         let mut extends = None;
         let mut blocks = HashMap::new();
         let mut macros = HashMap::new();
         let mut imports = HashMap::new();
-        let mut nested = vec![nodes];
+        let mut nested = vec![parsed.nodes()];
         let mut top = true;
 
         while let Some(nodes) = nested.pop() {
@@ -104,11 +118,13 @@ impl Context<'_> {
         }
 
         Ok(Context {
-            nodes,
+            nodes: parsed.nodes(),
             extends,
             blocks,
             macros,
             imports,
+            parsed,
+            path: Some(path),
         })
     }
 }
--- a/askama_derive/src/heritage.rs
+++ b/askama_derive/src/heritage.rs
@@ -76,7 +76,9 @@ impl Context<'_> {
             for n in nodes {
                 match n {
                     Node::Extends(e) if top => match extends {
-                        Some(_) => return Err("multiple extend blocks found".into()),
+                        Some(_) => {
+                            return Err(CompileError::no_file_info("multiple extend blocks found"))
+                        }
                         None => {
                             extends = Some(config.find_template(e.path, Some(path))?);
                         }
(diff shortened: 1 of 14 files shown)


askama_derive/src/input.rs
```diff
@@ -127,13 +132,16 @@ impl TemplateInput<'_> {
             let mut nested = vec![parsed.nodes()];
             while let Some(nodes) = nested.pop() {
                 for n in nodes {
-                    let mut add_to_check = |path: Rc<Path>| -> Result<(), CompileError> {
-                        if !map.contains_key(&path) {
+                    let mut add_to_check = |new_path: Rc<Path>| -> Result<(), CompileError> {
```
djc, line 135: Please avoid/isolate unrelated changes.

askama_derive/src/heritage.rs
```diff
@@ -36,33 +36,56 @@ impl Heritage<'_> {
...
     pub(crate) extends: Option<Rc<Path>>,
     pub(crate) blocks: HashMap<&'a str, &'a BlockDef<'a>>,
     pub(crate) macros: HashMap<&'a str, &'a Macro<'a>>,
     pub(crate) imports: HashMap<&'a str, Rc<Path>>,
+    path: Option<&'a Path>,
+    parsed: &'a Parsed,
 }
 
 impl Context<'_> {
+    pub(crate) fn new_empty(parsed: &Parsed) -> Context<'_> {
```
djc, line 51: Let's call this `empty()`.

djc (commented):
I like the end result too, but there is a lot in the diff I don't like. Please be careful about minimizing the necessary changes and make some smaller commits.

askama_derive/src/input.rs
```diff
@@ -55,18 +55,20 @@ impl TemplateInput<'_> {
                 PathBuf::from(format!("{}.{}", ast.ident, ext)).into()
             }
             (&Source::Source(_), None) => {
-                return Err("must include 'ext' attribute when using 'source' attribute".into())
+                return Err(CompileError::new(
```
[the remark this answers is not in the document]
GuillaumeGomez: I disagree. I removed the `From` implementations on purpose to ensure that we don't use `.into()` by mistake in the future. It's less convenient but it forces to think about whether or not you can provide context on this error or not. If not, then you use `None` and it's done.

Side effect: it also allows us (reviewers) to spot it more easily when reviewing and to check if it could in fact provide context to the error.

askama_derive/src/input.rs
GuillaumeGomez: It's not unrelated, I use the captured parent `path`, but to do so I needed to rename the closure argument. I use it here:

```rust
                            let source = get_template_source(
                                &new_path,
                                Some((&path, parsed.source(), n.template_content())),
                            )?;
```

askama_parser/src/expr.rs
```diff
@@ -44,36 +49,27 @@ macro_rules! expr_prec_layer {
...
-    Range(&'a str, Option<Box<Expr<'a>>>, Option<Box<Expr<'a>>>),
-    Group(Box<Expr<'a>>),
-    Tuple(Vec<Expr<'a>>),
-    Call(Box<Expr<'a>>, Vec<Expr<'a>>),
-    RustMacro(Vec<&'a str>, &'a str),
-    Try(Box<Expr<'a>>),
-    /// This variant should never be used directly. It is created when generating filter blocks.
-    Generated(String),
+#[derive(Clone, Debug)]
+pub struct Expr<'a> {
```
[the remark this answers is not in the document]
GuillaumeGomez: What would you name this new wrapper type? `ExprOrNode`?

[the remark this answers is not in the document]
GuillaumeGomez: That would force `Node::Expr` to use `Box` to prevent infinite type size too.

askama_derive/src/heritage.rs
GuillaumeGomez: 👍

askama_derive/src/input.rs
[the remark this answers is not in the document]
djc: This change should definitely not be in the same commit as all these other changes, because it is a noop change which makes the diff very verbose and thus hard to review. We can reintroduce it in a later commit, although I would suggest it might make more sense to attach `error()` methods to contexts like `TemplateInput` that can refer to the available context directly.

askama_parser/src/expr.rs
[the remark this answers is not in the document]
djc: More like:

```rust
struct WithSpan<'a, T> {
    inner: T,
    span: &'a str,
}

impl Deref for WithSpan<'a> {
    type Target = T;

    ..
}
```

GuillaumeGomez:
Finally reworked the whole git history and made the suggested changes.

askama_parser/src/lib.rs
```diff
@@ -155,6 +129,72 @@ impl fmt::Display for ParseError {
 pub(crate) type ParseErr<'a> = nom::Err<ErrorContext<'a>>;
 pub(crate) type ParseResult<'a, T = &'a str> = Result<(&'a str, T), ParseErr<'a>>;
 
+pub struct ErrorInfo {
```
djc, line 190: Although the direction is good, this still has too much going on. Suggest a few steps, maybe in a separate PR?

* Extract all of this code from `Ast::from_str()` into a `ParseError::new()` function
* Unify the row/column display to `{row}:{column}` for both path and non-path errors
* Extract an `ErrorInfo` type with an `impl Display` that takes care of the final formatting step, which will need to include a `path: Option<Rc<Path>>`
* To the extent that you need what is here `generate_row_and_column()` and `generate_error_info()` later in the PR, they should perhaps be constructors on `ErrorInfo`.

Also please keep functions in top-down order -- maybe do more self-review to check for style issues? Reviewing your changes is taking many round trips.

GuillaumeGomez: Feedback is appreciated. I don't plan to send another PR for this, all changes are tied together and at this point your suggestions (even if great) are only improvements over this code. I'll implement what you suggested though.

> Also please keep functions in top-down order -- maybe do more self-review to check for style issues? Reviewing your changes is taking many round trips.

Well, as you said, it's code styling, so we're entering personal preferences. I don't mind following a project code's style but if there is no test to enforce it and tell me what and where I did something wrong, then you shouldn't expect people to remember it or to implement it as such. Therefore, I don't plan to try to remember what style you want to apply until there is a check for it, I work on way too many different codebases to have motivation for this. If this is a blocker for you, then I'll just fork askama from this point and implement what I need on my side while backporting changes from your repository to mine. Just to be clear: this is not a hostile take or anything, I really love `askama` and really enjoy your suggestions. It's just that you have high expectations for PRs, and I don't have the tools nor the time to live up to your expectations. So I'll let it up to you. I don't mind if you send commits to my PRs to fix code formatting, this is very welcome as well.

GuillaumeGomez:
I'll try to update this PR in the next days since #1061 got merged. Still didn't applied your comments too, just came back from a hackfest.

PR #1061, opened 2024-06-07
Parser: improve error structure

djc:
Try to provide a better starting point for #1058.

### DIFF
--- a/askama_derive/Cargo.toml
+++ b/askama_derive/Cargo.toml
@@ -25,7 +25,7 @@ with-rocket = []
 with-warp = []
 
 [dependencies]
-parser = { package = "askama_parser", version = "0.2", path = "../askama_parser" }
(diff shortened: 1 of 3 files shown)


testing/tests/ui/excessive_nesting.stderr
```diff
@@ -1,4 +1,4 @@
-error: failed to parse template source at row 14, column 34 near:
+error: failed to parse template source near 14:34
```
GuillaumeGomez, line 1: I find this wording very confusing, in particular `near 14:34`. It doesn't follow the Rust "standard error output" and unless you know what it is already, it's hard to guess. I think the full sentence was better here. The alternative would be to retrieve the file name and then generate the file location like we do for the rest.

djc: Fair -- I killed the second commit.

GuillaumeGomez (approved):
Looks good, thanks!

PR #1062, opened 2024-06-10
filter: let `|truncate` only write '…' if needed

Kijewski:
The filter takes care not to split characters. If all data was printed, then '…' should not be printed.

### DIFF
--- a/askama/src/filters/mod.rs
+++ b/askama/src/filters/mod.rs
@@ -332,6 +332,11 @@ impl<S: fmt::Display> fmt::Display for TruncateFilter<S> {
                         while !s.is_char_boundary(rem) {
                             rem += 1;
                         }
+                        if rem == s.len() {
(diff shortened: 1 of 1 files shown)


Kijewski:
@djc, I guess it's up to you to say if you want to allow [Unicode-3.0](https://spdx.org/licenses/Unicode-3.0.html). :) Looks like MIT to me.

GuillaumeGomez (approved):

djc (approved):

djc:
> @djc, I guess it's up to you to say if you want to allow [Unicode-3.0](https://spdx.org/licenses/Unicode-3.0.html). :) Looks like MIT to me.

Not sure what do you mean by this?

djc:
Ah, the audit failure. Yeah, the idna crate update is a bit of a trashfire.

PR #1064, opened 2024-06-12
Add urlencode to filters.md

OscartGiles:
The urlencode filter was missing from the documentation. This adds it.

I placed it under `Built-in filters` rather than `Optional / feature gated filters`, even though it is behind the [urlencode](https://github.com/djc/askama/blob/ab5c6a3b39b53d63489160f0c381ae3c9ac7b806/askama/Cargo.toml#L20) feature gate, on the basis that this is in the default feature list. Let me know if preferred in the optional filters.


GuillaumeGomez (approved):

djc (approved):
Thanks!

Issue #1063, opened 2024-06-11
Include statement has no effect when used from a block fragment


OscartGiles:
When using a block fragment include statements don't seem to work.

See https://github.com/OscartGiles/askama/commit/74162b9453486a53a4de40bba905b4be31dcf3f0 for an example. 

Is this intented behaviour or a bug?

djc:
Do you get an error? What's the behavior you're seeing and how does it differ from the expected behavior?

OscartGiles:
Yes sorry, the test fails with:

```
running 1 test
thread 'test_fragment_include' panicked at testing/tests/block_fragments.rs:116:5:
assertion `left == right` failed
  left: "\n\n"
 right: "\nINCLUDED: world\n"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test test_fragment_include ... FAILED

failures:

failures:
    test_fragment_include
```

I was testing out block fragments for a project and had this issue. I was able to reproduce with this test.

Askama is awesome by the way! Really enjoying working with it.

djc:
Can you try if it works with #1054?

GuillaumeGomez:
Can you provide the template source code please? It'll be useful to add it as a regression test as well.

OscartGiles:
I've just created a PR with the failing test https://github.com/djc/askama/pull/1066

GuillaumeGomez:
Thanks!

PR #1066, opened 2024-06-13
Add a failing test for using include within a block fragment

OscartGiles:
Failing test with include in a fragment

See https://github.com/djc/askama/issues/1063


GuillaumeGomez:
I'll give a look when I have time but it should definitely work.

PR #1069, opened 2024-06-18
fuzz: add harness for filters and escape

manunio:
This pr:
- Adds a harness for filters and escape, this has helped in finding bugs like https://github.com/djc/askama/pull/1067
- Moves all fuzz targets to a root fuzz crate, I think this approach is better than creating it per crate.


djc:
Would you mind making this two separate commits (or PRs)? One that moves the existing parser fuzz tests to the new crate, and a separate one that adds the new filter tests.

djc (approved):

PR #1071, opened 2024-06-21
filter: add check against large input

manunio:
This will fix following recently reported bugs:
 - https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69772
 - https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69797


djc (approved):
Thanks!

PR #1072, opened 2024-06-23
Limit nested filters to avoid stack overflow 2

manunio:
This will fix: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=66029.
This pr includes all the work at https://github.com/djc/askama/pull/979 done by @djc as well as updated ui test.

For ui test: I think previous input for filter-recursion ui test was not right as it
was not triggering error for recursion itself, for example:

```rust
#[test]
fn fuzzed_filter_recursion() {
    const TEMPLATE: &str = include_str!("../tests/filter-recursion.txt");
    if let Err(e) = Ast::from_str(TEMPLATE, None, &Syntax::default()) {
        panic!("{e}");
    }
}
```

```sh
---- tests::fuzzed_filter_recursion stdout ----
thread 'tests::fuzzed_filter_recursion' panicked at askama_parser/src/tests.rs:952:9:
failed to parse template source at row 1, column 255 near:
"|A|AA|A|A|A|A|AA|A|A|A|A|AA|A|A|A|A|AA|A"...
```


GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

askama_parser/src/lib.rs
```diff
@@ -636,13 +636,10 @@ impl Level {
...
 
 #[allow(clippy::type_complexity)]
-fn filter(i: &str, level: Level) -> ParseResult<'_, (&str, Option<Vec<Expr<'_>>>)> {
-    let (i, (_, fname, args)) = tuple((
-        char('|'),
-        ws(identifier),
-        opt(|i| Expr::arguments(i, level, false)),
-    ))(i)?;
-    Ok((i, (fname, args)))
+fn filter<'a>(i: &'a str, level: &mut Level) -> ParseResult<'a, (&'a str, Option<Vec<Expr<'a>>>)> {
```
djc, line 639: `Level` is `Copy`, why break the convention of passing it by value here?

PR #1082, opened 2024-08-13
book: fix typo in Template Syntax page

m4tx:
The semicolon in `{% let entity = entity; %}` causes the template compilation to fail, so this commit removes it.


djc (approved):

djc:
Thanks!

PR #1089, opened 2024-08-26
Suggest disable html escaping in templates in templates section

### LINKED ISSUE #1088: Templates in templates need "none" escape
When rendering templates in templates like https://github.com/djc/askama/blob/main/book/src/template_syntax.md#templates-in-templates, the inner templates needs to be rendered without escaping; otherwise the Html of the inner template is printed literally.

If that that's the intended behavior, feel free to close this issue.

Otherwise, the tests in https://github.com/djc/askama/blob/main/testing/tests/render_in_place.rs should include Html tags to not miss that in the future. In the mean time, a workaround is to disable escaping manually:

```html
{% for my_template in my_templates %}
      {{ my_template|escape("none") }}
{% endfor %}
```

malteneuss:
Suggest in the book that when rendering Html templates in templates, we may want to disable Html escaping.

Fixes #1088

### DIFF
--- a/book/src/template_syntax.md
+++ b/book/src/template_syntax.md
@@ -584,6 +584,11 @@ let t = RenderInPlace { s1: SectionOne { a: "a", b: "b" } };
 assert_eq!(t.render().unwrap(), "Section 1: A=a\nB=b")
 ```
 
+**Note that if your inner template** like `SectionOne` **renders HTML content, you may want to
+disable escaping** when injecting it into an outer template, e.g. `{{ s1|escape("none") }}`;
+otherwise it will render the HTML content literally.
+Askama [escapes HTML variables](#html-escaping) by default.
+
 See the example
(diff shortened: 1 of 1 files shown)


book/src/template_syntax.md
```diff
@@ -584,6 +584,11 @@ let t = RenderInPlace { s1: SectionOne { a: "a", b: "b" } };
 assert_eq!(t.render().unwrap(), "Section 1: A=a\nB=b")
 ```
 
+**Note that if your inner template** like `SectionOne` **renders HTML content, you may want to
+disable escaping** when injecting it into an outer template, e.g. `{{ s1|escape("none") }}`;
```
djc, line 588: I'd use `safe` instead of `escape("none")`, which is a little more concise.

djc (commented):
Thanks!

book/src/template_syntax.md
malteneuss: Done

PR #1093, opened 2024-09-11
Fix include statements within block fragments

### LINKED ISSUE #1063: Include statement has no effect when used from a block fragment

mlabrenz117:
Fixes #1063. Includes test case courtesy of @OscartGiles in #1066.


djc (approved):

djc:
This is great, thanks!

PR #1097, opened 2024-09-27
Accept trailing comma in array

### LINKED ISSUE #1086: Allow Trailing Commas in Arrays Inside Template Expressions
Basically what the title says.

To give an example of what I mean, this here compiles just fine:

```html
<nav class="flex justify-between items-center max-w-7xl gap-x-12">
    {% for (route, name) in [
        ("/about", "About"),
        ("/contact", "Contact"),
        ("/events", "Events"),
        ("/history", "History")
    ] %}
    <div class="gap-x-6 leading-6 hover:bg-gray-200">
        <a href="{{ route }}"><span>{{ name }}</span></a>
    </div>
    {% endfor %}
</nav>
```

But this here doesn't:

```html
<nav class="flex justify-between items-center max-w-7xl gap-x-12">
    {% for (route, name) in [
        ("/about", "About"),
        ("/contact", "Contact"),
        ("/events", "Events"),
        ("/history", "History"),
    ] %}
    <div class="gap-x-6 leading-6 hover:bg-gray-200">
        <a href="{{ route }}"><span>{{ name }}</span></a>
    </div>
    {% endfor %}
</nav>
```

Note the little comma after the `("/history", "History")` tuple in the second example.

I'm not sure if this is trivial to implement or not, but I thought I might bring it to y'alls' attention.

It wasn't that of a big deal in the end though, so I wouldn't call this a "major issue" or whatever, not by a long shot. Though, since Rust allows such trailing commas, I think `askama` should as well.

For reference, this is the error I got:
```
error: problems parsing template source at row 30, column 31 near:
       ",\n    ] %}\n    <div class=\"gap-x-6 leadi"...
  --> src/routes/mod.rs:96:10
   |
96 | #[derive(Template)]
   |          ^^^^^^^^
   |
```

Running with `RUSTFLAGS="-Zmacro-backtrace" cargo +nightly build` gives the following:
```
error: problems parsing template source at row 30, column 31 near:
       ",\n    ] %}\n    <div class=\"gap-x-6 leadi"...
  --> src/routes/mod.rs:96:10
   |
96 | #[derive(Template)]
   |          ^^^^^^^^ in this derive macro expansion
   |
  ::: /home/max/.cargo/registry/src/index.crates.io-6f17d22bba15001f/askama_derive-0.12.5/src/lib.rs:22:1
   |
22 | pub fn derive_template(input: TokenStream) -> TokenStream {
   | --------------------------------------------------------- in this expansion of `#[derive(Template)]`
```

Neither error was too helpful, but it does point to line 30, column 31, which in my code is precisely the closing parenthesis before the trailing comma; careful reading helps at least.

aquarhead:
Fix #1086

### DIFF
--- a/askama_parser/src/expr.rs
+++ b/askama_parser/src/expr.rs
@@ -270,7 +270,7 @@ impl<'a> Expr<'a> {
                     separated_list0(char(','), ws(move |i| Self::parse(i, level))),
                     Self::Array,
                 ),
-                char(']'),
+                pair(opt(ws(char(','))), char(']')),
             )),
         )(i)
     }
--- a/testing/tests/loops.rs
+++ b/testing/tests/loops.rs
@@ -84,6 +84,16 @@ fn test_for_array() {
 }
 
 #[derive(Template)]
+#[template(source = "{% for i in [1, 2, 3, ] %}{{ i }}{% endfor %}", ext = "txt")]
+struct ForArrayTailingCommaTemplate;
+
+#[test]
+fn test_for_array_trailing_comma() {
+    let t = ForArrayTailingCommaTemplate;
+    assert_eq!(t.render().unwrap(), "123");
+}
+
+#[derive(Template)]
 #[template(
     source = "{% for i in [1, 2, 3].iter() %}{{ i }}{% endfor %}",
     ext = "txt"


aquarhead:
Just FYI you can allow other jobs in a matrix to run to complete instead of cancel when one of it fails https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast (it defaults to `true`)

I don't think the failure is related to my change?

djc:
That's right, the failure is unrelated. If you're able to add a separate commit that changes the matrix to allowing it to complete, that would be great!

askama_parser/src/expr.rs
```diff
@@ -270,7 +270,7 @@ impl<'a> Expr<'a> {
                     separated_list0(char(','), ws(move |i| Self::parse(i, level))),
                     Self::Array,
                 ),
-                char(']'),
+                alt((tag(",]"), tag("]"))),
```
djc, line 273: Hmm, I think we should accomodate any whitespace appearing between `,` and `]`.

djc (commented):
Thanks!

askama_parser/src/expr.rs
aquarhead: Good point, I'll add/update the test too

djc:
Test failure addressed in #1098.

djc (approved):

Issue #1100, opened 2024-10-08
Multiple `if let` blocks for the same variable sometimes fail to prepend self


anna-is-cute:
Hi! Using askama 0.12.1, I have encountered a strange issue. Sometimes, when using multiple `if let` blocks for the same variable, the second block fails to put the `self.` before the variable name in the generated code. If you change the template to say `self.var_name` instead, it works.

Here's a reproducible test case. Note that something about this structure/these field names causes the issue; I tried with a simple `struct Thing { field: usize }` and could not reproduce the issue.

```rs
use askama::Template;

#[derive(Template)]
#[template(
    source = r#"
{% if let Some(news) = news %}
<section>
    <header>Today's news</header>

    {% for source in news.sources %}
        <div class='list'>
            <strong>{{ source.name }}</strong>
            <ul>
                {% for story in source.stories %}
                    <li>
                        <div class='story'>
                            <div>{{ story.title }}</div>
                            {% if let Some(description) = story.description %}
                                <div>{{ description }}</div>
                            {% endif %}
                        </div>
                    </li>
                {% endfor %}
            </ul>
        </div>
    {% endfor %}
</section>
{% endif %}

{% if let Some(news) = news %}
    hi
{% endif %}
"#,
    ext = "html",
    print = "all",
)]
pub struct ATemplate {
    pub news: Option<NewsInfo>
}

pub struct NewsInfo {
    pub sources: Vec<NewsSource>,
}

pub struct NewsSource {
    pub name: String,
    pub stories: Vec<NewsStory>,
}

pub struct NewsStory {
    pub title: String,
    pub description: Option<String>,
}

fn main() {
    let temp = ATemplate {
        news: Some(NewsInfo {
            sources: vec![
                NewsSource {
                    name: "Test".to_string(),
                    stories: vec![
                        NewsStory {
                            title: "Title".to_string(),
                            description: Some("Description".to_string())
                        },
                    ],
                },
            ],
        }),
    };

    println!("{}", temp.render().unwrap());
}
```

If you need any more information, please let me know. Also thank you for the wonderful project!

djc:
Thanks for the detailed report! If you could submit this as a PR containing an integration test case, that would make my life easier. If you're able to look at the code generator to see what code it wrongly generates, you might even be able to fix it. My guess is that we're failing to clean up some local-scope variables somewhere...

anna-is-cute:
Fixed in 3c7389fd0703a9d3d75e312cf731d88a621110c1.

PR #1101, opened 2024-10-08
Add test for if let containing a for loop

anna-is-cute:
Add a test for #1100, currently failing.


djc:
See #1102.

djc:
Thanks for the test case! Not sure how long this has been there...

PR #1107, opened 2024-11-18
Canonicalize include paths before emitting

djkoloski:
This allows `CARGO_MANIFEST_DIR` to be a relative path, which can be useful in non-cargo build systems.


djc (approved):
Thanks!

Issue #877, opened 2023-10-06
relative template paths and include_str support


henke443:
I would want to have the feature.html template be in the same folder as feature.rs. 

It would be cool if you could do either one of:

A)

```rs

pub static FEATURE_TEMPLATE_HTML: &'static str = include_str!("feature.html");

#[derive(Template)]
#[template(source=FEATURE_TEMPLATE_HTML, ext = "html")] 
struct featureTemplate<'a> { 
    name: &'a str,
}
```

B)

```rs
#[derive(Template)] 
#[template(path="./feature.html")]  // And maybe just path="feature.html" is using the toml dirs? Like how "import" works in nodejs
struct featureTemplate<'a> { 
    name: &'a str, 
}
```

djc:
I think I'm open to either of these, if you want take a whack at implementing this. Have a look at https://github.com/djc/askama/blob/main/askama_derive/src/generator.rs#L81 and https://github.com/djc/askama/blob/main/askama_derive/src/input.rs#L11.

Kijewski:
The first option is impossible to implement. The proc macro will only see the token `FEATURE_TEMPLATE_HTML`, and it is currently not possible to evaluate the context. E.g. `#[doc = include_str()]` works, but `#[doc = SOME_CONSTANT]` does not.

But I like the second option!

Zonnex:
Yes please! I REALLY wish to co-locate the html file with the rust file. I've actively looked for alternatives to askama that does this, but I haven't found any, and I don't want to use anything else because I really like how askama does things.
I doubt I'm able to implement this myself, I'll take a look but I otherwise that someone else is able to!

BuriKizilkaya:
Is there any progress?

djc:
No, I guess there isn't. Do you want to submit a PR?

BuriKizilkaya:
I will try....

andrewbaxter:
> The first option is impossible to implement. The proc macro will only see the token `FEATURE_TEMPLATE_HTML`, and it is currently not possible to evaluate the context. (E.g. that's why `#[doc = include_str()]` works, but `#[doc = SOME_CONSTANT]` does not.)
> 
> But I like the second option!

I think ATM the second option is also impossible to implement: https://github.com/rust-lang/rust/issues/54725 (going on 6 years)

`include_str` and I assume `doc` are currently using compiler magic to do it IIUC.

joshka:
> i.e. I don't think there's a way to do this at the moment.

[...]

PR #1112, opened 2024-12-03
Load templates from relative paths

joshka:
Templates can now be placed directly next to the source file that they
are defined in as a default. This relies on an unstable rust compiler
feature, which exposes the source file to proc macros. See
<https://github.com/rust-lang/rust/issues/54725> for more info.

This requires the nightly compiler to run, and enabling the
proc_macro_span and procmacro2_semver_exempt cfg flags. To build / test:

```shell
RUSTFLAGS='--cfg proc_macro_span --cfg procmacro2_semver_exempt' \
  cargo +nightly build
```
Fixes: <https://github.com/djc/askama/issues/877>

---

TODO:
- [ ] docs
- [ ] work out the right name for the feature flag (there's likely a less ambiguous name than `relative-paths` for this)
- [ ] work out if it's possible to compile this without using the nightly compiler. The proc-macro2 crate has shims replacements for the unstable nightly code which can be enabled somehow, but I haven't dug deep enough to understand how / why that is.

### DIFF
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -29,6 +29,12 @@ with-axum = ["askama_derive/with-axum"]
 with-rocket = ["askama_derive/with-rocket"]
 with-warp = ["askama_derive/with-warp"]
 
+## Enables the ability to put templates in a directory relative to the source file that uses them.
+## Requires a nightly compiler and adding:
+## RUSTFLAGS='--cfg proc_macro_span --cfg procmacro2_semver_exempt'
+## to your cargo build command.
+relative-paths = ["askama_derive/relative-paths"]
+
 [dependencies]
 askama_derive = { version = "0.13", path = "../askama_derive" }
 askama_escape = { version = "0.11", path = "../askama_escape" }
--- a/askama_derive/src/config.rs
+++ b/askama_derive/src/config.rs
@@ -26,7 +26,22 @@ impl<'a> Config<'a> {
         template_whitespace: Option<&str>,
     ) -> std::result::Result<Config<'a>, CompileError> {
         let root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
-        let default_dirs = vec![root.join("templates")];
+        let root_path = root.join("templates");
+        let default_dirs;
+        #[cfg(feature = "relative-paths")]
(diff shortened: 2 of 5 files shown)


askama/Cargo.toml
```diff
@@ -29,6 +29,12 @@ with-axum = ["askama_derive/with-axum"]
 with-rocket = ["askama_derive/with-rocket"]
 with-warp = ["askama_derive/with-warp"]
 
+## Enables the ability to put templates in a directory relative to the source file that uses them.
+## Requires a nightly compiler and adding:
+## RUSTFLAGS='--cfg proc_macro_span --cfg procmacro2_semver_exempt'
+## to your cargo build command.
```
joshka, line 35: Consider adding the `document-features` crate so this info can automatically populate the lib.rs doc comment without having to duplicate it.

askama_derive/src/config.rs
```diff
@@ -26,7 +26,22 @@ impl<'a> Config<'a> {
         template_whitespace: Option<&str>,
     ) -> std::result::Result<Config<'a>, CompileError> {
         let root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
-        let default_dirs = vec![root.join("templates")];
+        let root_path = root.join("templates");
+        let default_dirs;
```
joshka, line 30: It's probable that the logic of which dirs end up in the list when config is taken into account might have to be a little more complex than this current logic. The crux of the problem is solved however, leaving the choice of how it should be behave remaining. I.e. if the user has a config file and it has dirs specified should we still resolve relative paths, should there be a config for this, ...

joshka:
> Wouldn't it be sufficient to have `template-relative-paths` or `template-paths-relative` as the flag?
> 
> Great work either way, I hope to see this finished 👍

Thanks - unless my schedule frees up a bit, I won't work on this for a couple of weeks. But it's likely that there's a few points that need to be worked out with the design / features of this (mentioned in the todo list above and in the comments in related issue. If you want to help out with that, perhaps make a note of these and work them to resolution with the repo owners. I'm also fine with you (or anyone else) taking over this PR to land it. Otherwise, remind me in February again about pushing this forward.

PR #1113, opened 2024-12-10
Update README.md: fix docs link to rinja-rs org

VojtaStanek:
The original link does not work


Kijewski:
Thank you for your PR! We are currently in the process of re-consolidation rinja and askama into one project again. The current URL is a work-in-progress, and subject to change.

